PR #4, opened 2018-04-18
[sfputilbase]: Add logic to parse the title of port_config.ini file

stcheng:
With adding more columns into the port_config.ini file, the previous
logic of checking len(line.split()) == 4 is no longer valid. With
adding the logic of parsing the title of the file, the port index
could be directly retrieved depending on which column it is.

Signed-off-by: Shu0T1an ChenG <shuche@microsoft.com>

### DIFF
--- a/sonic_sfp/sfputilbase.py
+++ b/sonic_sfp/sfputilbase.py
@@ -272,7 +272,10 @@ class SfpUtilBase(object):
         # if something already exists
         for line in f:
             line.strip()
+            title = []
             if re.search("^#", line) is not None:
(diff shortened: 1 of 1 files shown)


jleveque:
In the future, we should contain all port_config.ini parsing functions in one file to prevent things like this getting forgotten. We now have https://github.com/Azure/sonic-buildimage/blob/master/src/sonic-config-engine/portconfig.py, but we shouldn't need to install sonic-config-engine just to import it. I think minigraph.py and portconfig.py should be part of a new package that we can import into other applications like this *and* sonic-config-engine.

PR #10, opened 2018-08-09
[sonic_eeprom] If reading from what appears to be a corrupt cache file, delete file and read directly from EEPROM

jleveque:
Before this PR, if a corrupt cache file exists, decode-syseeprom will fail indefinitely until the file is manually deleted.

### DIFF
--- a/sonic_eeprom/eeprom_base.py
+++ b/sonic_eeprom/eeprom_base.py
@@ -233,9 +233,17 @@ class EepromDecoder(object):
         F.seek(self.s + offset)
         o = F.read(byteCount)
         if len(o) != byteCount:
-            raise RuntimeError("expected to read %d bytes from %s, " \
-                               %(byteCount, self.p) +
-                               "but only read %d" %(len(o)))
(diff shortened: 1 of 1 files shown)


sonic_eeprom/eeprom_base.py
```diff
@@ -233,9 +233,17 @@ def read_eeprom_bytes(self, byteCount, offset=0):
...
-            raise RuntimeError("expected to read %d bytes from %s, " \
-                               %(byteCount, self.p) +
-                               "but only read %d" %(len(o)))
+            # If we read from the cache file, it may be corrupt. Delete it
+            # and try again, this time reading from the actual EEPROM.
+            if not self.cache_update_needed:
+                os.remove(self.cache_name)
+                self.cache_update_needed = True
+                F.close()
+                return self.read_eeprom_bytes(byteCount, offset)
```
qiluo-msft: >read_eeprom_bytes [](start = 28, length = 17)

The code may achieve the goal. However, the recursion logic is not intuitive to solve a simple if-else problem.

I will not block this PR. Strongly suggest refactor the code.

PR #13, opened 2018-09-13
Add new platform management API. Currently with support for chassis, PSUs and fans

jleveque:
…


sonic_platform_base/chassis_base.py
```diff
@@ -0,0 +1,92 @@
...
+        return None 
+
+    @abc.abstractmethod
+    def get_reboot_cause(self):
+        """
+        Retrieves the cause of the previous reboot
+
+        Returns:
+            An integer, the number of power supply units available on this
+            chassis
```
lguohan: should be reboot cause.

sonic_platform_base/fan_base.py
```diff
@@ -0,0 +1,100 @@
...
+        Retrieves the speed of fan as a percentage of full speed
+
+        Returns:
+            An integer, the percentage of full fan speed, in the range 0 (off)
+                 to 100 (full speed)
+        """
+        return 0
+
+    @abc.abstractmethod
+    def get_expected_speed(self):
```
lguohan: get_target_speed

sonic_platform_base/chassis_base.py
jleveque: Copy/paste error. Thanks for catching.

sonic_platform_base/fan_base.py
jleveque: Changed here and in psu_base.py

sonic_platform_base/chassis_base.py
```diff
@@ -0,0 +1,94 @@
...
+except ImportError as e:
+    raise ImportError(str(e) + " - required module not found")
+
+
+class ChassisBase(device_base.DeviceBase):
+    """
+    Abstract base class for interfacing with a platform chassis
+    """
+
+    __metaclass__ = abc.ABCMeta
```
Staphylo: This `__metaclass__` syntax is python2 compliant but will not do anything for python3.
With this syntax instantiating ChassisBase will fail as expected in python2 but will succeed in python3.
If this compatibility matters to you, the six module will have to be used since the syntax are different in both language.

Also `__metaclass__` is already defined in `device_base.DeviceBase`

sonic_platform_base/fan_base.py
```diff
@@ -0,0 +1,100 @@
...
+    from . import device_base
+except ImportError as e:
+    raise ImportError(str(e) + " - required module not found")
+
+
+class FanBase(device_base.DeviceBase):
+    """
+    Abstract base class for interfacing with a fan module
+    """
+    __metaclass__ = abc.ABCMeta
```
Staphylo: metaclasses are inherited, see https://gist.github.com/Staphylo/5597a17d781afcb7c52b3c78bbef8c32

sonic_platform_base/fan_base.py
```diff
@@ -0,0 +1,100 @@
...
+        Retrieves the target (expected) speed of the fan
+
+        Returns:
+            An integer, the percentage of full fan speed, in the range 0 (off)
+                 to 100 (full speed)
+        """
+        return 0
+
+    @abc.abstractmethod
+    def get_speed_tolerance(self):
```
Staphylo: I'm assuming this is only for the fan daemon logic usage?

sonic_platform_base/fan_base.py
```diff
@@ -0,0 +1,100 @@
...
+        """
+        Retrieves the target (expected) speed of the fan
+
+        Returns:
+            An integer, the percentage of full fan speed, in the range 0 (off)
+                 to 100 (full speed)
+        """
+        return 0
+
+    @abc.abstractmethod
```
Staphylo: Does this method have to be abstract or a default implementation that is always overridable would do?

sonic_platform_base/fan_base.py
```diff
@@ -0,0 +1,100 @@
...
+    def set_speed(self, speed):
+        """
+        Sets the fan speed
+
+        Args:
+            speed: An integer, the percentage of full fan speed to set fan to,
```
Staphylo: I see everywhere that the fan speed is an int from 0 to 100.
Some device dealing with fans can have finer grained control (like 0 to 255) which means loosing a bit of precision.
Was this taken into consideration and integer deemed better than floats?

sonic_platform_base/psu_base.py
```diff
@@ -0,0 +1,99 @@
...
+    from . import device_base
+except ImportError as e:
+    raise ImportError(str(e) + " - required module not found")
+
+
+class PsuBase(device_base.DeviceBase):
+    """
+    Abstract base class for interfacing with a power supply unit
+    """
+    __metaclass__ = abc.ABCMeta
```
Staphylo: Same comment for metaclass inheritance.

sonic_platform_base/psu_base.py
```diff
@@ -0,0 +1,99 @@
...
+    # Possible fan directions
+    FAN_DIRECTION_INTAKE = "intake"
+    FAN_DIRECTION_EXHAUST = "exhaust"
+
+    # Possible fan status LED colors
+    STATUS_LED_COLOR_GREEN = "green"
+    STATUS_LED_COLOR_RED = "red"
+    STATUS_LED_COLOR_OFF = "off"
+
+    @abc.abstractmethod
```
keboliu: One suggestion:  handle the PSU fans in chassis class just like other fans. 
And is it possible that one PSU installed with multi fans?

sonic_platform_base/chassis_base.py
```diff
@@ -0,0 +1,94 @@
...
+
+    # Possible reboot causes
+    REBOOT_CAUSE_POWER_LOSS = "power_loss" 
+    REBOOT_CAUSE_THERMAL_OVERLOAD = "thermal_overload"
+    REBOOT_CAUSE_INSUFFICIENT_FAN = "insufficient_fan"
+    REBOOT_CAUSE_WATCHDOG = "watchdog"
+    REBOOT_CAUSE_SOFTWARE = "software"
+
+    # List of all fans available on the chassis
+    fan_list = []
```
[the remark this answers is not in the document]
jleveque: Agreed. Will address in the next commit. Thanks!

sonic_platform_base/chassis_base.py
```diff
@@ -0,0 +1,94 @@
...
+        Retrieves the base MAC address for the chassis
+
+        Returns:
+            A string containing the MAC address in the format
+            'XX:XX:XX:XX:XX:XX'
+        """
+        return None 
+
+    @abc.abstractmethod
+    def get_reboot_cause(self):
```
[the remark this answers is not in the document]
jleveque: > Reboot cause can be persistent on some product and have to be explicitly cleared.
> Otherwise reading this information after a reboot will show all the reload causes that happened previous times.
> 
> I would suggest either a default parameter `clear=False` to this method or to add a new one called `clear_reboot_cause`.
> Also I think this function should only be called once with `clear=True` and the result cached somewhere in a tmpfs. This means that future asks for reload cause information would go through the cache.

Since clearing the reboot cause is platform-specific functionality, I believe this logic should be left to the platform-specific implementation of the function. If the platform requires clearing the reboot cause, it should be done in the implementation and a file saved to tmpfs. The function can check for the presence of the file first before attempting to read/clear the register. I prefer this to adding a `clear=True` parameter to the base method, and I think adding a `clear_reboot_cause()` method is unnecessary, because there should be no need to explicitly clear the reboot cause.

> Some platform might also have limitation on the number of reboot causes that can be stored.

If you're concerned that not all platforms will support all reboot causes, that is expected. The base class will define all possible reboot causes that SONiC understands; the platform-specific implementation will simply return the subset which is applicable to that platform. If this doesn't address your concern, please elaborate.

sonic_platform_base/chassis_base.py
jleveque: Thanks for bringing this up, as this library will most likely be imported into both Python 2 and Python 3 code. I will address this in my next commit.

sonic_platform_base/fan_base.py
jleveque: Thanks. Will address in next commit.

sonic_platform_base/fan_base.py
jleveque: That is the initial intention. We may find other uses in the future, however.

sonic_platform_base/fan_base.py
jleveque: Assuming tolerance is most commonly implemented as a static attribute, we could define a `tolerance` member and have that returned here as the default implementation. Would that suffice for you?

sonic_platform_base/psu_base.py
jleveque: Thanks. Will address in next commit.

sonic_platform_base/chassis_base.py
```diff
@@ -0,0 +1,94 @@
...
+        """
+        fan = None
+
+        try:
+            fan = self.fan_list[index]
+        except IndexError:
+            sys.stderr.write("Fan index {} out of range (0-{})\n".format(
+                             index, len(self.fan_list)-1))
+
+        return fan
```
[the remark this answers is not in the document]
jleveque: Addressed in next commit.

sonic_platform_base/fan_base.py
jleveque: Would you prefer a 0.0 ... 1.0 float scale? This should allow for finer-grained control if necessary, while maintaining a more universal scale.

sonic_platform_base/chassis_base.py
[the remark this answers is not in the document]
Staphylo: > > Reboot cause can be persistent on some product and have to be explicitly cleared.
> > Otherwise reading this information after a reboot will show all the reload causes that happened previous times.
> > I would suggest either a default parameter `clear=False` to this method or to add a new one called `clear_reboot_cause`.
> > Also I think this function should only be called once with `clear=True` and the result cached somewhere in a tmpfs. This means that future asks for reload cause information would go through the cache.
> 
> Since the need to clear the reboot cause after reading is platform-specific functionality, I believe this logic should be left to the platform-specific implementation of the function. If the platform requires clearing the reboot cause, it should be done in the implementation and a file saved to tmpfs. The function can check for the presence of the file first before attempting to read/clear the register. I prefer this to adding a `clear=True` parameter to the base method, and I think adding a `clear_reboot_cause()` method is unnecessary, because there should be no need to explicitly clear the reboot cause.
> 

I would assume all the platforms to behave the same way with regard to the reboot causes. However this information would have to be crowd-sourced from other vendors as well for more relevant data.
I'm fine with letting the responsibility of clearing and caching the reboot cause to the plugin implementation.
Is there a place in SONiC where such cache files should go? Currently almost everything is backed by the flash including `/tmp`. The only place that is a `tmpfs` and would feel "suitable" seems to be `/run`. Is there some guidelines somewhere regarding this? I would be interested to know the reason why `/tmp` is not a `tmpfs`.

> > Some platform might also have limitation on the number of reboot causes that can be stored.
> 
> If you're concerned that not all platforms will support all reboot causes, that is expected. The base class will define all possible reboot causes that SONiC understands; the platform-specific implementation will simply return the subset which is applicable to that platform. If this doesn't address your concern, please elaborate.

My concern was a bit different. Your previous answer ended up answered my question but I'll elaborate a bit more.
Basically the concept of a reboot cause is to be stored on an external device until being read.
These devices usually have space restriction and can only save a given number of reboot cause after which they won't store anything until being cleared. For example if your device can store 2 reboot cause and you trigger 2 watchdogs and then 1 powerloss, you might not see the powerloss if you haven't cleared the reboot cause after each reboot.

sonic_platform_base/fan_base.py
Staphylo: I don't mind either way, integer percentages 0 to 100 sounds good to me.
I just wanted to bring this topic up as the [sysfs interface](https://www.kernel.org/doc/Documentation/hwmon/sysfs-interface) defines a pwm value to be ranging from 0 to 255.

sonic_platform_base/psu_base.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class PsuBase(device_base.DeviceBase):
+    """
+    Abstract base class for interfacing with a power supply unit
+    """
+    __metaclass__ = abc.ABCMeta
+
+    # Possible fan directions
+    FAN_DIRECTION_INTAKE = "intake"
```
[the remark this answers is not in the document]
keboliu: sounds good to me to have PSU derived from FAN base.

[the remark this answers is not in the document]
jleveque: @keboliu: My suggestion was to have a PSU object contain one fan object, rather than derive PsuBase from FanBase. I think this makes more sense, especially if certain PSU modules contain more than fan. Do you agree? Also, are you aware of any PSU module which contains more than one fan?

sonic_platform_base/fan_base.py
jleveque: If all platforms used sysfs, the 0-255 scale would be fine. However, we'd like a universal scale that will work with other interfaces (IPMI, etc.).

jleveque:
@keboliu:

> one general question, previously when add a new API to the base class, need to implement a stub to all sub class for each platform (as [Azure/sonic-buildimage#2017](https://github.com/Azure/sonic-buildimage/issues/2017) described), can this be avoided in the new implementation?

This has been address in: https://github.com/Azure/sonic-platform-common/pull/13/commits/bd6b6edfd0918437dcb71fc10640c9fe7a318b35. In lieu of an abstract base class, we will implement either a build-time or run-time check that will output all methods which are not implemented in order to alert vendors to methods they may have missed.

sonic_platform_base/platform_base.py
```diff
@@ -0,0 +1,64 @@
+#
+# platform_base.py
+#
+# Abstract base class for implementing platform-specific functionality in
+# SONiC. This is the root class of sonic_platform_base hierarchy. A platform
+# may contain one or more chassis. Classes derived from this class provide
```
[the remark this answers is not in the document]
jleveque: Agreed. This is addressed in https://github.com/Azure/sonic-platform-common/pull/13/commits/d2815714ea4a8b3eb0f88d7e2e43e95d80c245a0. Now 1 platform == 1 chassis. I have also added the concept of a "module" for modular chassis in https://github.com/Azure/sonic-platform-common/pull/13/commits/7ab83cfb90ab042cd11254a5a675d59a5b042d09. If in the future there is such a concept as a "multi-chassis switch" we can change this back to a list.

sonic_platform_base/watchdog_base.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class WatchdogBase(device_base.DeviceBase):
+    """
+    Abstract base class for interfacing with a hardware watchdog module
+    """
+
+    # Possible fan directions
+    FAN_DIRECTION_INTAKE = "intake"
+    FAN_DIRECTION_EXHAUST = "exhaust"
```
lguohan: why fan direction in watchdog base class?

sonic_platform_base/watchdog_base.py
```diff
@@ -0,0 +1,50 @@
...
+
+    def arm(self, seconds):
+        """
+        Arm the hardware watchdog with a timeout of <seconds> seconds
```
lguohan: can you describe the behavior more specifically? if the watchdog is already armed and call this api will refresh the watchdog timer.

sonic_platform_base/watchdog_base.py
```diff
@@ -0,0 +1,50 @@
+#
+# watchdog_base.py
+#
+# Abstract base class for implementing a platform-specific class with which
+# to interact with a hardware watchdog module in SONiC
```
lguohan: how do we know if the platform has such capability or not? do we have query API for this?

sonic_platform_base/watchdog_base.py
lguohan: hardware watchdog may have maximum value of the timeout. I think we need to have a error code to return when seconds exceed the maximum value that is support by the hardware.

jleveque: Currently the function will return False if it is unable to arm the watchdog. Would you prefer an integral return value to specify the reason?

Alternatively (or in addition to), we could ask the vendor to define a maximum value which the user can query. Thoughts?

sonic_platform_base/watchdog_base.py
jleveque: To avoid the need for additional functions with which to query device/capability, the current design is such that the querying is done by getting the device object. If the object is null (None), the device/capability is not present on the platform. For example:

```
#!/usr/bin/env python

import sonic_platform

platform = sonic_platform.platform.Platform()

chassis = platform.get_chassis()
if not chassis:
    print("Error getting chassis!")
    sys.exit(1)

watchdog = chassis.get_watchdog()
if not watchdog:
    print("Watchdog not present!")
else:
    watchdog.arm(10)
```

sonic_platform_base/watchdog_base.py
jleveque: Leftover after copy/paste; meant to be deleted. Resolved in d605896.

sonic_platform_base/chassis_base.py
```diff
@@ -0,0 +1,207 @@
...
+
+
+class ChassisBase(device_base.DeviceBase):
+    """
+    Base class for interfacing with a platform chassis
+    """
+
+    # Possible reboot causes
+    REBOOT_CAUSE_POWER_LOSS = "power_loss"
+    REBOOT_CAUSE_THERMAL_OVERLOAD = "thermal_overload"
```
padmanarayana: Would it be possible to differentiate between thermal overload of Host CPU vs ASIC ?

sonic_platform_base/chassis_base.py
```diff
@@ -0,0 +1,207 @@
...
+    REBOOT_CAUSE_INSUFFICIENT_FAN = "insufficient_fan"
+    REBOOT_CAUSE_WATCHDOG = "watchdog"
+    REBOOT_CAUSE_SOFTWARE = "software"
```
padmanarayana: Can we differentiate between warm vs cold reboot ?

sonic_platform_base/device_base.py
```diff
@@ -0,0 +1,66 @@
...
+                this method will block until a change is detected.
+
+        Returns:
+            (bool, dict):
+                - True if call successful, False if not;
+                - Dict where key is device ID and value is device event,
+                  status='1' represents device inserted,
+                  status='0' represents device removed. Ex. {'0': '1', '1': '0'}
+        """
+        raise NotImplementedError
```
padmanarayana: Just wanted to get confirmation that if a platform takes a Baseboard Management Controller (BMC) based approach (to handle thermals, events etc), these APIs can remain unimplemented..

sonic_platform_base/watchdog_base.py
```diff
@@ -0,0 +1,46 @@
...
+
+        Returns:
+            A boolean, True if watchdog is disarmed successfully, False if not
+        """
+        raise NotImplementedError
+
+    def is_armed(self):
+        """
+        Retrieves the armed state of the hardware watchdog and if armed, the
+        number of seconds remaining on the watchdog timer
```
padmanarayana: If watchdog is implemented by CPLD, say (instead of Host CPU), the number of seconds remaining may not be retrievable/supported.. Can the API be split into 2 : 1 to check if wdog is armed and another to retrieve remaining seconds?

sonic_platform_base/psu_base.py
```diff
@@ -0,0 +1,45 @@
...
+# psu_base.py
+#
+# Abstract base class for implementing a platform-specific class with which
+# to interact with a power supply unit (PSU) in SONiC
+#
+
+from . import device_base
+
+
+class PsuBase(device_base.DeviceBase):
```
zzhiyuan: Will PsuBase from sonic_psu be moved to this implementation?

sonic_platform_base/chassis_base.py
lguohan: we also need a REBOOT_CAUSE_NA = "na" if hardware cannot determine the exact reason.

jleveque: My intentions were that `REBOOT_CAUSE_SOFTWARE` would be the "N/A", meaning that the reboot was not caused by hardware. I can replace `REBOOT_CAUSE_SOFTWARE` with `REBOOT_CAUSE_NA` if you prefer.

sonic_platform_base/chassis_base.py
jleveque: Addressed in https://github.com/Azure/sonic-platform-common/pull/13/commits/5fbf599a801665066545d8261ac3e3dfba34d762. Now differentiating between three types of thermal overloads, REBOOT_CAUSE_THERMAL_OVERLOAD_CPU, REBOOT_CAUSE_THERMAL_OVERLOAD_ASIC and REBOOT_CAUSE_THERMAL_OVERLOAD_OTHER.

sonic_platform_base/watchdog_base.py
jleveque: Addressed in https://github.com/Azure/sonic-platform-common/pull/13/commits/5fbf599a801665066545d8261ac3e3dfba34d762.

sonic_platform_base/device_base.py
jleveque: The intention here is for daemons to receive notifications of changes in platform hardware state in order to update SONiC's view of device state. For instance, if a tranceiver, PSU or fan tray is added or removed. Can you provide an example of an event handled by the BMC which SONiC would not need to be made aware of?

sonic_platform_base/fan_base.py
Staphylo: No concerns. I'm was just curious if the fand was in the roadmap.

sonic_platform_base/chassis_base.py
padmanarayana: Agreed that this is reboot cause and not type.
REBOOT_CAUSE_POWER_LOSS can be changed as REBOOT_CAUSE_POWER_ON_RESET (which is typically the default if FPGA is unable to figure the reason)..

jleveque: @padmanarayana: What do you mean by "FPGA is unable to figure the reason"?

If the hardware is unable to determine a hardware-related reboot cause, the assumption should be that the machine was rebooted by software (i.e., `sudo reboot`). In this case, this method would return `REBOOT_CAUSE_SOFTWARE` (or `REBOOT_CAUSE_NA` if we change the name). I feel this is the scenario you are describing.

On the other hand, `REBOOT_CAUSE_POWER_LOSS` would be returned only if the previous reboot was caused by a hard power loss (i.e., device was unplugged from A/C).

sonic_platform_base/watchdog_base.py
padmanarayana: Thanks

sonic_platform_base/watchdog_base.py
jleveque: Return value is now an integer.

PR #18, opened 2018-12-04
[sfp_util] open eeprom files with buffering disabled

yxieca:
Buffering will cause read to read more than requested. For slow files
like SFP EEPROM, we want to read exactly what we asked for.

Signed-off-by: Ying Xie <ying.xie@microsoft.com>

### DIFF
--- a/sonic_sfp/sfputilbase.py
+++ b/sonic_sfp/sfputilbase.py
@@ -198,7 +198,7 @@ class SfpUtilBase(object):
             return False
         else:
             try:
-                sysfsfile = open(sysfs_sfp_i2c_client_eeprompath, "rb")
+                sysfsfile = open(sysfs_sfp_i2c_client_eeprompath, mode="rb", buffering=0)
                 sysfsfile.seek(offset)
                 sysfsfile.read(1)
             except IOError:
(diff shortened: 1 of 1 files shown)


sonic_sfp/sfputilbase.py
```diff
@@ -198,7 +198,7 @@ def _sfp_eeprom_present(self, sysfs_sfp_i2c_client_eeprompath, offset):
             return False
         else:
             try:
-                sysfsfile = open(sysfs_sfp_i2c_client_eeprompath, "rb")
```
qiluo-msft: Suggest open with 'with'.
[https://www.pythonforbeginners.com/files/with-statement-in-python](https://www.pythonforbeginners.com/files/with-statement-in-python)

sonic_sfp/sfputilbase.py
```diff
@@ -198,7 +198,7 @@ def _sfp_eeprom_present(self, sysfs_sfp_i2c_client_eeprompath, offset):
...
+                sysfsfile = open(sysfs_sfp_i2c_client_eeprompath, mode="rb", buffering=0)
```
qiluo-msft: >open(sysfs_sfp_i2c_client_eeprompath, mode="rb", buffering=0) [](start = 28, length = 61)

Encapsulate a function like: self.open_sysfsfile()

sonic_sfp/sfputilbase.py
yxieca: Done changing this place. thanks!

sonic_sfp/sfputilbase.py
yxieca: I looked at the code structure. It makes more sense to create a help function to return opened file and validate an offset at the same time. Because every place this function is called, there is another file opening following.

PR #20, opened 2019-02-01
Add basic QSFP-DD/OSFP support for sfputil

zzhiyuan:
Enable decode of QSFP-DD/OSFP transceivers, and vendor information.

### DIFF
--- a/sonic_sfp/inf8628.py
+++ b/sonic_sfp/inf8628.py
@@ -1,0 +1,149 @@
+#! /usr/bin/env python
+#----------------------------------------------------------------------------
+# QSFP-DD 8X Transceiver (QSFP Double Density)
+#----------------------------------------------------------------------------
+
+from __future__ import print_function
+
+try:
+    from sff8024 import type_of_transceiver
+    from sffbase import sffbase
+except ImportError as e:
+    raise ImportError ("%s - required module not found" % e)
+
+class inf8628InterfaceId(sffbase):
+
+    def decode_revision_compliance(self, eeprom_data, offset, size):
+        return '%c.%c' % (eeprom_data[offset][0], eeprom_data[offset][1])
+
+    def decode_module_state(self, eeprom_data, offset, size):
+        module_state_byte = eeprom_data[offset]
+        module_state = int(module_state_byte, 16) & 14
+        if module_state == 2:
+            return 'Low Power State'
+        elif module_state == 4:
+            return 'Power Up State'
+        elif module_state == 6:
+            return 'Ready State'
+        elif module_state == 8:
+            return 'Power Down State'
+        elif module_state == 10:
+            return 'Fault State'
+        return 'Unknown State %s' % module_state
+
+    version = '1.0'
+
+    interface_id = {
+            'Identifier':
+                {'offset': 0,
+                 'size': 1,
+                 'type': 'enum',
+                 'decode': type_of_transceiver},
+            'Revision Compliance':
+                {'offset': 1,
+                 'type': 'func',
+                 'decode': {'func': decode_revision_compliance}},
+            'Module State':
+                {'offset': 3,
+                 'type': 'func',
+                 'decode': {'func': decode_module_state}},
+            'Interrupt Asserted':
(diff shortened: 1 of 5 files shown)


zzhiyuan:
> Indentation is currently inconsistent. Please unify indentation to multiples of 4 spaces.

Fixed. Please check again and let me know if there are places I've missed. Thanks

sonic_sfp/inf8628.py
```diff
@@ -0,0 +1,149 @@
...
+        return sffbase.parse(self, self.vendor_rev, rev_raw_data, start_pos)
+
+    def parse_vendor_pn(self, pn_raw_data, start_pos):
+        return sffbase.parse(self, self.vendor_pn, pn_raw_data, start_pos)
+
+    def parse_vendor_sn(self, sn_raw_data, start_pos):
+        return sffbase.parse(self, self.vendor_sn, sn_raw_data, start_pos)
+
+    def dump_pretty(self):
+        if self.interface_data == None:
```
jleveque: `if self.interface_data is None:`

sonic_sfp/inf8628.py
```diff
@@ -0,0 +1,149 @@
...
+            {'offset': 0,
+             'size': 16,
+             'type': 'str'}
+        }
+
+    def __init__(self, eeprom_raw_data=None):
+        self.interface_data = None
+        start_pos = 0
+
+        if eeprom_raw_data != None:
```
jleveque: `if eeprom_raw_data is not None:`

sonic_sfp/inf8628.py
zzhiyuan: > if self.interface_data is None:

Done

sonic_sfp/inf8628.py
zzhiyuan: > if eeprom_raw_data is not None:

Done

keboliu:
@zzhiyuan  would you please share the doc of  INF-8628? I need to implement an extension to get more information from the SFP eeprom, and since inf-8628 support was added so the extension should also cover it, however, I only can find a pdf doc containing the cover of in-8628 but without content, so want to know whether you can share.

zzhiyuan:
> @zzhiyuan would you please share the doc of INF-8628? I need to implement an extension to get more information from the SFP eeprom, and since inf-8628 support was added so the extension should also cover it, however, I only can find a pdf doc containing the cover of in-8628 but without content, so want to know whether you can share.

http://www.qsfp-dd.com/wp-content/uploads/2018/09/QSFP-DD-mgmt-rev3p0-final-9-18-18-Default-clean.pdf

keboliu:
@zzhiyuan thank you very much!

keboliu:
@zzhiyuan Thanks for initiating the work for QSFP-DD/OSFP, want to understand whether you have a plan to implement the DOM parsing for QSFP-DD/OSFP? Also, want to get @lguohan and @jleveque 's view.

zzhiyuan:
> @zzhiyuan Thanks for initiating the work for QSFP-DD/OSFP, want to understand whether you have a plan to implement the DOM parsing for QSFP-DD/OSFP? Also, want to get @lguohan and @jleveque 's view.

Currently, I don't have plan on implementing additional parsing. This PR was enough to recognize the type of transceiver and the vendor information. I am not sure about the scope of information needed for the rest of SONiC.

Thanks

keboliu:
> > @zzhiyuan Thanks for initiating the work for QSFP-DD/OSFP, want to understand whether you have a plan to implement the DOM parsing for QSFP-DD/OSFP? Also, want to get @lguohan and @jleveque 's view.
> 
> Currently, I don't have plan on implementing additional parsing. This PR was enough to recognize the type of transceiver and the vendor information. I am not sure about the scope of information needed for the rest of SONiC.
> 
> Thanks

@zzhiyuan for the scope of DOM info needed for SONiC you can refer to this doc: https://github.com/Azure/SONiC/blob/master/doc/xrcvd/transceiver-monitor-hld.md#112-transceiver-dom-sensor-table

PR #22, opened 2019-02-14
[sonic_sfp] Extend sfpbase to get more info from SFP eeprom for xcvrd

keboliu:
```
sonic_sfp: Extend sfputilbase to support xcvrd get more fields from SFP eeprom.

* Extends sff8446 and sff8472 with more fields parsing functions.
* Extend sfputilbase to return more fields to xcvrd.
* Add handling for OSFP. 
  
  Signed-off-by: Kebo Liu kebol@mellanox.com

```

### DIFF
--- a/sonic_sfp/sff8436.py
+++ b/sonic_sfp/sff8436.py
@@ -493,6 +493,9 @@ class sff8436InterfaceId(sffbase):
 
     def parse_vendor_date(self, sn_raw_data, start_pos):
         return sffbase.parse(self, self.vendor_date, sn_raw_data, start_pos)
+    
+    def parse_vendor_oui(self, sn_raw_data, start_pos):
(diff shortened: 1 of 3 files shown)


jleveque:
Please fix conflicts.

keboliu:
do not merge first, need to handle new added  OSFP.

keboliu:
@jleveque  conflicts fixed and added special treatment for OSFP to avoid xcvrd fail. OSFP transceiver info memory map not fully defined in newly added inf8628.py,  it's also lack of DOM memory map definition. Since I don't have a platform that supports OSFP, I would like to add the fully support for it later or someone can also add it.

keboliu:
@lguohan joe requesting you for a review, would you please take a look.

liat-grozovik:
@lguohan can you please review?

PR #23, opened 2019-02-28
changes needed for SNMP docker upgrade to stretch build

samaity:
Signed-off-by: Sangita Maity [samaity@linkedin.com](mailto:samaity@linkedin.com)

**- What I did**

I was getting lots of import error during SNMP docker upgrade to stretch build. Made changes accordingly to fix those issues. May be python compatibility issue as by default it is using python3 and there's no exceptions module in python3

**- How I did it**
Made it compatible with python3 whenever required.

**- How to verify it**
Use this PR with snmp docker upgrade PR in sonic-buildimage.

### DIFF
--- a/sonic_eeprom/eeprom_base.py
+++ b/sonic_eeprom/eeprom_base.py
@@ -13,6 +13,9 @@ from __future__ import print_function
 
 try:
     import exceptions
+except ImportError:
+    import builtins as exceptions
+try:
     import binascii
     import optparse
     import os
--- a/sonic_eeprom/eeprom_base.py
+++ b/sonic_eeprom/eeprom_base.py
@@ -12,9 +12,9 @@
 from __future__ import print_function
 
 try:
-    import exceptions
+    import exceptions			# Python 2
 except ImportError:
-    import builtins as exceptions
+    import builtins as exceptions	# Python 3
 try:
     import binascii
     import optparse
--- a/sonic_eeprom/eeprom_base.py
+++ b/sonic_eeprom/eeprom_base.py
@@ -12,9 +12,9 @@
 from __future__ import print_function
 
 try:
-    import exceptions			# Python 2
+    import exceptions              # Python 2
 except ImportError:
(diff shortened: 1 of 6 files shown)


sonic_eeprom/eeprom_base.py
```diff
@@ -13,6 +13,9 @@
 
 try:
     import exceptions
+except ImportError:
+    import builtins as exceptions
```
jleveque: Suggest to add comments here for posterity's sake:

```
try:
    import exceptions              # Python 2
except ImportError:
    import builtins as exceptions  # Python 3
```

sonic_eeprom/eeprom_dts.py
```diff
@@ -2,8 +2,11 @@
 # Copyright 2012 Cumulus Networks LLC, all rights reserved
 
 try:
-    import os
     import exceptions
+except ImportError:
+    import builtins as exceptions
```
jleveque: Same as above

sonic_sfp/sff8436.py
```diff
@@ -16,7 +16,7 @@
     import types
     from math import log10
     from sff8024 import type_of_transceiver
-    from sffbase import sffbase
+    from .sffbase import sffbase
```
jleveque: Why are these dots necessary now? What was the error you were seeing?

samaity: I was getting import error like **module not found**.

sonic_eeprom/eeprom_base.py
samaity: Added comments for all the modifications behind both python 2 and python 3 support

sonic_sfp/sff8436.py
samaity: Dot module supports both Python 2 and Python 3 using explicit relative import methods. Added comments wherever needed

lguohan:
please resolve conflict as well.

PR #27, opened 2019-04-11
[sonic_platform_base] Add sonic_sfp and sonic_eeprom to sonic_platform_base

keboliu:
[sonic_platform_base]: Add sonic_sfp and sonic_eeprom to sonic_platform_base

* Add a copy of sonic_sfp and sonic_eeprom to sonic_platform_base
* Add sonic_sfp and sonic_eeprom as submodules of sonic_platform_base
* Make original sonic_sfp and sonic_eeprom as symbol link to the new subfolder under sonic_platform_base to avoid duplicate files.

  
  Signed-off-by: Kebo Liu kebol@mellanox.com


henry-ma:
@jleveque could you review and approve this? thanks.

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -0,0 +1,1108 @@
...
+        except:
+            raise
+
+        parse_fmt_port_config_ini = (os.path.basename(porttabfile) == "port_config.ini")
+
+        # Read the porttab file and generate dicts
+        # with mapping for future reference.
+        #
+        # TODO: Refactor this to use the portconfig.py module that now
+        # exists as part of the sonic-config-engine package.
```
qiluo-msft: I notice the TODO is quite reasonable, and there are lots of repeated port_config parser in code base. Can we start take the action?

jleveque: This is a good time to to make the change, because there are open PRs to begin transitioning from the port_config.ini file to a new 'platform.json' file. The logic to do the parsing and also to fallback to the old port_config.ini is all in the portconfig.py module: https://github.com/Azure/sonic-buildimage/pull/3909

PR #30, opened 2019-04-26
[sonic_fw]: Add firmware management api base

mudsut4ke:
Add base class for firmware management utillity (https://github.com/Azure/sonic-buildimage/pull/2824)

Signed-off-by: Wirut Getbamrung wgetbumr@celestica.com

### DIFF
--- a/sonic_fw/fw_manage_base.py
+++ b/sonic_fw/fw_manage_base.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+#
+# fw_manage_base.py
+#
+# Abstract base class for implementing platform-specific
+#  Firmware management functionality for SONiC
+#
+
+try:
+    import abc
+except ImportError as e:
+    raise ImportError(str(e) + " - required module not found")
+
+
+class FwBase(object):
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def get_module_list(self):
+        """
+        Retrieves the list of module that available on the device
+
+        :return: A list : the list of module that available on the device
+        """
+        return []
+
+    @abc.abstractmethod
+    def get_fw_version(self, module_name):
+        """
+        Retrieves the firmware version of module
+
+        :param module_name: A string, module name
+        :return: A Dict, firmware version object
+            - Example of return object of module that doesn't have sub modules
+            {
+                "module_name": "BIOS",
+                "fw_version": "1.0.0"
+                "has_submodule" = False
+            }
+            - Example of return object of module that have sub modules
+            {
+                "module_name": "CPLD",
+                "fw_version": {
+                    "CPLD1" : "1.0.0",
+                    "CPLD2" : "1.1.0"
+                }
+                "has_submodule" = True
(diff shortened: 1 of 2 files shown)


sonic_fw/fw_manage_base.py
```diff
@@ -0,0 +1,63 @@
...
+                "module_name": "BIOS",
+                "fw_version": "1.0.0"
+                "has_submodule" = False
+            }
+            - Example of return object of module that have sub modules
+            {
+                "module_name": "CPLD",
+                "fw_version": {
+                    "CPLD1" : "1.0.0",
+                    "CPLD2" : "1.1.0"
```
lguohan: in this case, i feel we should have CPLD1, CPLD2 as module name, each module will have only one version.

mudsut4ke: In that case,the list of supported module will be ["BIOS", "CPLD1", "CPLD2"] instead of ["BIOS", "CPLD"] and get_fw_version function should return only version string

Is that good?

jleveque:
@mudsut4ke: Is there any reason to keep this as a separate `sonic_fw` package, or would be better to add `get_fw_version()` and `install_firmware()` methods to the ChassisBase and ModuleBase classes `sonic_platform_base`?

mudsut4ke:
@jleveque , Maybe a reason that firmware install tool is installed in sonic platform and cannot be used on pmon docker

jleveque:
@mudsut4ke: Can you elaborate? We would like to consolidate all platform-specific abstraction into the new `sonic_platform_base` package.

sonic_fw/fw_manage_base.py
lguohan: yes

mudsut4ke:
@jleveque , In my case the CPLD upgrade tool installed in base image (https://github.com/Azure/sonic-buildimage/pull/2446)

if you install `sonic_platform_base` package in both the pmon docker and base image, that will cover my requirement

mudsut4ke:
@jleveque , Any comment?

jleveque:
I think it would be best to add this functionality to the ChassisBase and ModuleBase classes in the new platform API, and we can install the package in the base image as well as the pmon container.

sonic_fw/fw_manage_base.py
lguohan: can you update the comments?

lguohan:
@mudsut4ke , any update on this?

mudsut4ke:
@lguohan , I will move this tool to `sonic_platform_base` follow Joe suggesttion

@jleveque , In `ChassisBase` they Identified BIOS/FPGA/CPLD  as component
Should I add new `ComponentBase` ? or Just add the firmware management to `ModuleBase` is enough ?

jleveque:
@mudsut4ke: I think adding the firmware management functions to `ChassisBase` and `ModuleBase` is enough.

mudsut4ke:
Move to https://github.com/Azure/sonic-platform-common/pull/30

PR #35, opened 2019-06-24
[ChassisBase] Make reboot cause constant strings human-readable

jleveque:
- This will make reporting the reboot cause to the user easier, as we can use these strings directly

### DIFF
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -17,14 +17,14 @@ class ChassisBase(device_base.DeviceBase):
     DEVICE_TYPE = "chassis"
 
     # Possible reboot causes
-    REBOOT_CAUSE_POWER_LOSS = "power_loss"
-    REBOOT_CAUSE_THERMAL_OVERLOAD_CPU = "thermal_overload_cpu"
-    REBOOT_CAUSE_THERMAL_OVERLOAD_ASIC = "thermal_overload_asic"
-    REBOOT_CAUSE_THERMAL_OVERLOAD_OTHER = "thermal_overload_other"
-    REBOOT_CAUSE_INSUFFICIENT_FAN = "insufficient_fan"
-    REBOOT_CAUSE_WATCHDOG = "watchdog"
-    REBOOT_CAUSE_HARDWARE_OTHER = "hardware_other"
-    REBOOT_CAUSE_NON_HARDWARE = "non_hardware"
+    REBOOT_CAUSE_POWER_LOSS = "A/C Power Loss"
+    REBOOT_CAUSE_THERMAL_OVERLOAD_CPU = "Thermal Overload: CPU"
+    REBOOT_CAUSE_THERMAL_OVERLOAD_ASIC = "Thermal Overload: ASIC"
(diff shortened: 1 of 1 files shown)


sonic_platform_base/chassis_base.py
```diff
@@ -17,14 +17,14 @@ class ChassisBase(device_base.DeviceBase):
...
-    REBOOT_CAUSE_THERMAL_OVERLOAD_OTHER = "thermal_overload_other"
-    REBOOT_CAUSE_INSUFFICIENT_FAN = "insufficient_fan"
-    REBOOT_CAUSE_WATCHDOG = "watchdog"
-    REBOOT_CAUSE_HARDWARE_OTHER = "hardware_other"
-    REBOOT_CAUSE_NON_HARDWARE = "non_hardware"
+    REBOOT_CAUSE_POWER_LOSS = "A/C Power Loss"
+    REBOOT_CAUSE_THERMAL_OVERLOAD_CPU = "Thermal Overload: CPU"
+    REBOOT_CAUSE_THERMAL_OVERLOAD_ASIC = "Thermal Overload: ASIC"
+    REBOOT_CAUSE_THERMAL_OVERLOAD_OTHER = "Thermal Overload: Other"
+    REBOOT_CAUSE_INSUFFICIENT_FAN = "Insufficient Fan Speed"
```
qiluo-msft: >REBOOT_CAUSE_INSUFFICIENT_FAN [](start = 4, length = 29)

REBOOT_CAUSE_INSUFFICIENT_FAN_SPEED

sonic_platform_base/chassis_base.py
```diff
@@ -17,14 +17,14 @@ class ChassisBase(device_base.DeviceBase):
...
     # Possible reboot causes
-    REBOOT_CAUSE_POWER_LOSS = "power_loss"
-    REBOOT_CAUSE_THERMAL_OVERLOAD_CPU = "thermal_overload_cpu"
-    REBOOT_CAUSE_THERMAL_OVERLOAD_ASIC = "thermal_overload_asic"
-    REBOOT_CAUSE_THERMAL_OVERLOAD_OTHER = "thermal_overload_other"
-    REBOOT_CAUSE_INSUFFICIENT_FAN = "insufficient_fan"
-    REBOOT_CAUSE_WATCHDOG = "watchdog"
-    REBOOT_CAUSE_HARDWARE_OTHER = "hardware_other"
-    REBOOT_CAUSE_NON_HARDWARE = "non_hardware"
+    REBOOT_CAUSE_POWER_LOSS = "A/C Power Loss"
```
qiluo-msft: >A/C  [](start = 31, length = 4)

Why specify 'A/C'?

jleveque: Just to make it more clear to anyone reviewing the logs that the A/C power was lost (i.e., an external power event).

sonic_platform_base/chassis_base.py
jleveque: Done.

sonic_platform_base/chassis_base.py
jleveque: Changed to "Power Loss" to be more generic, in case power was lost due to a defective power supply, not loss of A/C.

PR #37, opened 2019-07-02
sonic_sfp: avoid possible key error in get_physical_to_logical()

ds952811:
This is part of PR#31 of sonic-platform-daemons, there could be a key error
in get_physical_to_logical() in SFP monitor due to a timing issue, and
it will later be recovered automatically.

Signed-off-by: Dante (Kuo-Jung) Su <dante.su@broadcom.com>


ds952811:
For your reference, the PR#31 of sonic-platform-daemons is available as below:

https://github.com/Azure/sonic-platform-daemons/pull/31

PR #47, opened 2019-07-24
Add SSD Health API and generic implementation

andriymoroz-mlnx:
Signed-off-by: Andriy Moroz <c_andriym@mellanox.com>

### DIFF
--- a/sonic_platform_base/sonic_ssd/ssd_generic.py
+++ b/sonic_platform_base/sonic_ssd/ssd_generic.py
@@ -1,0 +1,157 @@
+#
+# ssd_generic.py
+#
+# Generic implementation of the SSD health API
+# SSD models supported:
+#  - InnoDisk
+#  - StorFly
+#  - Virtium
+
+try:
+    import exceptions              # Python 2
+except ImportError:
+    import builtins as exceptions  # Python 3
+try:
+    import re
+    import subprocess
+    from .ssd_base import SsdBase
+except ImportError as e:
+    raise ImportError (str(e) + "- required module not found")
+
+SMARTCTL = "smartctl {} -a"
+INNODISK = "iSmart -d {}"
+VIRTIUM  = "SmartCmd -m {}"
+
+NOT_AVAILABLE = "N/A"
+
+
+class SsdUtil(SsdBase):
+    """
+    Generic implementation of the SSD health API
+    """
+    model = NOT_AVAILABLE
+    serial = NOT_AVAILABLE
+    firmware = NOT_AVAILABLE
+    temperature = NOT_AVAILABLE
+    health = NOT_AVAILABLE
+    ssd_info = NOT_AVAILABLE
+    vendor_ssd_info = NOT_AVAILABLE
+
+    def __init__(self, diskdev):
+        self.vendor_ssd_utility = {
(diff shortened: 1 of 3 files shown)


sonic_platform_base/sonic_ssd/ssd_generic.py
```diff
@@ -0,0 +1,157 @@
...
+        return self.serial
+
+    def get_vendor_output(self):
+        """
+        Retrieves vendor specific data for the given disk device
+
+        Returns:
+            A string holding some vendor specific disk information
+        """
+        return self.vendor_ssd_info
```
sandycelestica: Except the attributes you list. It's better to add "capacity" "P/E cycle" "Bad block" "Remaining time" .

andriymoroz-mlnx: the attributes you suggest are probably specific to InnoDisk SSDs
For example StorFly disks does not have it but provide attribute #168 (NAND Endurance) which initial value is 20000. If compare to P/E from InnoDisk which is 3000 I think StorFly are not 6 times more reliable but rather use different units. That's why I would prefer to show such info with the "--vendor" option
"Bad block" value is also ambiguous. Depending on SSD type (SLC, TLC, MLC) the endurance of flash cells can be different. Of cause manufacturer knows about it and compensate worse endurance with the greater amount of reserved cells. That's why the absolute value of the bad (reallocated) cells does not represent the disk health state. Sometimes it is used to calculate disk health as `((<total number of reserved cells> - <number of reallocated cells> / <total number of reserved cells>)*100`
"Remaining time" (InnoDisk utility calls this parameter Lifespan) is also provided not by all vendors and is very rough estimation. It is highly dependent on disk usage patterns.

Someday we can add daemon to the pmon which will periodically query current disk health and raise alarm once it reaches some threshold.

sonic_platform_base/sonic_ssd/ssd_generic.py
```diff
@@ -0,0 +1,157 @@
...
+        if self.model:
+            model_short = self.model.split()[0]
+            if self.vendor_ssd_utility.has_key(model_short):
+                self.fetch_vendor_ssd_info(diskdev, model_short)
+                self.parse_vendor_ssd_info(model_short)
+            else:
+                # No handler registered for this disk model
+                pass
+        else:
+            # Failed to get disk model
```
liat-grozovik: suggest to add an error message or at least a warning to indicate info failed. otherwise the pass does not indicate it

andriymoroz-mlnx: Updated to return "Unknown" as disk model name. So, the cli output will be
```
root@r-anaconda-10:/home/admin# show platform ssdhealth
Device Model : Unknown
Health       : N/A%
Temperature  : N/AC

```

liat-grozovik: I think that if we define N/A the % or the C should not be printed. Can you change it?

andriymoroz-mlnx: Fixed in CLI utility https://github.com/Azure/sonic-utilities/pull/587#issuecomment-531804187 (utility adds units sign)

PR #50, opened 2019-08-01
[sfputilbase] Extend get_transceiver_change_event return event

keboliu:
Extend the return of the function to make the function can return some system-level event/error.
like the system not ready, the system becomes ready and the system fails.

By getting these new events upper layer apps(xcvrd) will have more awareness of the system status and handle accordingly.   These events are not mandatory,  vendors can choose to not implement this and xcvrd will not be impacted.

### DIFF
--- a/sonic_platform_base/sonic_sfp/sfputilbase.py
+++ b/sonic_platform_base/sonic_sfp/sfputilbase.py
@@ -1253,8 +1253,12 @@ class SfpUtilBase(object):
          event. When timeout is non-zero, the function can also return when the timer expires.
          When timer expires, the return status is True and events is empty.
         :returns: (status, events)
-        :status: Boolean, True if call successful, False if not;
-        :events: dictionary for pysical port index and the SFP status,
-        status='1' represent plug in, '0' represent plug out like {'0': '1', '31':'0'}
+        :status: Boolean, True if call successful and no system level event/error occurred, 
+         False if call not success or system level event/error occurred.
+        :events: dictionary for physical port index and the SFP status,
+         status='1' represent plug in, '0' represent plug out like {'0': '1', '31':'0'}
+         when it comes to system level event/error, the index will be '-1',
+         and status can be 'system_not_ready', 'system_become_ready', 'system_fail',
+         like {'-1':'system_not_ready'}.
         """
         return


sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -1253,8 +1253,12 @@ def get_transceiver_change_event(self, timeout=0):
          event. When timeout is non-zero, the function can also return when the timer expires.
          When timer expires, the return status is True and events is empty.
         :returns: (status, events)
-        :status: Boolean, True if call successful, False if not;
-        :events: dictionary for pysical port index and the SFP status,
-        status='1' represent plug in, '0' represent plug out like {'0': '1', '31':'0'}
+        :status: Boolean, True if call successful and no system level event/error occurred, 
+         False if call not success or system level event/error occurred.
```
qiluo-msft: What is the status value if call successful and system level event occurred?

jleveque: It would be False: "False if call not success **or system level event/error occurred**"

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -1253,8 +1253,12 @@ def get_transceiver_change_event(self, timeout=0):
...
+        :events: dictionary for physical port index and the SFP status,
+         status='1' represent plug in, '0' represent plug out like {'0': '1', '31':'0'}
+         when it comes to system level event/error, the index will be '-1',
+         and status can be 'system_not_ready', 'system_become_ready', 'system_fail',
```
qiluo-msft: >system_become_ready [](start = 48, length = 19)

From design perspective, `system_become_ready` is not necessary. If this event is really happen, the logic could continue run as a normal successful case or a timeout case. Then all the system level events are actually errors in concept.

stephenxs: Hi Qi, 
I admit that in terms of state machine any normal event, like sfp module insertion or removal, can indicate the xcvrd has entered a normal state. However, an explicit notification can make the flow more clear. Without `system_bdcom_ready` we can hardly tell whether the xcvrd is in a normal state, monitoring the sfp modules' state, or waiting for ready.
In this sense, I suggest remaining this state.

PR #54, opened 2019-08-19
[xcvrd] Fix crashing due to missing fields of some module types

zzhiyuan:
- Added default cable_type and cable_length.

From SFF-8436 regarding cable length fields:
"A value of zero means that the Module does not support [this mode] or that the length information must be determined from the Module technology."

It is possible that all cable length fields in eeprom are 0, and this would lead to xcvrd crashing due to fetching of this data. Fallback is added in this case.


zzhiyuan:
> Can you please split this into two pull requests? One for the xcvrd fix, the other for making sonic_eeprom a symlink?

Removed the symlink change in this PR. Will create another PR with that change.

Thanks

PR #56, opened 2019-08-22
update pice_base.py

carycelestica:


sonic_platform_base/pcie_base.py
```diff
@@ -0,0 +1,38 @@
...
+
+try:
+    import abc
+except ImportError as e:
+    raise ImportError (str(e) + " - required module not found")
+
+class PcieBase(object):
+    __metaclass__ = abc.ABCMeta
+    
+    @abc.abstractmethod
```
keboliu: why need "@abc.abstractmethod"?

PR #57, opened 2019-08-27
Platform Driver Development Framework (PDDF): 1) Changes to psu base class (1.0 APIs) to support PDD PSU CLI utils 2) Adding fan_base class to support PDDF fan CLI utils

fk410167:
- Adding changes to psu_base class to support PDDF PSU cli utility. 
- Added fan_base class to support PDDF FAN cli utility.

All these changes are w.r.t 1.0 APIs. These will be migrated to 2.0 APIs in future.

### DIFF
--- a/sonic_fan/fan_base.py
+++ b/sonic_fan/fan_base.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python
+#
+# fan_base.py
+#
+# Abstract base class for implementing platform-specific
+#  FAN control functionality for SONiC
+#
+
+try:
+    import abc
+except ImportError as e:
+    raise ImportError (str(e) + " - required module not found")
+
+class FanBase(object):
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
+    def get_num_fans(self):
+        """
+        Retrieves the number of FANs supported on the device
+
+        :return: An integer, the number of FANs supported on the device
+        """
+        return 0
+
+    @abc.abstractmethod
+    def get_status(self, index):
+        """
+        Retrieves the operational status of FAN defined
+                by index 1-based <index>
+
+        :param index: An integer, 1-based index of the PSU of which to query status
+        :return: Boolean,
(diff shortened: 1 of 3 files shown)


sonic_fan/fan_base.py
```diff
@@ -0,0 +1,91 @@
...
+
+try:
+    import abc
+except ImportError as e:
+    raise ImportError (str(e) + " - required module not found")
+
+class FanBase(object):
+    __metaclass__ = abc.ABCMeta
+
+    @abc.abstractmethod
```
keboliu: can you remove the "abstractmethod" restriction? understood that you are following the convention in other plugin base classes. But with this restriction, it requests all the vendors to implement these functions or at least write some stub functions if they don't want to implement them, otherwise, it will fail when loading the plugins on the vendor's platform. We are in the progress to replace plugins with new platform API, so it's highly possible that some vendors don't want to implement these new plugins, so please consider this.

And I suggest you think about using new platform API instead of adding new plugins,  eventually we will use new platform API, put efforts on adding new plugins is kind of waste in my view.

sonic_psu/psu_base.py
```diff
@@ -47,3 +47,75 @@ def get_psu_presence(self, index):
         """
         return False
 
+    def get_model(self, idx):
```
keboliu: you may align the method defied in new platform API, if eventually you still want to go with new plugins.

keboliu:
> keboliu,
> 
> Please visit the PR: #62
> I have incorporated your comments here.

can you close this PR since you created a new one?

fk410167:
Please visit the PR: #62
I have incorporated the feedback comments there.

sonic_fan/fan_base.py
jleveque: Agree with @keboliu -- plugins have been deprecated in favor of the new object-oriented platform 2.0 API. Developing new plugins seems unnecessary as they should no longer be used by platform developers.

PR #62, opened 2019-09-12
Platform Driver Development Framework (PDDF): 1) Changes to psu base class (1.0 APIs) to support PDDF CLI utils, 2) Adding fan_base class to support PDDF fan CLI utils (comments incorporated)

fk410167:
* Adding changes to psu_base class to support PDDF PSU cli utility.
* Added fan_base class to support PDDF FAN cli utility.

All these changes are w.r.t 1.0 APIs. These will be migrated to 2.0 APIs in future. Comments to 
https://github.com/Azure/sonic-platform-common/pull/57
are addressed in this PR.

### DIFF
--- a/sonic_fan/fan_base.py
+++ b/sonic_fan/fan_base.py
@@ -66,8 +66,8 @@ class FanBase(object):
 
     def get_speed_rear(self, index):
         """
-        Retrieves the speed of a rear FAN in the tray in revolutions per minute defined
-                by 1-based index <index>
+        Retrieves the speed of a rear FAN in the tray (applicable only for 2-fan tray) 
+                in revolutions per minute defined by 1-based index <index>
 
         :param index: An integer, 1-based index of the FAN of which to query speed
         :return: integer, denoting rear FAN speed
--- a/sonic_fan/fan_base.py
+++ b/sonic_fan/fan_base.py
@@ -1,0 +1,84 @@
+#!/usr/bin/env python
+#
+# fan_base.py
+#
+# Base class for implementing platform-specific
+#  FAN control functionality for SONiC
+#
+
+try:
+    import abc
+except ImportError as e:
+    raise ImportError (str(e) + " - required module not found")
+
+class FanBase(object):
+    __metaclass__ = abc.ABCMeta
+
+    def get_num_fans(self):
+        """
+        Retrieves the number of FANs supported on the device
+
+        :return: An integer, the number of FANs supported on the device
+        """
+        return 0
+
+    def get_status(self, index):
+        """
+        Retrieves the operational status of FAN defined
+                by index 1-based <index>
+
+        :param index: An integer, 1-based index of the PSU of which to query status
+        :return: Boolean,
+            - True if FAN is running with some speed 
+            - False if FAN has stopped running
+        """
+        return False
+
+    def get_presence(self, index):
+        """
+        Retrieves the presence status of a FAN defined
+                by 1-based index <index>
+
+        :param index: An integer, 1-based index of the FAN of which to query status
+        :return: Boolean, True if FAN is plugged, False if not
+        """
+        return False
+    
+    def get_direction(self, index):
+        """
+        Retrieves the airflow direction of a FAN defined
+                by 1-based index <index>
+
+        :param index: An integer, 1-based index of the FAN of which to query status
+        :return: string, denoting FAN airflow direction
+        """
+        return ""
+    
+    def get_speed(self, index):
+        """
+        Retrieves the speed of a Front FAN in the tray in revolutions per minute defined
+                by 1-based index <index>
+
+        :param index: An integer, 1-based index of the FAN of which to query speed
+        :return: integer, denoting front FAN speed
+        """
+        return 0
+
+    def get_speed_rear(self, index):
+        """
+        Retrieves the speed of a rear FAN in the tray in revolutions per minute defined
+                by 1-based index <index>
+
+        :param index: An integer, 1-based index of the FAN of which to query speed
+        :return: integer, denoting rear FAN speed
+        """
+        return 0
+    
+    def set_speed(self, val):
+        """
+        Sets the speed of all the FANs to a value denoted by the duty-cycle percentage val
+
+        :param val: An integer, <0-100> denoting FAN duty cycle percentage 
+        :return: Boolean, True if operation is successful, False if not
+        """
+        return False
--- a/sonic_psu/psu_base.py
+++ b/sonic_psu/psu_base.py
@@ -110,7 +110,7 @@ class PsuBase(object):
         """
         return 0
 
-    def get_fan_rpm(self, idx, fan_idx):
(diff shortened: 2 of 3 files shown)


sonic_fan/fan_base.py
```diff
@@ -0,0 +1,84 @@
...
+
+        :param index: An integer, 1-based index of the FAN of which to query speed
+        :return: integer, denoting front FAN speed
+        """
+        return 0
+
+    def get_speed_rear(self, index):
+        """
+        Retrieves the speed of a rear FAN in the tray in revolutions per minute defined
+                by 1-based index <index>
```
keboliu: my understanding is that although there could have multi fans(usually 2) in a tray, but each fan has its own tachometer, so these fans in the same tray should still be treated as separated fans, In my view, this function is not needed.

sonic_psu/psu_base.py
```diff
@@ -47,3 +47,75 @@ def get_psu_presence(self, index):
...
+    def get_output_power(self, idx):
+        """
+        Retrieves the output power in micro watts of a power supply unit (PSU) defined
+                by 1-based index <idx>
+        :param idx: An integer, 1-based index of the PSU of which to query o/p power
+        :return: An integer, value of o/p power in micro Watts if PSU is good, else zero
+        """
+        return 0
+
+    def get_fan_rpm(self, idx, fan_idx):
```
keboliu: "get_fan_speed" ?

sonic_psu/psu_base.py
```diff
@@ -47,3 +47,75 @@ def get_psu_presence(self, index):
...
+        :return: An integer, value of o/p current in mA if PSU is good, else zero
+        """
+        return 0
+
+    def get_output_power(self, idx):
+        """
+        Retrieves the output power in micro watts of a power supply unit (PSU) defined
+                by 1-based index <idx>
+        :param idx: An integer, 1-based index of the PSU of which to query o/p power
+        :return: An integer, value of o/p power in micro Watts if PSU is good, else zero
```
keboliu: the return value should be a float?

sonic_psu/psu_base.py
```diff
@@ -47,3 +47,75 @@ def get_psu_presence(self, index):
...
+    def get_model(self, idx):
+        """
+        Retrieves the model number/name of a power supply unit (PSU) defined
+                by 1-based index <idx>
+        :param idx: An integer, 1-based index of the PSU of which to query model number
+        :return: String, denoting model number/name
+        """
+        return ""
+
+    def get_mfr_id(self, idx):
```
keboliu: can you give an example string for this id? Not sure the difference between mfr_id and serial number.

sonic_fan/fan_base.py
fk410167: While mentioning the platform inventory, one fan-tray is usually considered as one fan unit.  ODM too, provide the APIs to retrieve 'rear-fan' speed. 'lm-sensors' (or sensors command) denotes rear fan of FAN tray 1 as fan11, rear fan of FAN tray 2 as fan22 etc.  

Open Config model FAN related CLIs also provide the fan speed in term of front and rear speed. 

Set speed FAN APIs also set speed of both the fans in a single call. Hence a separate API is provided for rear FAN speed.

sonic_psu/psu_base.py
fk410167: Since fan speed unit is RPM, this name is chosen for the API. I can change this to 'get_fan_speed' if you think that represents the objective better.

sonic_psu/psu_base.py
fk410167: mfr_id is the 'Manufacturer's ID'. Either name or some abbreviation which denotes the manufacturer. e.g. 
1. On Delta platforms, PSU mfr_id is 'DELTA'
2. On some Accton platforms, PSU mfr_id is '3Y POWER'
3. On soem Quanta platforms, PSU mfr_id is 'ARTESYN'

sonic_psu/psu_base.py
fk410167: Correct. However, psu driver takes care of converting the float value into milli or micro unit depending upon the required granularity. The SysFS interface reads the converted values.

sonic_psu/psu_base.py
keboliu: if you look at the function names in fan_base, they are using 'speed', I would suggest using 'speed' for consistency.

sonic_fan/fan_base.py
keboliu: on some vendors platform, they do treat fans in the same tray as separated fans and hide the rear/front info from user, in this case, this function is not applicable.

fk410167: Hi Keboliu,
If some vendors treat 'rear' fans as separate, indexing them is left to discretion of the vendors. e.g Accton uses,
fan1 (front) --------- fan11 (rear)
fan2 (front) --------- fan12(rear) etc

Some other platform may use fan1 (tray1-front), fan2(tray1-rear), fan3 (tray2-front), fan4 (tray2-rear) etc.
 
If separate indexing is being used, vendors can use 'get_speed' API. If tray is numbered and fans inside a tray are mentioned as 'front' and 'rear' then let 'get_speed_rear' API be used. Since some vendors also use tray numbering, can't we keep both APIs?

sonic_psu/psu_base.py
fk410167: Sure. I will update the PR

fk410167:
Changes as per the feedback --- 824881f

Above is the change incorporating the feedbacks from keboliu.

sonic_fan/fan_base.py
keboliu: I am ok with keeping both APIs.

sonic_psu/psu_base.py
keboliu: what's your decision on return values? Still in integer or change to float?

sonic_fan/fan_base.py
jleveque: I think this function may require more thorough commenting to ensure vendors understand that `get_speed_rear()` is only to be implemented  for the rear fan of a 2-fan tray.

sonic_psu/psu_base.py
fk410167: We have decided to go with integer values here with milli volts, amps etc granularity.

sonic_fan/fan_base.py
fk410167: Added extra comment to mention the same.

fk410167:
@fk410167
Improving the comments section of get_speed_rear() API -------- ee58546

Added extra comment explaining the usage of get_speed_rear()

fk410167:
Thanks keboliu for approving. 

sonic-buildimage PR ( https://github.com/Azure/sonic-buildimage/pull/3387/ ) has dependency over this PR ( https://github.com/Azure/sonic-platform-common/pull/62 ) and sonic-utilities PR ( https://github.com/Azure/sonic-utilities/pull/624 ).

PR #63, opened 2019-10-02
[sfputilbase] Fix xcvr and sfpshow crash with OSFP

pphuchar:
In __get_transceiver_dom_threshold_info_dict()__ ad a case to handle osfp port type.
In get_transceiver_info_dict() update __specification_compliance__ from N/A to empty dictionary.
This prevents error in _sfpshow_ trying to prase data from STATE_DB with _eval_ function.


jleveque:
@keboliu: Please review.

PR #65, opened 2019-11-12
[sonic_platform_base] Reallocate empty lists in the constructor of the base class

ds952811:
… the base class

When we use a mutable object as a default value in the base class, a value that can be modified in place in the subclasses, such as the empty lists of
the chassis base class, and it is most likely unwanted.

And hence, this commit will try to address this issue by allocating new empty
list objects at the constructor of the base class, and the subclass should
rely on the constructor of superclass for this.

As a result, it simplifies the source code, and reduce the maintenance efforts

Signed-off-by: Dante (Kuo-Jung) Su <dante.su@broadcom.com>

### DIFF
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -28,33 +28,41 @@ class ChassisBase(device_base.DeviceBase):
 
     # List of ComponentBase-derived objects representing all components
     # available on the chassis
-    _component_list = []
+    _component_list = None
 
     # List of ModuleBase-derived objects representing all modules
     # available on the chassis (for use with modular chassis)
-    _module_list = []
+    _module_list = None
 
     # List of FanBase-derived objects representing all fans
     # available on the chassis
-    _fan_list = []
+    _fan_list = None
 
     # List of PsuBase-derived objects representing all power supply units
     # available on the chassis
-    _psu_list = []
+    _psu_list = None
 
     # List of ThermalBase-derived objects representing all thermals
     # available on the chassis
-    _thermal_list = []
+    _thermal_list = None
 
     # List of SfpBase-derived objects representing all sfps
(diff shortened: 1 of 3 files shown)


ds952811:
Here is an example to help you understand what's exactly the problem we try to address.
Please note we have 2 different results of FAN number in the following test, and it's because the LIST is a mutable object, and it's modified in place in the subclasses

root@sonic:~# cat test.py
#!/usr/bin/env python

import sonic_platform.platform

chassis = sonic_platform.platform.Platform().get_chassis()
if chassis is None:
    print("Error getting chassis!")
    sys.exit(1)

print("FAN number: {}".format(chassis.get_num_fans()))

chassis = sonic_platform.platform.Platform().get_chassis()
if chassis is None:
    print("Error getting chassis!")
    sys.exit(1)

print("FAN number: {}".format(chassis.get_num_fans()))

root@sonic:~# ./test.py
FAN number: 6
FAN number: 12
root@sonic:~#

jleveque:
So just to confirm, you're claiming that with the current code, after the second call to `sonic_platform.platform.Platform().get_chassis()` you will have double the number of devices (int this case, fans) because you have populated the global lists twice?

ds952811:
> So just to confirm, you're claiming that with the current code, after the second call to `sonic_platform.platform.Platform().get_chassis()` you will have double the number of devices (int this case, fans) because you have populated the global lists twice?

Yes, as the fan list of the base class will be modified in place, so the number of devices will be double at 2nd call, and triple at 3rd call.

The same issues apply to all the LIST objects in the base classes (As they're mutable objects)

ds952811:
Here is a good article to help you have a better understanding of this issue 
https://blog.florimond.dev/python-mutable-defaults-are-the-source-of-all-evil

PR #67, opened 2019-11-21
[sonic_platform_common/fan]add fan direction "not applicable"

stephenxs:
[sonic_platform_common/fan]add fan direction "not applicable"
it is for the devices that don't support fetching fan's direction for now

### DIFF
--- a/sonic_platform_base/fan_base.py
+++ b/sonic_platform_base/fan_base.py
@@ -18,6 +18,7 @@ class FanBase(device_base.DeviceBase):
     # Possible fan directions (relative to port-side of device)
     FAN_DIRECTION_INTAKE = "intake"
     FAN_DIRECTION_EXHAUST = "exhaust"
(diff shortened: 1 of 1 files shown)


sonic_platform_base/fan_base.py
```diff
@@ -18,6 +18,7 @@ class FanBase(device_base.DeviceBase):
     # Possible fan directions (relative to port-side of device)
     FAN_DIRECTION_INTAKE = "intake"
     FAN_DIRECTION_EXHAUST = "exhaust"
+    FAN_DIRECTION_NOT_APPLICABLE = "not applicable"
```
jleveque: Maybe make the string `N/A` to be more brief?

stephenxs: > Maybe make the string `N/A` to be more brief?

Updated.

PR #68, opened 2019-11-29
DeviceBase.get_name should raise NotImplementedError like other membe…

Junchao-Mellanox:
Fix issue: DeviceBase.get_name does not raise NotImplementedError like other members.


keboliu:
@jleveque I think we need this fix, what's your view?

jleveque:
@keboliu: Yes, looks like this line was missed.

PR #70, opened 2019-12-07
Fix EEPROM vendor extension TLV number counting issue

keboliu:
TlvInfoDecoder class is using a member variable _TLV_NUM_VENDOR_EXT to counting the vendor extension TLV numbers in the system EEPROM, each time function update_eeprom_db be called, _TLV_NUM_VENDOR_EXT will be accumulated, which will cause EEPROM extension TLV number not correct after an update. 

The fix is to replace this member variable with a local variable in the function update_eeprom_db, the local variable is set to 0 before start to count the TLV,  this will eliminate this issue

revised formatting by adding extra blank lines to align with PEP8


keboliu:
@jleveque would you please help to label this fix for **bmtor** and **201911** branch?

jleveque:
@keboliu: We have not yet created either of those branches in this repo.

PR #73, opened 2019-12-27
Support thermal policy management including policy file loading, information collecting, condition matching and action executing policy

Junchao-Mellanox:
What I did?

- Thermal policy is a feature to allow SONiC user control switch's behavior via a JSON configuration file. For example, user can configure that when any FAN absence, all other FAN should speed to 100%. Thermal policy feature divide such use case into three concepts: info, condition and action.

- Info represents some kind of status of the switch, the default info of mellanox platform are FAN info and PSU info. Info object collect switch status and wait for thermal policy to "consume" it. Info types are defined in JSON, each info type must correspond to a ThermalPolicyInfoBase-derived class. Vendo can define their own info class.

- Condition consume Info objects and determine if a certain event happened. In previous example, the condition is "any FAN absence". Condition types are defined in JSON, each condition type must correspond to a ThermalPolicyConditionBase-derived class. Thermal manager create and initialize condition objects while loading JSON file.

- Action object execute something when conditions match. In previous example, the action is "set other FAN speed to 100%". Action types are defined in JSON, each action type must correspond to a ThermalPolicyActionBase-derived class. Thermal manager create and initialize action objects while loading JSON file.

A thermal policy file can define arbitrary number of info and policy. Each policy can define arbitrary number of conditions and actions. If all conditions of a policy match, all actions of the policy will be executed.


setup.py
```diff
@@ -19,6 +19,7 @@
         'sonic_platform_base.sonic_ssd',
         'sonic_psu',
         'sonic_sfp',
+        'sonic_platform_base.sonic_thermal_control'
```
jleveque: Please move this line below `sonic_platform_base.sonic_ssd` to retain alphabetical order

sonic_platform_base/sonic_thermal_control/thermal_action_base.py
```diff
@@ -0,0 +1,17 @@
+from .thermal_json_object import ThermalJsonObject
+
+
+class ThermalPolicyActionBase(ThermalJsonObject):
+    """
+    Base class for thermal action. Once all thermal conditions in a thermal policy are matched,
+    all predefined thermal action would be executed.
```
jleveque:
```suggestion
    all predefined thermal actions will be executed.
```

sonic_platform_base/sonic_thermal_control/thermal_policy.py
```diff
@@ -0,0 +1,67 @@
+from .thermal_json_object import ThermalJsonObject
+
+
+class ThermalPolicy(object):
+    """
+    Class representing a thermal policy. A thermal policy object is initialized by policy.json.
```
jleveque: There are multiple comments in this PR which refer to the file "policy.json". However, this API allows the user to specify the path to the file in the `load()` function, thus the filename can differ. Consider changing these comments from "policy.json" to "JSON policy file".

jleveque:
Please fix new conflicts.

liat-grozovik:
@stepanblyschak please review as well

keboliu:
@jleveque @liat-grozovik my view is that this PR is ready to merge, would you please help to proceed?

abdosi:
this is merged via direct PR for 201911

PR #78, opened 2020-03-11
[Mellanox] Add validation for thermal policy 

Junchao-Mellanox:
Add validation for thermal policy file:

1. validate duplicate/empty thermal condition
2. validate duplicate/empty thermal policy
3. validate thermal policy with same condition but different actions


abdosi:
@Junchao-Mellanox is this still needed for 201911. I assumed it is already merge via direct PR 201911.

@rlhui @liat-grozovik

Junchao-Mellanox:
It is already merge via direct PR 201911.

PR #82, opened 2020-04-13
[component]: Introduce new firmware management API

nazariig:
Signed-off-by: Nazarii Hnydyn <nazariig@mellanox.com>

**Summary:**
The purpose of this PR is to propagate some new component firmware management APIs which are intended to make the process more versatile. These new API's will be used mostly to extend `fwutil` functionality in order to support new platform components (e.g, SSD).

**Motivation:**
1. Inform user in advance about required component firmware update actions when operating in manual mode
2. Allow `fwutil` to read component firmware version directly from image instead of configuration file if not provided

### DIFF
--- a/sonic_platform_base/component_base.py
+++ b/sonic_platform_base/component_base.py
@@ -51,7 +51,7 @@ class ComponentBase(object):
         """
         raise NotImplementedError
 
-    def get_update_notification(self, image_path):
+    def get_firmware_update_notification(self, image_path):
         """
         Retrieves a notification on what should be done in order to complete
         the component firmware update
@@ -90,18 +90,5 @@ class ComponentBase(object):
 
         Raises:
             RuntimeError: update failed
-        """
-        raise NotImplementedError
-
-    def is_delayed_firmware_update_supported(self, image_path):
-        """
-        Retrieves a value indicating whether immediate actions are required
-        to complete the component firmware update (e.g., reboot, power cycle, etc.)
-
-        Args:
-            image_path: A string, path to firmware image
-
-        Returns:
-            A boolean, True if delayed firmware update supported, False if not
         """
         raise NotImplementedError
--- a/sonic_platform_base/component_base.py
+++ b/sonic_platform_base/component_base.py
@@ -39,14 +39,69 @@ class ComponentBase(object):
         """
         raise NotImplementedError
 
+    def get_available_firmware_version(self, image_path):
+        """
+        Retrieves the available firmware version of the component
+
+        Args:
+            image_path: A string, path to firmware image
+
+        Returns:
+            A string containing the available firmware version of the component
+        """
+        raise NotImplementedError
+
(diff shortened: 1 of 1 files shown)


sonic_platform_base/component_base.py
```diff
@@ -50,3 +78,30 @@ def install_firmware(self, image_path):
             A boolean, True if install was successful, False if not
         """
         raise NotImplementedError
+
+    def update_firmware(self, image_path):
+        """
+        Updates firmware of the component.
+        It's API's responsibility to complete firmware update
+        in case some extra steps are required (e.g., reboot, power cycle, etc.)
```
jleveque, line 86: Can you add more detail to this docstring and/or the docstring of `install_firmware` above to distinguish the differences in functionality between these two functions? Currently they appear as though they accomplish the same thing.

nazariig: @jleveque The difference between these two is that originally `install_firmware` was implemented to do installation only (we have return value True/False). The update process is a bit different - it assumes not only firmware installation but also all the rest steps which are needed to load the new version. In this case there is no chance to handle the return code, since there might be power cycle or cold reboot as part of update operation. The only reliable way to handle the errors is to throw the exceptions. For this purpose we need a dedicated API - `update_firmware`. A call to this new API will guarantee that firmware update has been completed and no user interaction is required.

The `install_firmware` is still needed to provide user full control on firmware installation process in case delayed update is expected: for example if the power cycle/cold reboot should be done later.

jleveque: @nazariig: Can you update the docstrings of these functions to add some of that detail so whomever is implementing them understands what each does? Currently, the docstrings are very similar; it's not quite clear what the difference is.

nazariig: @jleveque sure. I will extend the docstrings and add more info.

nazariig: @jleveque done. Please review the latest changes.

jleveque: Much better. Thanks!

sonic_platform_base/component_base.py
```diff
@@ -39,9 +39,37 @@ def get_firmware_version(self):
         """
         raise NotImplementedError
 
+    def get_available_firmware_version(self, image_path):
+        """
+        Retrieves the available firmware version of the component
```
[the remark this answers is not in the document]
nazariig: @liat-grozovik sure. I will update the description.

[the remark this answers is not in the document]
nazariig: @liat-grozovik fixed.

abdosi:
@nazariig create PR for 201911 . Cherry-pick is not happening as there is some issue with merge commit

nazariig:
@abdosi Done: https://github.com/Azure/sonic-platform-common/pull/87

PR #85, opened 2020-04-24
[eeprom] Add try-except to catch the IOError

kuanyu99:
What I did:

- Handling the IOError produced during interacting with the eeprom through I/O operation


kuanyu99:
@jleveque Could you help to review, please?

PR #89, opened 2020-05-28
Can't get SFP's DOM information via get_transceiver_dom_info_dict()

chiourung:
The default value of calibration_type is 0:
    def __init__(self, eeprom_raw_data=None, calibration_type=0):
        self._calibration_type = calibration_type
If the calibration_type of the SFP is 1, it wouldn't get the correct DOM information.

What I do:
It should get calibration type defore parse DOM information.

Signed-off-by: chiourung_huang <chiourung_huang@edge-core.com>


jleveque:
@chiourung: Can you please update the description to include an example of the problem you were facing which this fixes?

PR #91, opened 2020-06-08
[chassis_base] Add platform API support for system LED

Junchao-Mellanox:
Why I did this?

To support system health feature, we need get/set system LED status.

How I did this?

Add 2 new platform API to chassis_base.py for vendor to implement:
- set_status_led
- get_status_led

### DIFF
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -387,6 +387,26 @@ class ChassisBase(device_base.DeviceBase):
         """
         raise NotImplementedError
 
+    def get_asic_temperature(self):
+        """
+        Retrieves current temperature reading from ASIC sensor
+
+        Returns:
+            A float number of current temperature in Celsius up to nearest thousandth
+            of one degree Celsius, e.g. 30.125 
+        """
+        raise NotImplementedError
+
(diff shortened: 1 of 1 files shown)


sonic_platform_base/chassis_base.py
```diff
@@ -60,6 +60,9 @@ class ChassisBase(device_base.DeviceBase):
     # Object derived from eeprom_tlvinfo.TlvInfoDecoder indicating the eeprom on the chassis
     _eeprom = None
 
+    # System status LED
+    _led = None
```
jleveque, line 64: Should we make the variable name more descriptive, like `_status_led`?

Junchao-Mellanox: sure

jleveque:
@Junchao-Mellanox: Your description above states that you are adding 4 new API functions, but the PR only contains two. If you only plan to add two, can you please update the description?

Junchao-Mellanox:
sure

jleveque:
Why I did this?

To support system health feature, we need get/set system LED status.

How I did this?

Add 2 new platform API to chassis_base.py for vendor to implement:
- set_status_led
- get_status_led

PR #96, opened 2020-06-21
[sfputil] Make SfpUtilHelper.get_physical_to_logical noexcept as in SfpUtilBase

vboykox:
The SfpUtil interface used in xcvrd expects get_physical_to_logical to return None in the case of unknown port:
https://github.com/Azure/sonic-platform-daemons/blob/master/sonic-xcvrd/scripts/xcvrd#L906

Note SfpUtilBase implementation already conforms the interface in xcvrd:
https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/sonic_sfp/sfputilbase.py#L603


vboykox:
@keboliu @lguohan please review

PR #100, opened 2020-07-09
Sfputil base and helper class changes for multi-ASIC

judyjoseph:
This is for changes needed to support multi-asic platforms. 
(i) Add a new API to parse the port_config.ini files in different directories named based on the asic_ID's 
(ii) Add a new dictionary which maps the interface to the asic_id 
(iii) Check if it is a backplane interface, if so skip from adding it to logical_port list

### DIFF
--- a/sonic_platform_base/sonic_sfp/sfputilbase.py
+++ b/sonic_platform_base/sonic_sfp/sfputilbase.py
@@ -16,6 +16,7 @@ try:
     from natsort import natsorted
     from portconfig import get_port_config
     from sonic_py_common import device_info
+    from sonic_py_common.interface import backplane_prefix
 
     from . import bcmshell       # Dot module supports both Python 2 and Python 3 using explicit relative import methods
     from sonic_eeprom import eeprom_dts
@@ -459,7 +460,7 @@ class SfpUtilBase(object):
                 portname = line.split()[0]
 
                 # Ignore if this is an internal backplane interface
-                if portname.startswith(daemon_base.get_internal_interface_prefix()):
+                if portname.startswith(backplane_prefix()):
                     continue
 
                 bcm_port = str(port_pos_in_file)
@@ -640,10 +641,7 @@ class SfpUtilBase(object):
 
     def get_asic_id_for_logical_port(self, logical_port):
         """Returns the asic_id list of physical ports for the given logical port"""
-        if logical_port in self.logical_to_asic.keys():
-            return self.logical_to_asic[logical_port]
-        else:
-            return None
+        return self.logical_to_asic.get(logical_port)
 
     def is_logical_port_ganged_40_by_4(self, logical_port):
         physical_port_list = self.logical_to_physical[logical_port]
--- a/sonic_platform_base/sonic_sfp/sfputilbase.py
+++ b/sonic_platform_base/sonic_sfp/sfputilbase.py
@@ -31,6 +31,9 @@ except ImportError as e:
 PLATFORM_JSON = 'platform.json'
 PORT_CONFIG_INI = 'port_config.ini'
 
+# TODO, to move this definition to a common place
+INTERNAL_INTERFACE_PREFIX = "Ethernet-BP"
+
 # definitions of the offset and width for values in XCVR info eeprom
 XCVR_INTFACE_BULK_OFFSET = 0
 XCVR_INTFACE_BULK_WIDTH_QSFP = 20
@@ -156,6 +159,9 @@ class SfpUtilBase(object):
     # List of logical port names available on a system
     """ ["swp1", "swp5", "swp6", "swp7", "swp8" ...] """
     logical = []
+
+    # Mapping of logical port names available on a system to ASIC num
+    logical_to_asic = {}
 
     # dicts for easier conversions between logical, physical and bcm ports
     logical_to_bcm = {}
@@ -371,7 +377,7 @@ class SfpUtilBase(object):
 
         return False
 
-    def read_porttab_mappings(self, porttabfile):
+    def read_porttab_mappings(self, porttabfile, asic_inst = 0):
         logical = []
         logical_to_bcm = {}
         logical_to_physical = {}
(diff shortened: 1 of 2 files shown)


sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -20,6 +20,7 @@
     from portconfig import get_port_config
     from collections import OrderedDict
     from natsort import natsorted
+    from sonic_daemon_base import daemon_base
     from sonic_daemon_base.daemon_base import DaemonBase
```
jleveque, line 24: These lines have overlap in their imports. Condense to one import.

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -369,7 +373,7 @@ def _is_valid_port(self, port_num):
 
         return False
 
-    def read_porttab_mappings(self, porttabfile):
+    def read_porttab_mappings(self, porttabfile, asic_inst = 0):
```
jleveque, line 376: Remove spaces around equals sign for default parameter assignments.

sonic_platform_base/sonic_sfp/sfputilhelper.py
```diff
@@ -41,7 +44,7 @@ class SfpUtilHelper(object):
     def __init__(self):
         pass
 
-    def read_porttab_mappings(self, porttabfile):
+    def read_porttab_mappings(self, porttabfile, asic_inst = 0):
```
jleveque, line 47: Remove spaces around equals sign for default parameter assignments.

sonic_platform_base/sonic_sfp/sfputilhelper.py
```diff
@@ -184,3 +207,10 @@ def is_logical_port(self, port):
             return 1
         else:
             return 0
+
+    def get_asicId_for_logical_port(self, logical_port):
```
jleveque, line 211: Function names should be all lowercase: `get_asic_id_for_logical_port`

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -613,6 +637,13 @@ def is_logical_port(self, port):
         else:
             return 0
 
+    def get_asicId_for_logical_port(self, logical_port):
```
jleveque, line 640: Function names should be all lowercase: `get_asic_id_for_logical_port`

sonic_platform_base/sonic_sfp/sfputilhelper.py
jleveque: Looks like this was missed in the most recent commit.

judyjoseph: Thanks Joe, I have updated the space here as well.

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -453,12 +456,16 @@ def read_porttab_mappings(self, porttabfile):
                 # so we use the port's position in the file (zero-based) as bcm_port
                 portname = line.split()[0]
 
+                # Ignore if this is an internal backplane interface
+                if portname.startswith(daemon_base.get_internal_interface_prefix()):
+                    continue
```
smaheshm, line 461: Is json format file depracated. I don't see similar check for json format files.

smaheshm:
What's the difference between sfputilbase and sfputilhelper. There's lot of common code between them.

sonic_platform_base/sonic_sfp/sfputilhelper.py
```diff
@@ -184,3 +207,10 @@ def is_logical_port(self, port):
             return 1
         else:
             return 0
+
+    def get_asic_id_for_logical_port(self, logical_port):
+        """Returns the asic_id list of physical ports for the given logical port"""
+        if logical_port in self.logical_to_asic.keys():
+            return self.logical_to_asic[logical_port]
+        else:
+            return None
```
smaheshm, line 216: this can be simpler: 

return self.logical_to_asic.get(logical_port)

sonic_platform_base/sonic_sfp/sfputilbase.py
judyjoseph: We do support only the port_config.ini files for multi_asic. The json format files was recently introduced. So I was currently adding the logic to handle multiple port_config.ini files alone.

jleveque: All port_config.ini/platform.json parsing should eventually be moved to sonic-py-common so that it is only done in one place.

judyjoseph: @jleveque I could move the json/ini file parsing from here --> to  portconfig.py. But what I find is that I will have to **write this logic as new API's** in  portconfig.py -- as the port_config.ini/platform.json parsing logic implemented here in sfpbase module is different to that is already present @ https://github.com/Azure/sonic-buildimage/blob/da69d57a8dffa37e285d7b93671d055ba193c61d/src/sonic-config-engine/portconfig.py#L142.

We could have separate parsing API's for now, and later on unify them in portconfig.py as a single parsing API ?

jleveque: In that case, It's probably best to leave it here for now and do the move and unification at the same time, and maybe also move it all to sonic-py-common at the same time, as well.

jleveque:
@smaheshm:

> What's the difference between sfputilbase and sfputilhelper. There's lot of common code between them.

sfputilbase.py was the base class for the old platform 1.0 plugins, and is deprecated. sfphelper.py is used as a helper class for the new platform 2.0 API. Hence why there is duplicate code. Once all platforms have moved to using the new 2.0 API, sfputilbase.py will be removed.

smaheshm:
Looks good!

abdosi:
@judyjoseph Please create PR for 201911

judyjoseph:
> @judyjoseph Please create PR for 201911

Raised this new PR https://github.com/Azure/sonic-platform-common/pull/117

PR #102, opened 2020-07-24
[sfp] Add parsing the dom_capability to sff8472

kuanyu99:
- What I did
Following SFF-8472 to read diagnostic monitoring type register.
Use that register to check if diagnostic monitoring function implemented or not.

- How I did it
Check the register position from SFF-8472 document.
Read the register value from the EEPROM.

- How to verify it
Compare one transceiver which support diagnostic monitoring function to another transceiver which doesn't support it.
Run 'show interfaces transceiver eeprom' command.

### DIFF
--- a/sonic_platform_base/sonic_sfp/sfputilbase.py
+++ b/sonic_platform_base/sonic_sfp/sfputilbase.py
@@ -123,6 +123,12 @@ qsfp_compliance_code_tup = ('10/40G Ethernet Compliance Code', 'SONET Compliance
                             'Fibre Channel link length/Transmitter Technology',
                             'Fibre Channel transmission media', 'Fibre Channel Speed')
 
+qsfp_dom_capability_tup = ('Tx_power_support', 'Rx_power_support',
+                           'Voltage_support', 'Temp_support')
+
+# Add an EOL to prevent this being viewed as string instead of list
+sfp_dom_capability_tup = ('sff8472_dom_support', 'EOL')
+
 class SfpUtilError(Exception):
     """Base class for exceptions in this module."""
     pass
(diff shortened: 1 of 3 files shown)


kuanyu99:
Hi @jleveque,
Could you help to review this, please?

jleveque:
@keboliu: Please review.

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -954,6 +968,11 @@ def get_transceiver_info_dict(self, port_num):
                         compliance_code_dict[key] = sfp_interface_bulk_data['data']['Specification compliance']['value'][key]['value']
                 transceiver_info_dict['specification_compliance'] = str(compliance_code_dict)
 
+                for key in qsfp_dom_capability_tup:
+                    if key in sfp_dom_capability_data['data']:
+                        dom_capability_dict[key] = "yes" if sfp_dom_capability_data['data'][key]['value'] == 'on' else "no"
+                transceiver_info_dict['dom_capability'] = str(dom_capability_dict)
```
keboliu, line 974: Is it necessary to expose this to transceiver_info_dict? My understanding is that this is only some internal flag, if DOM supported, means can get the related meaningful monitor info from the EEPROM.

kuanyu99: This is a verification method for the SFP type transceiver for now.
If user buy a SFP type transceiver with DOM capability, there is no way to check DOM capability of the transceiver in SONIC currently. With this, the user can check if he/she buy a fake one or not at least.
And in the future, we can enhance this part to check more dom info according to its dom_capability.

keboliu (approved):

jleveque (approved):

PR #106, opened 2020-08-04
[component] add auto_update_firmware() to support the auto update.

sujinmkang:
To support the component firmware automatic update, add auto_update_firmware() to platform component api
This is to support fwutil auto_update command to update  the platform component firmware automatically.
Fwutil HLD to support auto_update interfaces/commands.
https://github.com/Azure/SONiC/pull/648

### DIFF
--- a/sonic_platform_base/component_base.py
+++ b/sonic_platform_base/component_base.py
@@ -88,7 +88,7 @@ class ComponentBase(object):
         """
         raise NotImplementedError
 
-    def update_firmware(self, image_path):
+    def update_firmware(self, image_path, boot_action):
         """
         Updates firmware of the component
 
@@ -98,6 +98,7 @@ class ComponentBase(object):
 
         Args:
             image_path: A string, path to firmware image
+            boot_action: A string, boot action following the upgrade
 
         Raises:
             RuntimeError: update failed
--- a/sonic_platform_base/component_base.py
+++ b/sonic_platform_base/component_base.py
@@ -120,11 +120,16 @@ class ComponentBase(object):
                          - none/fast/warm/cold
 
         Returns:
-            Output: A string, status and info
-                status: True or False, firmware auto-update status
-                info: The detail information of the firmware auto-update status.
-                      - "updated"/"installed"(which needs a reboot for the firmware to be active)/"scheduled"
-                      - "ns_boot_type/image_error/exec_fail"
+            Output: A return code
+                return_code: An integer number, status of component firmware auto-update
+                    - return code of a positive number indicates successful auto-update
+                        - status_installed = 1
+                        - status_updated = 2
+                        - status_scheduled = 3
+                    - return_code of a negative number indicates failed auto-update
+                        - status_err_boot_type = -1
+                        - status_err_image = -2
+                        - status_err_others = -3
 
         Raises:
             RuntimeError: auto-update failure cause
--- a/sonic_platform_base/component_base.py
+++ b/sonic_platform_base/component_base.py
@@ -104,22 +104,21 @@ class ComponentBase(object):
         """
         raise NotImplementedError
 
-    def auto_update_firmware(self, image_path, boot_action, command):
+    def auto_update_firmware(self, image_path, boot_action):
         """
         Updates firmware of the component
 
         This API performs firmware update automatically based on boot_action: it assumes firmware installation 
-        and creating a loading task in a single call.
+        and/or creating a loading task during a boot action, if needed, in a single call.
         In case platform component requires some extra steps (apart from calling Low Level Utility)
         to load the installed firmware (e.g, reboot, power cycle, etc.) - this will be done automatically during the reboot.
-        The loading task will be created by API and it will be performed by the reboot script.
+        The loading task will be created by API.
 
         Args:
             image_path: A string, path to firmware image
             boot_action: A string, boot action following the upgrade
-            command: A string, command to be performed
 
         Raises:
-            RuntimeError: update failed
+            RuntimeError: auto-update failure cause
         """
         raise NotImplementedError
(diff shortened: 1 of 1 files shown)


sujinmkang:
@padmanarayana can you review this?

sonic_platform_base/component_base.py
```diff
@@ -103,3 +103,22 @@ def update_firmware(self, image_path):
...
+        Updates firmware of the component
+
+        This API performs firmware update automatically based on boot_action: it assumes firmware installation 
+        and/or creating a loading task during a boot action, if needed, in a single call.
+        In case platform component requires some extra steps (apart from calling Low Level Utility)
+        to load the installed firmware (e.g, reboot, power cycle, etc.) - this will be done automatically during the reboot.
+        The loading task will be created by API.
+
+        Args:
+            image_path: A string, path to firmware image
```
padmanarayana, line 118: image_path : can it be mentioned that the firmware_image be a self extracting binary..

rkdevi27, line 118: This API will have any return values? Please mention if required.

padmanarayana: Got clarification from HLD.

sonic_platform_base/component_base.py
```diff
@@ -103,3 +103,28 @@ def update_firmware(self, image_path):
...
+            boot_action: A string, boot action following the upgrade
+
+        Returns:
+            Output: A string, status and info
+                status: True or False, firmware auto-update status
```
nazariig, line 123: @sujinmkang suggest to use status codes:
```
status_installed=1
status_updated=2
status_scheduled=3
```

sonic_platform_base/component_base.py
```diff
@@ -103,3 +103,28 @@ def update_firmware(self, image_path):
             RuntimeError: update failed
         """
         raise NotImplementedError
+
+    def auto_update_firmware(self, image_path, boot_action):
```
nazariig, line 107: @sujinmkang since this API is a part of SONiC firmware upgrade framework, please provide the description on which boot types are supported

sonic_platform_base/component_base.py
sujinmkang: @nazariig the original idea was for status to indicate the auto-update performed status which means performed successfully or not. For info, it gives more detail how the update has been performed. And platform vendor can return any information for the failure that is why the info was open.
But we can define as below. do you have any other failure cause to be added?
------------------------------------------
status           |   info
------------------------------------------
True              |  Installed = 1
True              |  Updated = 2 
True              |  Scheduled = 3
False             |  NS_BootType = -1
False             |  ImageError  = -2
False             |  ExecutionFail (or exception) = -3
--------------------------------------------

padmanarayana (approved):

rkdevi27 (approved):

sujinmkang:
@nazariig can I submit this PR if you don't have any other comment?

sonic_platform_base/component_base.py
```diff
@@ -103,3 +103,35 @@ def update_firmware(self, image_path):
...
+        Args:
+            image_path: A string, path to firmware image
+            boot_action: A string, boot action following the upgrade
+                         - none/fast/warm/cold
+
+        Returns:
+            Output: A return code
+                return_code: An integer number, status of component firmware auto-update
+                    - return code of a positive number indicates successful auto-update
+                        - status_installed = 1
```
nazariig, line 126: @sujinmkang IMHO, status installed is not relevant for this flow, since it requires some additional steps from user (e.g., power cycle for CPLD/FPGA, etc.), thus such an operation can't be claimed as automatic firmware upgrade.

sonic_platform_base/component_base.py
```diff
@@ -103,3 +103,35 @@ def update_firmware(self, image_path):
             RuntimeError: update failed
         """
         raise NotImplementedError
+
+    def auto_update_firmware(self, image_path, boot_action):
```
nazariig, line 107: @sujinmkang suggest renaming: `boot_action` -> `boot_type`

sonic_platform_base/component_base.py
sujinmkang: @nazariig This has been discussed several time. auto-update is expected to be able to perform install/ update/ creating a task for the components firmware update if the update is available for the boot_type. 
The decision should be done by the platform api. 
For example, with cold boot, cpld firmware update can be done if the reboot plugin can handle the powercycle flag to complete the cpld firmware update. For warm/fast reboot, the cpld firmware update won't happen and the platform api will return boot_type error? fwutil will not exit for the boot_type error since the firmware auto-update is not just supported in that boot_type case.

nazariig: > For warm/fast reboot, the cpld firmware update won't happen and the platform api will return boot_type error? fwutil will not exit for the boot_type error since the firmware auto-update is not just supported in that boot_type case.

@sujinmkang that is clear, but i am talking about the case when we have `status_installed`.
What is the use case for this status code? Basically this means that user still needs to do some manual steps to finish firmware upgrade and this is something completely different to what we call `automatic`.

sonic_platform_base/component_base.py
```diff
@@ -103,3 +103,35 @@ def update_firmware(self, image_path):
...
+            Output: A return code
+                return_code: An integer number, status of component firmware auto-update
+                    - return code of a positive number indicates successful auto-update
+                        - status_installed = 1
+                        - status_updated = 2
+                        - status_scheduled = 3
+                    - return_code of a negative number indicates failed auto-update
+                        - status_err_boot_type = -1
+                        - status_err_image = -2
+                        - status_err_others = -3
```
nazariig, line 132: @sujinmkang maybe it's better to use `err_unknown`?

sujinmkang: @nazariig  please review this PR again.

nazariig (approved):

PR #124, opened 2020-10-01
[sonic-platform-base] Introduce APIs for modular chassis support

mprabhu-nokia:
sonic-platform-base: Changes to enhance module_base.py and chassis_base.py for modular chassis

HLD: https://github.com/Azure/SONiC/pull/646

**- What I did**
Enhance ModuleBase with new APIs to repesent pluggable cards in a voq-chassis.
Enhance ChassisBase with new APIs

**- How I did it**
Add helper APIs in module_base.py

- get_name()
- get_description()
- get_slot()
- get_type()
- get_status()
- reboot()
- set_admin_state()

Add helper APIs in chassis_base.py 

- get_module_index()
- get_supervisor_slot()
- get_my_slot()

**- How I verified it**


sonic_platform_base/card_base.py
```diff
@@ -0,0 +1,80 @@
+#
+# card_base.py
+#
+# Abstract base class for implementing a platform-specific class with which
+# to interact with a card in SONiC chassis
+#
+
+class CardBase(object):
```
jleveque, line 8: The existing `ModuleBase` class was designed as a placeholder base class for supporting removable cards (i.e., "modules") in a modular chassis. Can we not extend or inherit from that class? This class appears to make `ModuleBase` obsolete.

mprabhu-nokia: It seemed like ModuleBase was designed for a non-disaggregated chassis i.e. if line-cards were not having a CPU and their full representation needed to be on the control-card. Intention is not to make ModuleBase obsolete. The CardBase is complementary.

As suggested, we can make it a derived class retaining much of the definition. Does class ModuleCardBase(ModuleBase) work? Other suggestions are welcome.

jleveque: Do the APIs in ModuleBase also apply to what you consider a "card" here? The concept of a "Module" was to represent any pluggable card. If you feel there is a difference, can you please explain further what the differentiation is?

mprabhu-nokia: The APIs of ModuleBase did not seem to apply since for Modular Chassis the ChassisBase covers what ModuleBase does. 
In CardBase we want to represent just the state, type and instance(slot number) of each component like line-card, fabric-card individually. This would be similar to FanBase and PsuBase for example. 

We did deliberate on whether to re-use. If you think we should roll the below APIs in ModuleBase and re-use whatever DeviceBase will provide like status() etc, we can.

jleveque: At the very least, it should inherit from `DeviceBase`. ModuleBase was created to assume that a card could contain an EEPROM, fans, thermals, components (e.g., BIOS, FPGA, CPLD, etc.) and potentially a dedicated power supply. Since you're more familiar with the types of cards in use, is this a correct assumption?

mprabhu-nokia: Card does have eeprom, thermals, components. I think the confusion arises with the disaggregated chassis where a sonic instance runs on the card and all of these get covered by chassis.py. So, seemed redundant unless we want a full line-card (or other cards) representation on Control-Card.

Is it okay if we add the missing APIs from CardBase to ModuleBase and make progress?

jleveque: > Is it okay if we add the missing APIs from CardBase to ModuleBase and make progress?

Yes, please go ahead.

shyam77git: With plan to leverage ModuleBase, can you please update/clarify on the plan to use following existing APIs of chassisBase.py instead of the new ones suggested/listed here?

Use following : get_num_modules() ; get_all _modules() ; get_module() etc.
instead of adding following new helper APIs in chassis_base.py
    get_num_cards(),
    get_card()
    get_all_cards()
    get_controlcard_instance()
    get_my_instance()

sonic_platform_base/card_base.py
```diff
@@ -0,0 +1,80 @@
...
+            string: The status-string of the card
+        """
+        raise NotImplementedError
+
+    def reboot_card(self):
+        """
+        Request to reboot/reset the card
+
+        Returns:
+            bool: True if the request has been successful, False if not
```
lguohan, line 69: please clarify in the description what does it mean by successful? does it mean the line card reboot into a os?

sonic_platform_base/card_base.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        raise NotImplementedError
+
+    def set_admin_state(self, up):
+        """
+        Request to keep the card in administratively up/down state
```
lguohan, line 75: what does administratively down mean? power off? need clarification here.

sonic_platform_base/chassis_base.py
```diff
@@ -463,6 +488,62 @@ def get_status_led(self):
...
+            card = self._card_list[index]
+        except IndexError:
+            sys.stderr.write("Card index {} out of range (0-{})\n".format(
+                             index, len(self._card_list)-1))
+
+        return card
+
+    def get_card_index(self, instance, module_type):
+        """
+        Retrieves card represented by <instance, module-type>
```
lguohan, line 536: what module_type do we have here, can you clarify in the description?

sonic_platform_base/card_base.py
```diff
@@ -0,0 +1,80 @@
...
+            string: module-type MODULE_TYPE_CONTROL, MODULE_TYPE_LINE etc
+        """
+        raise NotImplementedError
+
+    def get_status(self):
+        """
+        Retrieves the status of the card
+
+        Returns:
+            string: The status-string of the card
```
lguohan, line 60: what are the values for the status? can you enumerate?

sonic_platform_base/card_base.py
lguohan: is there difference between reboot and reset? if not, why do we need reboot and reset in the description?

sonic_platform_base/card_base.py
```diff
@@ -0,0 +1,80 @@
...
+    def get_type(self):
+        """
+        Retrieves the type of the card
+
+        Returns:
+            string: module-type MODULE_TYPE_CONTROL, MODULE_TYPE_LINE etc
+        """
+        raise NotImplementedError
+
+    def get_status(self):
```
mprabhu-nokia, line 55: Introduce Empty, Offline, Present, Fault, Online states.

sonic_platform_base/card_base.py
```diff
@@ -0,0 +1,80 @@
...
+#
+
+class CardBase(object):
+    """
+    Abstract base class for implementing a platform specific class to
+    represent a control-card, line-card or fabric-card of a chassis
+    """
+
+    # Possible card types
+    MODULE_TYPE_CONTROL = "CONTROL-CARD"
```
lguohan, line 15: we need to unify the naming convention here. for example, i see supervisor card is also used in the design document. can we use supervisor card universally?

sonic_platform_base/card_base.py
lguohan: what would be the default value?

sonic_platform_base/chassis_base.py
```diff
@@ -118,6 +123,26 @@ def get_reboot_cause(self):
         """
         raise NotImplementedError
 
+    def get_controlcard_instance(self):
```
mprabhu-nokia, line 126: Change instance to slot.

sonic_platform_base/chassis_base.py
```diff
@@ -118,6 +123,26 @@ def get_reboot_cause(self):
...
+        """
+        Retrieves the slot/instance of the control card of the modular chassis
```
minionatwork, line 128: clarify what is "slot/instance" means.. Is it just a "slot" or "instance"?

sonic_platform_base/card_base.py
```diff
@@ -0,0 +1,80 @@
...
+    MODULE_TYPE_LINE    = "LINE-CARD"
+    MODULE_TYPE_FABRIC  = "FABRIC-CARD"
+
+    def get_name(self):
+        """
+        Retrieves the name of the card
```
shyam77git, line 21: the name here is LINE-CARD/ FABRIC-CARD etc.?
or is this the platform/vendor specified name (i.e. PID (productID))?
If its former, then get_description() reflects the platform/vendor supplied PID?

Can you update the comment section of both these APIs?

sonic_platform_base/card_base.py
```diff
@@ -0,0 +1,80 @@
...
+
+        Returns:
+            string: The name of the card
+        """
+        raise NotImplementedError
+
+    def get_description(self):
+        """
+        Retrieves the description of the card
+
```
shyam77git, line 31: Scenario: There could be a case where PID (actual card_type) of remote LC is not available when chassisD is traversing through all cards of the chassis to initialize them.

get_change_event() should facilitate platform to update card description/PID later to chassisD when available.
Can  we ensure that get_change_event() caters to this case?

sonic_platform_base/card_base.py
shyam77git: IMO, default to be Empty.
Platform to determine and update card's 'presence' state as Present.

Follow-on Q:
If platform detects card's presence later or in future, when LC is inserted in an empty slot later, platform would propagate the updated card state via get_change_event() API call?

sonic_platform_base/card_base.py
mprabhu-nokia: Yes, we will re-use what the ModuleBase.py provides. All APIs introduced in ChassisBase.py will remain as is.

sonic_platform_base/card_base.py
mprabhu-nokia: There is no difference between reboot/reset here. It will mean the Module will go offline. The vendor specific API implementation can return successful if the underlying shutdown/powerdown command was issued.

sonic_platform_base/card_base.py
mprabhu-nokia: Yes, it means powered-down. Will be addressed in the comments in the updated code.

sonic_platform_base/card_base.py
mprabhu-nokia: # Possible card status
    MODULE_STATUS_EMPTY   = "Empty"
    MODULE_STATUS_OFFLINE = "Offline"
    MODULE_STATUS_PRESENT = "Present"
    MODULE_STATUS_FAULT   = "Fault"
    MODULE_STATUS_ONLINE  = "Online"

sonic_platform_base/chassis_base.py
mprabhu-nokia: Fixed

sonic_platform_base/chassis_base.py
mprabhu-nokia: This API will be fixed based on review comments. It will be get_module_index(self, module_name)

mprabhu-nokia (commented):
All comments are addressed.
get_change_event() - Will be implemented in an incremental PR. We will make a note of it.
get_name() - Standard name format as updated in the comments
get_description() - PID of the card that can be updated when the information is available.

sonic_platform_base/module_base.py
```diff
@@ -17,6 +17,18 @@ class ModuleBase(device_base.DeviceBase):
     # Device type definition. Note, this is a constant.
     DEVICE_TYPE = "module"
 
+    # Possible card types
+    MODULE_TYPE_CONTROL = "CONTROL-CARD"
+    MODULE_TYPE_LINE    = "LINE-CARD"
+    MODULE_TYPE_FABRIC  = "FABRIC-CARD"
+
+    # Possible card status
+    MODULE_STATUS_EMPTY   = "Empty"
```
mprabhu-nokia, line 26: Add more description for each of the states.

sonic_platform_base/module_base.py
```diff
@@ -17,6 +17,18 @@ class ModuleBase(device_base.DeviceBase):
     # Device type definition. Note, this is a constant.
     DEVICE_TYPE = "module"
 
+    # Possible card types
+    MODULE_TYPE_CONTROL = "CONTROL-CARD"
```
mprabhu-nokia, line 21: Change to Supervisor

sonic_platform_base/chassis_base.py
```diff
@@ -118,6 +118,26 @@ def get_reboot_cause(self):
         """
         raise NotImplementedError
 
+    def get_controlcard_slot(self):
```
lguohan, line 121: rename controlcard to supervisor.

sonic_platform_base/module_base.py
```diff
@@ -77,6 +89,83 @@ def get_system_eeprom_info(self):
...
+    def get_name(self):
+        """
+        Retrieves the name of the module prefixed by CONTROL-CARD, LINE-CARD,
+        FABRIC-CARD
+
+        Returns:
+            string: A string providing the name of the card prefixed by one of the
+            MODULE_TYPE_CONTROL, MODULE_TYPE_LINE, MODULE_TYPE_FABRIC followed by
+            an index.
+            Ex. A Chassis having 1 control-card, 4 line-cards and 6 fabric-cards
```
mprabhu-nokia, line 101: Start index naming from 0

sonic_platform_base/card_base.py
mprabhu-nokia: Follow up -
* Platform (software) driven fault management handling - bottom up approach.

sonic_platform_base/card_base.py
mprabhu-nokia: Yes, get_description() reflects the platform/vendor supplied PID

sonic_platform_base/module_base.py
```diff
@@ -77,6 +89,83 @@ def get_system_eeprom_info(self):
...
+            can provide names CONTROL-CARD1, LINE-CARD1 to LINE-CARD4,
```
lguohan, line 102: CONTROL-CARD1 -> SUPERVISOR1

sonic_platform_base/module_base.py
```diff
@@ -77,6 +89,83 @@ def get_system_eeprom_info(self):
         """
         raise NotImplementedError
 
+    def get_name(self):
```
lguohan, line 92: is this get my name?

sonic_platform_base/module_base.py
```diff
@@ -77,6 +89,83 @@ def get_system_eeprom_info(self):
...
+            string: A string providing the name of the card prefixed by one of the
+            MODULE_TYPE_CONTROL, MODULE_TYPE_LINE, MODULE_TYPE_FABRIC followed by
+            an index.
+            Ex. A Chassis having 1 control-card, 4 line-cards and 6 fabric-cards
+            can provide names CONTROL-CARD1, LINE-CARD1 to LINE-CARD4,
+            FABRIC-CARD1 to FABRIC-CARD6
+        """
+        raise NotImplementedError
+
+    def get_description(self):
```
lguohan, line 107: do we need set_description so that user can change it?

sonic_platform_base/module_base.py
```diff
@@ -77,6 +89,83 @@ def get_system_eeprom_info(self):
...
+            is vendor specific.
+        """
+        raise NotImplementedError
+
+    def get_slot(self):
+        """
+        Retrieves the platform vendor's slot number of the module
+
+        Returns:
+            string: slot representation, usually number
```
lguohan, line 122: what does "usually number" mean here?

sonic_platform_base/module_base.py
```diff
@@ -77,6 +89,83 @@ def get_system_eeprom_info(self):
...
+        """
+        raise NotImplementedError
+
+    def get_type(self):
+        """
+        Retrieves the type of the module.
+
+        Returns:
+            string: A string providing the module-type. Supported values are
+            MODULE_TYPE_CONTROL, MODULE_TYPE_LINE, MODULE_TYPE_FABRIC
```
lguohan, line 132: MODULE_TYPE_SUPERVISOR

sonic_platform_base/module_base.py
```diff
@@ -77,6 +89,83 @@ def get_system_eeprom_info(self):
...
+        """
+        Retrieves the platform vendor's product description of the module
+
+        Returns:
+            string: A string providing the product description of the module. This
+            is vendor specific.
+        """
+        raise NotImplementedError
+
+    def get_slot(self):
```
lguohan, line 117: what is the difference between this get_slot and the get_my_slot above defined in chassis_base.py?

mprabhu-nokia:
> 
> 
> There are different variants/use-cases of reboot such as the whole-card or CPU-only based on fault(s)/error(s)/user-specification etc.
> This should be facilitated in the args list passed to reboot_card() API. Could be enum list or some such

Discuss additional options for reboot options - linux, cpu complex or entire card.

sonic_platform_base/module_base.py
```diff
@@ -77,6 +89,83 @@ def get_system_eeprom_info(self):
...
+        Retrieves the status of the card
+
+        Returns:
+            string: A string providing the status of the module. Support values
+            are MODULE_STATUS_EMPTY, MODULE_STATUS_OFFLINE, MODULE_STATUS_FAULT,
+            MODULE_STATUS_PRESENT, MODULE_STATUS_ONLINE
+        """
+        raise NotImplementedError
+
+    def reboot_card(self):
```
lguohan, line 147: should we simplify the name just reboot(self) since other apis do not have card in the api name.

mprabhu-nokia: yes, will make this change.

sonic_platform_base/module_base.py
```diff
@@ -17,6 +17,18 @@ class ModuleBase(device_base.DeviceBase):
...
+    MODULE_TYPE_CONTROL = "CONTROL-CARD"
+    MODULE_TYPE_LINE    = "LINE-CARD"
+    MODULE_TYPE_FABRIC  = "FABRIC-CARD"
+
+    # Possible card status
+    MODULE_STATUS_EMPTY   = "Empty"
+    MODULE_STATUS_OFFLINE = "Offline"
+    MODULE_STATUS_PRESENT = "Present"
+    MODULE_STATUS_FAULT   = "Fault"
+    MODULE_STATUS_ONLINE  = "Online"
```
lguohan, line 30: can you clarify the meaning of each status in the description.

sonic_platform_base/chassis_base.py
```diff
@@ -118,6 +118,26 @@ def get_reboot_cause(self):
...
+
+    def get_my_slot(self):
+        """
+        Retrieves the slot of this card of the modular chassis
+
+        Returns:
+            An integer, the slot identifier of this card of the
+            modular-chassis.
+        """
+        return 0
```
lguohan, line 139: why the return value is zero?

sonic_platform_base/chassis_base.py
```diff
@@ -118,6 +118,26 @@ def get_reboot_cause(self):
...
 
+    def get_controlcard_slot(self):
+        """
+        Retrieves the slot of the control card of the modular chassis
+
+        Returns:
+            An integer, the slot identifier of the control card of the
+            modular-chassis.
+        """
+        return 0
```
lguohan, line 129: why the return value is 0?

sonic_platform_base/chassis_base.py
mprabhu-nokia: change to 'raise NotImplemented'. Add in documentation.

mprabhu-nokia: return NotAvailable to make it backward compatible.

sonic_platform_base/module_base.py
shyam77git: Since we decided to start card/module indexing with '0', can we change this to SUPERVISOR0.
Also, plan is to insert Supervisor/control-card in slot0 if vendor/platform has support of inserting two Supervisors in the Modular chassis.
Tomorrow, peer-Supervisor would be taken as SUPERVISOR1

sonic_platform_base/card_base.py
mprabhu-nokia: Both get_change_event() and fault_management handling are taken up as enhancement items.

sonic_platform_base/module_base.py
mprabhu-nokia: Done.

sonic_platform_base/module_base.py
mprabhu-nokia: Fixed.

sonic_platform_base/module_base.py
mprabhu-nokia: Did not see a need for this at this point. This comes from the platform vendor.

sonic_platform_base/module_base.py
mprabhu-nokia: Fixed in the description.

sonic_platform_base/module_base.py
mprabhu-nokia: Fixed.

sonic_platform_base/module_base.py
mprabhu-nokia: get_slot() is an API in Module class. On the supervisor, in the module_list[], we can loop through and get a slot of each card using get_slot() API on each module object.

get_my_slot() is to get the slot number of where this SONiC instance is running in dis-aggregated chassis i.e. on supervisor0. it might give 1, and line-card0 it might give slot 2.

sonic_platform_base/module_base.py
mprabhu-nokia: Added.

sonic_platform_base/chassis_base.py
mprabhu-nokia: Added back raise NotImplemented. Will catch exception and handle it when used.

sonic_platform_base/chassis_base.py
mprabhu-nokia: Added back raise NotImplemented. Will catch exception and handle it when used.

sonic_platform_base/module_base.py
mprabhu-nokia: This will return one of the fixed format names i.e. SUPERVISOR0, LINE-CARD1 or FABRIC-CARD3 for example.

sonic_platform_base/chassis_base.py
```diff
@@ -109,6 +109,31 @@ def get_reboot_cause(self):
...
+        Retrieves the slot of this module of the modular chassis.
+        On the supervisor, it will return the slot of the supervisor-module.
+        On the linecard, it will return the slot of the linecard module where
+        this instance of SONiC is running.
+
+        Returns:
+            An integer, the slot identifier of this module of the
+            modular-chassis.
+        """
+        return NotImplementedError
```
jleveque, line 135: Since these methods only apply to modular chassis, the docstrings should state what is expected of a fixed chassis platform.

sonic_platform_base/chassis_base.py
```diff
@@ -197,6 +222,19 @@ def get_module(self, index):
...
+        Retrieves module index from the module name
+
+        Args:
+            name: A string, prefixed by SUPERVISOR, LINE-CARD or FABRIC-CARD
+            Ex. SUPERVISOR0, LINE-CARD1, FABRIC-CARD5
+
+        Returns:
+            An index of the ModuleBase object in the module_list
+        """
+        raise NotImplementedError
```
jleveque, line 236: This method seems like it is platform-independent. Vendors shouldn't need to implement this, as the logic appears as though it will be common. Why is this method added here?

sonic_platform_base/module_base.py
```diff
@@ -68,6 +86,83 @@ def get_system_eeprom_info(self):
         """
         raise NotImplementedError
 
+    def get_name(self):
```
jleveque, line 89: This class inherits from DeviceBase, which already defines a `get_name()` method.

sonic_platform_base/module_base.py
```diff
@@ -68,6 +86,83 @@ def get_system_eeprom_info(self):
...
+        """
+        Retrieves the type of the module.
+
+        Returns:
+            string: A string providing the module-type. Supported values are
+            MODULE_TYPE_SUPERVISOR, MODULE_TYPE_LINE, MODULE_TYPE_FABRIC
+        """
+        raise NotImplementedError
+
+    def get_status(self):
```
jleveque, line 133: This class inherits from DeviceBase, which already defines a `get_status()` method.

jleveque (changes requested):

sonic_platform_base/chassis_base.py
mprabhu-nokia: Fixed.

sonic_platform_base/chassis_base.py
mprabhu-nokia: This is not fixed since we have a common module_list[] across multiple module-types. We start names with SUPERVISOR0, LINE-CARD0 etc. Also, each vendor could have different number of cards of each type based on different modular chassis types.
We need a name to index mapping. This seemed like a common API not specific to module object. So, may be similar to get_all_modules() or get_num_modules() etc.

sonic_platform_base/module_base.py
mprabhu-nokia: Added this mainly to highlight the docstring since we need names in certain standard format. If not, we can remove it. Please suggest.

sonic_platform_base/module_base.py
mprabhu-nokia: Added this mainly to highlight the docstring since it was recommended we need status to be one of the mentioned status-types from MODULE_STATUS_EMPTY, MODULE_STATUS_OFFLINE, MODULE_STATUS_PRESENT, MODULE_STATUS_FAULT and MODULE_STATUS_ONLINE. If not, we can remove it. Please suggest.

shyam77git:
> > There are different variants/use-cases of reboot such as the whole-card or CPU-only based on fault(s)/error(s)/user-specification etc.
> > This should be facilitated in the args list passed to reboot_card() API. Could be enum list or some such
> 
> Discuss additional options for reboot options - linux, cpu complex or entire card.

Can this be facilitated to reboot() api as part of current code/PR?

sonic_platform_base/module_base.py
```diff
@@ -68,6 +89,83 @@ def get_system_eeprom_info(self):
...
+        Request to keep the card in administratively up/down state.
+        The down state will power down the module and the status should show
+        MODULE_STATUS_OFFLINE.
+        The up state will take the module to MODULE_STATUS_PRESENT,
+        MODULE_STATUS_FAULT or MODULE_STAUS_ONLINE states.
+
+        Returns:
+            bool: True if the request has been issued successfully, False if not
+        """
+        raise NotImplementedError
```
jleveque, line 167: Can you add an "Args" section of this docstring to describe the type of value expected to be passed in as `up`?

sonic_platform_base/chassis_base.py
```diff
@@ -197,6 +232,19 @@ def get_module(self, index):
 
         return module
 
+    def get_module_index(self, module_name):
+        """
+        Retrieves module index from the module name
+
+        Args:
+            name: A string, prefixed by SUPERVISOR, LINE-CARD or FABRIC-CARD
```
jleveque, line 240: Arg is called `name` here, but in the method signature it is `module_name`. Please align them.

jleveque (changes requested):

sonic_platform_base/module_base.py
```diff
@@ -17,6 +17,35 @@ class ModuleBase(device_base.DeviceBase):
     # Device type definition. Note, this is a constant.
     DEVICE_TYPE = "module"
 
+    # Possible card types for modular chassis
+    MODULE_TYPE_SUPERVISOR = "SUPERVISOR"
+    MODULE_TYPE_LINE    = "LINE-CARD"
+    MODULE_TYPE_FABRIC  = "FABRIC-CARD"
+
+    # Possible card status for modular chassis
+    #Module state is Empty if no module is inserted in the slot
```
jleveque, line 26: Please add a space after `#`, and do for all other comments like this below.

jleveque (changes requested):

sonic_platform_base/module_base.py
mprabhu-nokia: Done.

sonic_platform_base/chassis_base.py
mprabhu-nokia: Fixed.

sonic_platform_base/module_base.py
mprabhu-nokia: Fixed.

jleveque:
@mprabhu-nokia: Can you please update the PR title and description since they no longer reflect the changes here?

mprabhu-nokia:
> 
> 
> @mprabhu-nokia: Can you please update the PR title and description since they no longer reflect the changes here?

Agree. Fixed both to reflect latest changes.

jleveque:
@Staphylo: Can you please review this, also?

Staphylo:
LGTM

sonic_platform_base/chassis_base.py
```diff
@@ -109,6 +109,41 @@ def get_reboot_cause(self):
...
+        supervisor-module.
+
+        On the fixed-platforms, the API can be ignored.
+
+        Users of the API can catch the exception and return a default
+        ModuleBase.MODULE_INVALID_SLOT and bypass code for fixed-platforms.
+
+        Returns:
+            An integer, the slot identifier of the supervisor module of the
+            modular-chassis.
```
jleveque, line 125: Can you also specify whether the slot identifier is expected to be 0-based or 1-based? Please see the docstring for `get_fan()` for an example.

sonic_platform_base/chassis_base.py
```diff
@@ -109,6 +109,41 @@ def get_reboot_cause(self):
...
+        this instance of SONiC is running.
+
+        On the fixed-platforms, the API can be ignored.
+
+        Users of the API can catch the exception and return a default
+        ModuleBase.MODULE_INVALID_SLOT and bypass code for fixed-platforms.
+
+        Returns:
+            An integer, the slot identifier of this module of the
+            modular-chassis.
```
jleveque, line 143: Can you also specify whether the slot identifier is expected to be 0-based or 1-based? Please see the docstring for `get_fan()` for an example.

sonic_platform_base/chassis_base.py
```diff
@@ -197,6 +232,19 @@ def get_module(self, index):
...
+    def get_module_index(self, module_name):
+        """
+        Retrieves module index from the module name
+
+        Args:
+            module_name: A string, prefixed by SUPERVISOR, LINE-CARD or FABRIC-CARD
+            Ex. SUPERVISOR0, LINE-CARD1, FABRIC-CARD5
+
+        Returns:
+            An index of the ModuleBase object in the module_list
```
jleveque, line 244: Suggest rewording to specify return value type:
```
An integer, the index of the ModuleBase object in the module_list
```

sonic_platform_base/module_base.py
```diff
@@ -68,6 +97,92 @@ def get_system_eeprom_info(self):
...
 
+    def get_name(self):
+        """
+        Retrieves the name of the module prefixed by SUPERVISOR, LINE-CARD,
+        FABRIC-CARD
+
+        Returns:
+            string: A string providing the name of the card prefixed by one of the
+            MODULE_TYPE_SUPERVISOR, MODULE_TYPE_LINE, MODULE_TYPE_FABRIC followed by
+            an index.
```
jleveque, line 108: Suggest rewording:

```
            string: A string providing the name of the module prefixed by one of the predefined constants:
            MODULE_TYPE_SUPERVISOR, MODULE_TYPE_LINE, or MODULE_TYPE_FABRIC, followed by
            an integer representing the index of the module.
```

sonic_platform_base/module_base.py
```diff
@@ -68,6 +97,92 @@ def get_system_eeprom_info(self):
...
+        """
+        raise NotImplementedError
+
+    def get_type(self):
+        """
+        Retrieves the type of the module.
+
+        Returns:
+            string: A string providing the module-type. Supported values are
+            MODULE_TYPE_SUPERVISOR, MODULE_TYPE_LINE, MODULE_TYPE_FABRIC
```
jleveque, line 140: Suggest rewording:

```
            string: A string specifying the module type. Must be one of the following predefined constants:
            MODULE_TYPE_SUPERVISOR, MODULE_TYPE_LINE, MODULE_TYPE_FABRIC
```

Staphylo (approved):
LGTM

sonic_platform_base/module_base.py
```diff
@@ -68,6 +97,92 @@ def get_system_eeprom_info(self):
...
+        raise NotImplementedError
+
+    def get_status(self):
+        """
+        Retrieves the status of the card
+
+        Returns:
+            string: A string providing the status of the module. Support values
+            are MODULE_STATUS_EMPTY, MODULE_STATUS_OFFLINE, MODULE_STATUS_FAULT,
+            MODULE_STATUS_PRESENT, MODULE_STATUS_ONLINE
```
jleveque, line 151: Suggest rewording:

```
            string: A string providing the status of the module.
            Must be one of the following predefined constants:
            MODULE_STATUS_EMPTY, MODULE_STATUS_OFFLINE, MODULE_STATUS_FAULT,
            MODULE_STATUS_PRESENT, MODULE_STATUS_ONLINE
```

sonic_platform_base/module_base.py
```diff
@@ -68,6 +97,92 @@ def get_system_eeprom_info(self):
...
+        raise NotImplementedError
+
+    def reboot(self, reboot_type):
+        """
+        Request to reboot the module
+
+        Args:
+            reboot_type: A string, choose from one of pre-defined module
+            reboot-types. MODULE_REBOOT_DEFAULT, MODULE_REBOOT_CPU_COMPLEX,
+            MODULE_REBOOT_FPGA_COMPLEX
```
jleveque, line 162: Suggest rewording:

```
            reboot_type: A string, must be one of the predefined module
            reboot-type constants: MODULE_REBOOT_DEFAULT, MODULE_REBOOT_CPU_COMPLEX,
            MODULE_REBOOT_FPGA_COMPLEX
```

jleveque (changes requested):

sonic_platform_base/chassis_base.py
mprabhu-nokia: This would be slightly different. This returns the physical-slot of the module in the chassis. It is vendor specific. Made changes to the comment.

sonic_platform_base/chassis_base.py
mprabhu-nokia: This would be slightly different. This returns the physical-slot of the module in the chassis. It is vendor specific. Made changes to the comment.

sonic_platform_base/module_base.py
mprabhu-nokia: Fixed. Please review.

sonic_platform_base/module_base.py
mprabhu-nokia: Fixed. Please review.
Am renaming this API to get_oper_status() since the device_base.get_status() returns a bool. So, don't want users to get confused.

sonic_platform_base/module_base.py
mprabhu-nokia: Fixed. Please review.

sonic_platform_base/module_base.py
mprabhu-nokia: Fixed. Please review.

sonic_platform_base/chassis_base.py
mprabhu-nokia: Fixed. Please review.

sonic_platform_base/module_base.py
jleveque: @mprabhu-nokia: I see you have renamed this to `get_oper_status()`, making it a new, module-specific method. That's good. However, I still wonder if it is necessary. There are already `get_presence()` and `get_status()` methods inherited from DeviceBase, both which return boolean values. `MODULE_STATUS_EMPTY` and `MODULE_STATUS_PRESENT` are redundant with `get_presence()`, and `get_status()` can return True which represents `MODULE_STATUS_ONLINE` or False which would represent either `MODULE_STATUS_OFFLINE` or `MODULE_STATUS_FAULT`. The only thing that might not be easy to differentiate between is `MODULE_STATUS_OFFLINE` or `MODULE_STATUS_FAULT`.

What are your thoughts on this?

mprabhu-nokia: This was discussed in the review meeting. I had initially proposed just presence and status. However, multiple reviewers wanted more stateful information to be provided and displayed in the DB (and show command). 
Reviewers wanted more states(status strings) and reason-code to be added as well. Those can be taken up as enhancement PRs.

jleveque:
@mprabhu-nokia: Once this PR merges, can you please also add regression tests for these new API methods in the sonic-mgmt repo: 
https://github.com/Azure/sonic-mgmt/blob/master/tests/platform_tests/api/test_module.py?

jleveque (approved):

sonic_platform_base/module_base.py
```diff
@@ -161,6 +174,10 @@ def set_admin_state(self, up):
         The up state will take the module to MODULE_STATUS_PRESENT,
         MODULE_STATUS_FAULT or MODULE_STAUS_ONLINE states.
```
shyam77git, line 175: up state request on MODULE_STATUS_EMPTY would have no effect (i.e. NO-OP).
up state can't take the module to  MODULE_STATUS_PRESENT. 
Module has to be PRESENT for 'up' state to take it to MODULE_STAUS_ONLINE or MODULE_STATUS_FAULT.

In that case, suggest removing MODULE_STATUS_PRESENT from the comment

sonic_platform_base/module_base.py
```diff
@@ -35,6 +35,9 @@ class ModuleBase(device_base.DeviceBase):
     #Module state is Online when fully operational
```
shyam77git: Consider a use-case: 
Temperature sensor on a LC or FC crosses its critical threshold. Platform expected to power-down/shutdown the card. thermCtlD (via policy file) asks platform to do the needful. 

Per the above card state definitions, card can't be moved to MODULE_STATUS_OFFLINE as its not user/admin-config initiated. Card to be moved to MODULE_STATUS_FAULT. However, card is actually powered-off by the platform (to protect the board/card) whereas MODULE_STATUS_FAULT definition says card is still powered-on though in a fault state. 

Propose to resolve this ambiguity by introducing MODULE_STATUS_POWERED_OFF (or MODULE_STATUS_POWERED_DOWN).

mprabhu-nokia:
> 
> 
> @mprabhu-nokia: Once this PR merges, can you please also add regression tests for these new API methods in the sonic-mgmt repo:
> https://github.com/Azure/sonic-mgmt/blob/master/tests/platform_tests/api/test_module.py?

Yes, will do.

sonic_platform_base/module_base.py
mprabhu-nokia: Addressed. Thanks.

sonic_platform_base/module_base.py
mprabhu-nokia: Added MODULE_STATUS_POWERED_DOWN.

shyam77git (approved):

jleveque:
@lguohan: Since you had outstanding comments, can you please review again before merge?

PR #125, opened 2020-10-02
[SONIC_SFP] adding support for writing the eeprom address space inside sfputilbase. 

vdahiya12:
Summary:
This PR provides support for writing a buffer into eeprom address space. Using this the platform api once it imports the sfp_util it would be able to call this write eeprom functionality as a way to implement the write_api. 
### Type of change

- [ ] Bug fix
- [] Test case(new/improvement)

### Approach
Added changes in the sonic_sfp directory of sonic-platform-common

To be able to provide the functionality to write to the eeprom address space helper utility for the platform api to import

#### How did you do it?
Added the API's with the proper implementation and convention support
for writing inside the eeprom

Tested these api's on Arista 7050CX-3 
Instantiate an SFP object from arista.utils.sonic_platform.sfp.
verify that the Credo interfaces are up , then call the API with param

#### Any platform specific information?

Platform must have optoe driver support for  instantiation of sysfs/eeprom 


jleveque (changes requested):
Suggest adding docstrings to the mux-related functions which describe the return values and also detail how the offset is calculated (how it maps to the byte of a specific page).

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -11,8 +11,10 @@
     import os
     import re
     import sys
+    import types
     from collections import OrderedDict
 
+    from ctypes import create_string_buffer
```
jleveque, line 17: Looks like these imports are no longer necessary?

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -369,6 +371,40 @@ def _read_eeprom_devid(self, port_num, devid, offset, num_bytes = 256):
 
         return eeprom_raw
 
+    def _write_eeprom_specific_bytes(self, sysfsfile_eeprom, offset, num_bytes, write_buffer):
+
```
jleveque, line 375: Suggest deleting this blank line to match the style of the rest of the file

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -369,6 +371,40 @@ def _read_eeprom_devid(self, port_num, devid, offset, num_bytes = 256):
...
+            sysfsfile_eeprom.write(write_buffer)
+            sysfsfile_eeprom.close()
+        except IOError:
+            print("Error: writing EEPROM sysfs file")
+            return False
+
+        return True
+
+    def _write_eeprom_devid(self, port_num, devid, offset, num_bytes, write_buffer):
+
```
jleveque, line 387: Suggest deleting this blank line to match the style of the rest of the file

jleveque (changes requested):

sonic_platform_base/sonic_sfp/sfputilbase.py
vdahiya12: fixed

sonic_platform_base/sonic_sfp/sfputilbase.py
vdahiya12: fixed

sonic_platform_base/sonic_sfp/sfputilbase.py
vdahiya12: fixed

jleveque (approved):

jleveque (approved):

sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -369,6 +369,38 @@ def _read_eeprom_devid(self, port_num, devid, offset, num_bytes = 256):
 
         return eeprom_raw
 
+    def _write_eeprom_specific_bytes(self, sysfsfile_eeprom, offset, num_bytes, write_buffer):
+        try:
+            sysfsfile_eeprom.seek(offset)
+            sysfsfile_eeprom.write(write_buffer)
+            sysfsfile_eeprom.close()
```
Staphylo, line 376: I would suggest not closing a file descriptor provided by a parent function.
The `_write_eeprom_devid` which is currently the only user is already taking care of this.

If what you want to flush the buffer you can call `sysfsfile_eeprom.flush()`.
However adding `buffering=0` to `open` would prevent any buffering from happening in the first place, which I recommend in this case.

Here is an example that would be problematic by keeping the close here.
```
with open(path, mode='wb', buffering=0) as eeprom:
    self._write_eeprom_specific_bytes(eeprom, 0, 3, b'abc')
    self._write_eeprom_specific_bytes(eeprom, 3, 3, b'def') # exception because eeprom is closed
```

vdahiya12: fixed

jleveque (approved):

PR #126, opened 2020-10-06
[SONIC_SFP] adding abstract methods for reading and writing the eeprom address space within platform api

vdahiya12:
Summary:
This PR defines the base class methods for reading and writing eeprom address space, which is going to be utilized by 
xcvrd and cli for as a part of mux driver implementation for the Credo Y cable. With this change vendors will have to implement these api's as a part of SFP's since they inherit sffbase class which in turn will provide the upper layers to be able to read/write the sysfs eeprom address space

### Type of change

- [ ] Bug fix
- [ ] Test case(new/improvement)

### Approach
Added changes in the sonic_sfp directory of sonic-platform-common

To configure the Credo Y cable through the Platform API's which is the fundamental for dual TOR support for SONIC

#### How did you do it?
Added the API's with the abstract definitions 
for the basic configurations/capabilities of Credo Y cable within sfputilbase.py

Nothing to verify

#### Any platform specific information?

Platform must have optoe driver support for  instantiation of sysfs/eeprom 

### DIFF
--- a/sonic_platform_base/sfp_base.py
+++ b/sonic_platform_base/sfp_base.py
@@ -307,8 +307,7 @@ class SfpBase(device_base.DeviceBase):
 
     def read_eeprom(self, offset, num_bytes):
         """
-        read eeprom specfic bytes for a given port_num and device id
-        beginning from a random offset with size as num_bytes
+        read eeprom specfic bytes beginning from a random offset with size as num_bytes
 
         Args:
              offset :
@@ -323,8 +322,8 @@ class SfpBase(device_base.DeviceBase):
 
     def write_eeprom(self, offset, num_bytes, write_buffer):
         """
-        write eeprom specfic bytes for a given port_num and device id
-        beginning from a random offset with size as num_bytes and write_buffer as the required bytes
+        write eeprom specfic bytes beginning from a random offset with size as num_bytes 
+        and write_buffer as the required bytes
 
         Args:
              offset :
--- a/sonic_platform_base/sfp_base.py
+++ b/sonic_platform_base/sfp_base.py
@@ -308,7 +308,7 @@ class SfpBase(device_base.DeviceBase):
     def read_eeprom(self, offset, num_bytes):
         """
         read eeprom specfic bytes for a given port_num and device id
-        begining from a random offset with size as num_bytes
+        beginning from a random offset with size as num_bytes
 
         Args:
              offset :
@@ -317,22 +317,22 @@ class SfpBase(device_base.DeviceBase):
                      Integer, the number of bytes to be read
 
         Returns:
-            a bytearray , raw sequence of bytes which are read from the offset of size num_bytes
+            bytes, raw sequence of bytes which are read from the offset of size num_bytes
         """
         raise NotImplementedError
 
     def write_eeprom(self, offset, num_bytes, write_buffer):
         """
         write eeprom specfic bytes for a given port_num and device id
-        begining from a random offset with size as num_bytes and write_buffer as the required bytes
+        beginning from a random offset with size as num_bytes and write_buffer as the required bytes
 
         Args:
              offset :
                      Integer, the offset from which the read transaction will start
              num_bytes:
-                     Integer, the number of bytes to be read
+                     Integer, the number of bytes to be written
              write_buffer:
-                     bytearary, raw bytes buffer which is to be written beginning at the offset
+                     bytes, raw bytes buffer which is to be written beginning at the offset
 
         Returns:
             a Boolean, true if the write succeeded and false if it did not succeed.
--- a/sonic_platform_base/sfp_base.py
+++ b/sonic_platform_base/sfp_base.py
@@ -316,7 +316,8 @@ class SfpBase(device_base.DeviceBase):
                      Integer, the number of bytes to be read
 
         Returns:
-            bytearray, raw sequence of bytes which are read from the offset of size num_bytes
+            bytearray, if raw sequence of bytes are read correctly from the offset of size num_bytes
+            None , if the read_eeprom fails
         """
         raise NotImplementedError
 
--- a/sonic_platform_base/sfp_base.py
+++ b/sonic_platform_base/sfp_base.py
@@ -317,7 +317,7 @@ class SfpBase(device_base.DeviceBase):
 
         Returns:
             bytearray, if raw sequence of bytes are read correctly from the offset of size num_bytes
-            None , if the read_eeprom fails
+            None, if the read_eeprom fails
         """
         raise NotImplementedError
 
--- a/sonic_platform_base/sfp_base.py
+++ b/sonic_platform_base/sfp_base.py
@@ -305,45 +305,37 @@ class SfpBase(device_base.DeviceBase):
         """
         raise NotImplementedError
 
-    def read_eeprom_devid(self, port_num, devid, offset, num_bytes):
+    def read_eeprom(self, offset, num_bytes):
         """
         read eeprom specfic bytes for a given port_num and device id
         begining from a random offset with size as num_bytes
 
         Args:
-             port_num :
-                     Integer, index of physical port from where to read the bytes
-             devid :
-                     Integer, device id of the i2c bus required for reading for eeprom addr space
              offset :
                      Integer, the offset from which the read transaction will start
              num_bytes:
                      Integer, the number of bytes to be read
 
         Returns:
-            a list , raw sequence of bytes which are read from the offset of size num_bytes
+            a bytearray , raw sequence of bytes which are read from the offset of size num_bytes
         """
         raise NotImplementedError
 
-    def write_eeprom_devid(self, port_num, devid, offset, num_bytes, write_buffer):
+    def write_eeprom(self, offset, num_bytes, write_buffer):
         """
         write eeprom specfic bytes for a given port_num and device id
         begining from a random offset with size as num_bytes and write_buffer as the required bytes
 
         Args:
-             port_num :
-                     Integer, index of physical port from where to read the bytes
-             devid :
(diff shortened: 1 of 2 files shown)


sonic_platform_base/sonic_sfp/sffbase.py
```diff
@@ -247,3 +250,27 @@ def dump_pretty(self, indict):
                     self.dec_indent()
             else:
                 print(self._indent, elem, ': ', elem_val)
+
+    @abc.abstractmethod
+    def read_eeprom_devid(self, port_num, devid, offset, num_bytes):
```
vdahiya12, line 255: @jleveque I think we should omit the port_num and let it come from sfp index ? 
Or 
Do we want the flexiblity of writing to other ports as well from a different port ?

jleveque: This is not the file for the abstract methods which the vendor implements. It is https://github.com/Azure/sonic-platform-common/blob/master/sonic_platform_base/sfp_base.py

vdahiya12: fixed

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,47 @@ def set_power_override(self, power_override, power_set):
             False if not
         """
         raise NotImplementedError
+
+    def read_eeprom_devid(self, port_num, devid, offset, num_bytes):
```
jleveque, line 308: There should be no need for port_num or devid here. If the platform uses sysfs/i2c to communicate with the EEPROM, the vendor will derive those based on which transceiver this function is called on and can in turn call the functions in sfputilbase.py.

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,47 @@ def set_power_override(self, power_override, power_set):
...
+                     Integer, the offset from which the read transaction will start
+             num_bytes:
+                     Integer, the number of bytes to be read
+
+        Returns:
+            a list , raw sequence of bytes which are read from the offset of size num_bytes
+        """
+        raise NotImplementedError
+
+    def write_eeprom_devid(self, port_num, devid, offset, num_bytes, write_buffer):
```
jleveque, line 328: There should be no need for port_num or devid here. If the platform uses sysfs/i2c to communicate with the EEPROM, the vendor will derive those based on which transceiver this function is called on and can in turn call the functions in sfputilbase.py.

jleveque (changes requested):

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,47 @@ def set_power_override(self, power_override, power_set):
...
+                     Integer, index of physical port from where to read the bytes
+             devid :
+                     Integer, device id of the i2c bus required for reading for eeprom addr space
+             offset :
+                     Integer, the offset from which the read transaction will start
+             num_bytes:
+                     Integer, the number of bytes to be read
+
+        Returns:
+            a list , raw sequence of bytes which are read from the offset of size num_bytes
```
jleveque, line 324: Should this return a list or a byte array? We also must keep in mind we are transitioning to Python 3, where strings are handled differently. We should plan ahead.

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,47 @@ def set_power_override(self, power_override, power_set):
...
+                     Integer, device id of the i2c bus required for reading for eeprom addr space
+             offset :
+                     Integer, the offset from which the read transaction will start
+             num_bytes:
+                     Integer, the number of bytes to be read
+             write_buffer:
+                     list, raw bytes buffer which is to be written beginning at the offset
+
+        Returns:
+            a Boolean , true if the write succeeded and false if it did not succeed.
```
jleveque, line 346: Remove space before comma

jleveque (changes requested):

sonic_platform_base/sfp_base.py
vdahiya12: i agree port_num should come from SFP class instantiation, 
I think for devid it could only be 0x50 eeprom or 0x51 dom, 
But we can have fix this has 0x50 since we only have one i2c device to be read. 
So i will omit both these .

sonic_platform_base/sfp_base.py
vdahiya12: I will change this to a byte array .

sonic_platform_base/sfp_base.py
vdahiya12: fixed

sonic_platform_base/sfp_base.py
vdahiya12: fixed

sonic_platform_base/sfp_base.py
vdahiya12: fixed

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,39 @@ def set_power_override(self, power_override, power_set):
...
+        begining from a random offset with size as num_bytes
+
+        Args:
+             offset :
+                     Integer, the offset from which the read transaction will start
+             num_bytes:
+                     Integer, the number of bytes to be read
+
+        Returns:
+            a bytearray , raw sequence of bytes which are read from the offset of size num_bytes
```
jleveque, line 320: Remove space before comma. Also, should we make this a `bytearray`, or `bytes` object?

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,39 @@ def set_power_override(self, power_override, power_set):
...
+    def write_eeprom(self, offset, num_bytes, write_buffer):
+        """
+        write eeprom specfic bytes for a given port_num and device id
+        begining from a random offset with size as num_bytes and write_buffer as the required bytes
+
+        Args:
+             offset :
+                     Integer, the offset from which the read transaction will start
+             num_bytes:
+                     Integer, the number of bytes to be read
```
jleveque, line 333: Docstrings mention "read" instead of "write". Copy/paste mistake?

jleveque (changes requested):

sonic_platform_base/sfp_base.py
vdahiya12: fixed

sonic_platform_base/sfp_base.py
vdahiya12: in python > 3 I think bytes and bytearray are fairly interchangeable so i have changed this to bytes.

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,39 @@ def set_power_override(self, power_override, power_set):
...
+
+        Returns:
+            bytes, raw sequence of bytes which are read from the offset of size num_bytes
+        """
+        raise NotImplementedError
+
+    def write_eeprom(self, offset, num_bytes, write_buffer):
+        """
+        write eeprom specfic bytes for a given port_num and device id
```
jleveque, line 326: port_num and device id are no longer given here

jleveque (changes requested):

sonic_platform_base/sfp_base.py
vdahiya12: fixed

sonic_platform_base/sfp_base.py
jleveque: In Python 3 bytes and bytearray are fairly interchangeable, but seeing as we have not fully transitioned to Python 3 and need to retain support for Python 2 until all vendors have created Python 3 versions of their packages, we might want to err on the side of backward-compatibility. In that case, `bytearray` might be the way to go because it is basically the same between Python 2 and 3. Thoughts?

vdahiya12: i think if we keep just bytes, that would be like using str in python2.x, which might break things if we transition to python3.x as then bytes is not string in python > 3. 
so "bytearray" should be the way to go. I agree.

jleveque (approved):

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,38 @@ def set_power_override(self, power_override, power_set):
             False if not
         """
         raise NotImplementedError
+
+    def read_eeprom(self, offset, num_bytes):
```
lguohan, line 308: maybe we can add a bulk api, where you can have a list of [offset, number_bytes], so that it can read a few fields at once. same for the write_eeprom api.

also, do we need read/write a bit? is that supported by i2c bus?

jleveque: >maybe we can add a bulk api, where you can have a list of [offset, number_bytes], so that it can read a few fields at once. same for the write_eeprom api.

We would also need to add these bulk functions to the underlying layer. Do you foresee a performance issue if we do not support bulk transfers?

> also, do we need read/write a bit? is that supported by i2c bus?

No, the unit of transfer for I2C is one byte.

sonic_platform_base/sfp_base.py
```diff
@@ -304,3 +304,39 @@ def set_power_override(self, power_override, power_set):
...
+
+        Args:
+             offset :
+                     Integer, the offset from which the read transaction will start
+             num_bytes:
+                     Integer, the number of bytes to be read
+
+        Returns:
+            bytearray, if raw sequence of bytes are read correctly from the offset of size num_bytes
+            None , if the read_eeprom fails
```
jleveque, line 320: Delete space before comma

jleveque (changes requested):

sonic_platform_base/sfp_base.py
vdahiya12: fixed

PR #136, opened 2020-10-20
Common power consumption and supply APIs for modular chassis

mprabhu-nokia:
sonic-platform-base: Changes to introduce APIs for modular chassis for power-consumption and supplied

HLD: Azure/SONiC#646

PSUd APIs for power requirement calculations

get_maximum_supplied_power() - per PSU
get_status_master_led() - get master psu led status. Class method.
set_status_master_led() - set master psu led status. Class method.

get_maximum_consumed_power(self) - per consumer API. Consumers are modules, Fans

### DIFF
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -26,12 +26,6 @@ class ChassisBase(device_base.DeviceBase):
     REBOOT_CAUSE_HARDWARE_OTHER = "Hardware - Other"
     REBOOT_CAUSE_NON_HARDWARE = "Non-Hardware"
 
-    # Possible fan status LED colors
-    STATUS_LED_COLOR_GREEN = "green"
-    STATUS_LED_COLOR_AMBER = "amber"
-    STATUS_LED_COLOR_RED = "red"
-    STATUS_LED_COLOR_OFF = "off"
-
     # List of ComponentBase-derived objects representing all components
     # available on the chassis
     _component_list = None
@@ -150,6 +144,17 @@ class ChassisBase(device_base.DeviceBase):
         """
         return NotImplementedError
 
+    def is_modular_chassis(self):
+        """
+        Retrieves whether the sonic instance is part of modular chassis
+
+        Returns:
+            A bool value, should return False by default or for fixed-platforms.
+            Should return True for supervisor-cards, line-cards etc running as part
+            of modular-chassis.
+        """
+        return False
+
     ##############################################
     # Component methods
     ##############################################
@@ -192,45 +197,6 @@ class ChassisBase(device_base.DeviceBase):
                              index, len(self._component_list)-1))
 
         return component
-
-    def get_all_power_consumers(self):
-        """
-        Retrieves a list of all power consumers in a modular chassis
-
-        Returns:
-            A list, where each entry is a unique power consumer. For example,
-            'FAN', 'CARD' etc
-        """
-        return None
-
-    def get_total_consumer_power(self, consumer):
-        """
-        Retrives the maximum total power drawn by the consumer component
-
-        Returns:
-            A float, with value of the maximum consumable power of the
-            component.
-        """
-        return 0.0
-
-    def get_status_master_led(self, device_type):
-        """
-        Gets the state of the Master status LED for a given device-type
-
-        Returns:
-            A string, one of the predefined STATUS_LED_COLOR_* strings.
-        """
-        return STATUS_LED_COLOR_OFF
-
-    def set_status_master_led(self, device_type, color):
-        """
-        Gets the state of the Master status LED for a given device-type
-
-        Returns:
-            bool: True if status LED state is set successfully, False if
-                  not
-        """
-        return False
 
     ##############################################
     # Module methods
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -26,6 +26,12 @@ class ChassisBase(device_base.DeviceBase):
     REBOOT_CAUSE_HARDWARE_OTHER = "Hardware - Other"
     REBOOT_CAUSE_NON_HARDWARE = "Non-Hardware"
 
+    # Possible fan status LED colors
+    STATUS_LED_COLOR_GREEN = "green"
+    STATUS_LED_COLOR_AMBER = "amber"
+    STATUS_LED_COLOR_RED = "red"
+    STATUS_LED_COLOR_OFF = "off"
+
     # List of ComponentBase-derived objects representing all components
     # available on the chassis
     _component_list = None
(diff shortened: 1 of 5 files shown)


sonic_platform_base/chassis_base.py
```diff
@@ -152,6 +158,45 @@ def get_component(self, index):
...
+        """
+        Retrives the maximum total power drawn by the consumer component
+
+        Returns:
+            A float, with value of the maximum consumable power of the
+            component.
+        """
+        return 0.0
+
+    def get_status_master_led(self, device_type):
```
Staphylo, line 181: There seems to be an overlap with https://github.com/Azure/sonic-platform-common/commit/9f3be4446f708d5a171b1be530a980ef2e822838

minionatwork: This one is for master led. 
per PSU led is set by 9f3be44

sonic_platform_base/chassis_base.py
```diff
@@ -152,6 +158,45 @@ def get_component(self, index):
...
+        """
+        Retrieves a list of all power consumers in a modular chassis
+
+        Returns:
+            A list, where each entry is a unique power consumer. For example,
+            'FAN', 'CARD' etc
+        """
+        return None
+
+    def get_total_consumer_power(self, consumer):
```
mprabhu-nokia, line 171: Move this API to FanBase and ModuleBase

sonic_platform_base/chassis_base.py
Staphylo: As per the discussion I would be advocating for the following to manage status led
```
Chassis.get_status_led(self, name)
```

This way we could agree on a common name for the `master_led` and psud could change if if necessary.

```
led = chassis.get_status_led(chassis.LED_MASTER)
if led:
    led.set_color(chassis.COLOR_AMBER)
```

All other status led would also benefit from the same feature set and adapt to various platform designs.
Given that platforms might or might not have the same leds, platform daemon could conditionally set the leds when they exist.

@jleveque do you have some thoughts?

jleveque: @Staphylo: I think your idea is reasonable.

@Junchao-Mellanox: How does this sound to you?

Junchao-Mellanox: From my understanding, if user wants to get/set PSU/FAN led, you could do it like:

```
led_color = chassis.get_psu(psu_index).get_status_led()
chassis.get_psu(psu_index).set_status_led("red")
led_color = chassis.get_fan(fan_index).get_status_led()
chassis.get_psu(fan_index).set_status_led("green")
```

So I would like to understand, on which circumstances that we need the new API get_status_master_led/set_status_master_led.

BTW the method "get_status_led" returns the led color instead of a led object, we don't have led object defined in platform API for now.

mprabhu-nokia: These are applicable for the modular chassis. For fixed platform I can draw a parallel to system led. The master led will indicate comprehensive state of each type of peripheral.

For example, master PSU led will indicate where the chassis is fully powered, is low or under powered. Similarly, there is a Master fabric led or fan led state. The individual leds on a chassis sometime could not be visible on the front of the chassis. The Master Led on the supervisor will provide these indicators.

Junchao-Mellanox: OK, understood. Then it looks good to me. As a PSU master led is supposed to indicate the PSU state, another option is to add the master led as a class level method in psu_base.py, fan_base.py and fan_drawer.py. Something like:

```python
@classmethod
def get_status_master_led(cls):
    ...
```

mprabhu-nokia: @Junchao-Mellanox - The classmethod approach is good. However, the implementation will still be in Psu(PsuBase) class. I will not be able to implement in the base class. If I implement in Psu class, then I have to try and import Psu from sonic_platform.psu. 
Is my understanding correct?

Junchao-Mellanox: Yes, you are correct.

sonic_platform_base/chassis_base.py
mprabhu-nokia: Done.

sonic_platform_base/chassis_base.py
mprabhu-nokia: Fixed. Please also see https://github.com/Azure/sonic-platform-daemons/pull/104 for usage.

shyam77git:
As discussed in 10/21 sync-up mtg., LEDs are just no specific to PSUs only.
Other boards on the chassis - Supervisor, LC, FC and Fan Trays too have LEDs + chassis itself has LEDs.
Recommend having a LED daemon/service as part of PMON HLD to manage (get/set/monitor) them all and have some 'show led' CLI to display them. Note: this to be different from current ledd daemon/service which manages optics/interfaces LEDs

shyam77git:
Recommend get_status_master_led() and set_status_master_led() as optional methods as not all platform has/support this.

mprabhu-nokia:
> 
> 
> Recommend get_status_master_led() and set_status_master_led() as optional methods as not all platform has/support this.

Yes, these will be made optional.

> 
> 
> Recommend get_status_master_led() and set_status_master_led() as optional methods as not all platform has/support this.

Yes, these are optional.

jleveque:
Please fix conflicts

mprabhu-nokia:
> 
> 
> Please fix conflicts

Resolved conflicts.

sonic_platform_base/device_base.py
```diff
@@ -11,6 +11,17 @@ class DeviceBase(object):
     peripheral device
     """
 
+    #Device-types
+    DEVICE_TYPE_PSU     = "PSU"
+    DEVICE_TYPE_FAN     = "FAN"
+    DEVICE_TYPE_FANDRAWER = "FAN-DRAWER"
```
jleveque, line 17: Why do we need these in DeviceBase? DeviceBase is meant to be an abstract base class with no concept of the types of classes which inherit from it.

mprabhu-nokia: Removed it since the implementation in PSUd has changed. Previously, we were getting a list of power consumers and wanted a common place for standard device-types. Also, modules are now limited to supervisor, line-cards and fabric-cards.

jleveque:
@Staphylo, @Junchao-Mellanox, @minionatwork: Please review again.

minionatwork (approved):

Junchao-Mellanox (approved):

Staphylo (approved):

jleveque:
@mprabhu-nokia: Please also add corresponding tests for these APIs to the platform API tests in the sonic-mgmt repo.

mprabhu-nokia:
> 
> 
> @mprabhu-nokia: Please also add corresponding tests for these APIs to the platform API tests in the sonic-mgmt repo.

Will do.

PR #141, opened 2020-11-13
[sonic_sfp]: Update config for multi-ASIC

mudsut4ke:
From this change https://github.com/Azure/sonic-platform-common/pull/100 , It missing mapping of logical port names to ASIC instance on device that used platform.json

So this new change will fix it by add logical_to_asic mapping dict to a condition for device that used _platform.json_

**Signed-off-by**: Wirut Getbamrung [wgetbumr@celestica.com]


judyjoseph (approved):
Looks good ! thanks

PR #143, opened 2020-11-16
[sonic-platform-common] Fix name error and import error 

Junchao-Mellanox:
Why I did this?

Found name error and import error when testing with python3 pmon daemons

How I fix this?

1. Fix name error in psu_base.py
2. change absolute import to relative import for sff8024

### DIFF
--- a/sonic_platform_base/psu_base.py
+++ b/sonic_platform_base/psu_base.py
@@ -27,7 +27,7 @@ class PsuBase(device_base.DeviceBase):
     _thermal_list = []
 
     # Status of Master LED
-    psu_master_led_color = device_base.DeviceBase.STATUS_LED_COLOR_OFF
+    psu_master_led_color = None
 
     def __init__(self):
         self._fan_list = []
@@ -35,6 +35,8 @@ class PsuBase(device_base.DeviceBase):
         # List of ThermalBase-derived objects representing all thermals
         # available on the PSU
         self._thermal_list = []
+
+        self.psu_master_led_color = self.STATUS_LED_COLOR_OFF
 
     def get_num_fans(self):
         """
--- a/sonic_platform_base/psu_base.py
+++ b/sonic_platform_base/psu_base.py
@@ -27,7 +27,7 @@ class PsuBase(device_base.DeviceBase):
     _thermal_list = []
 
     # Status of Master LED
-    psu_master_led_color = STATUS_LED_COLOR_OFF
+    psu_master_led_color = device_base.DeviceBase.STATUS_LED_COLOR_OFF
 
     def __init__(self):
         self._fan_list = []
--- a/sonic_platform_base/sonic_sfp/qsfp_dd.py
+++ b/sonic_platform_base/sonic_sfp/qsfp_dd.py
@@ -5,17 +5,17 @@
 from __future__ import print_function
 
 try:
-    from sff8024 import type_of_transceiver    # Dot module supports both Python 2 and Python 3 using explicit relative import methods
-    from sff8024 import type_abbrv_name    # Dot module supports both Python 2 and Python 3 using explicit relative import methods
-    from sff8024 import connector_dict    # Dot module supports both Python 2 and Python 3 using explicit relative import methods
-    from sff8024 import ext_type_of_transceiver    # Dot module supports both Python 2 and Python 3 using explicit relative import methods
-    from sff8024 import type_of_media_interface    # Dot module supports both Python 2 and Python 3 using explicit relative import methods
-    from sff8024 import host_electrical_interface    # Dot module supports both Python 2 and Python 3 using explicit relative import methods
-    from sff8024 import nm_850_media_interface    # Dot module supports both Python 2 and Python 3 using explicit relative import methods
-    from sff8024 import sm_media_interface    # Dot module supports both Python 2 and Python 3 using explicit relative import methods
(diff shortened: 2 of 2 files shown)


sonic_platform_base/psu_base.py
```diff
@@ -27,7 +27,7 @@ class PsuBase(device_base.DeviceBase):
     _thermal_list = []
 
     # Status of Master LED
-    psu_master_led_color = STATUS_LED_COLOR_OFF
+    psu_master_led_color = device_base.DeviceBase.STATUS_LED_COLOR_OFF
```
jleveque, line 30: Why is this necessary? Can we not use `self.` instead, since this constant should be inherited?

Junchao-Mellanox: Here is the class level variable and cannot use the key word ```self```  here.

jleveque: Have you tested this? I believe it should be possible via inheritance. See https://stackoverflow.com/a/10672749

Junchao-Mellanox: Yes, I tested it. A NameError will be reported. I did a small test just now:

```python
class A:
    VAR = 1


class B(A):
    VAR1 = VAR

b = B()
```

And the output:

```
  File "D:/code/test/test1.py", line 5, in <module>
    class B(A):
  File "D:/code/test/test1.py", line 6, in B
    VAR1 = VAR
NameError: name 'VAR' is not defined
```

jleveque: You didn't add `self.`. Try this:

```python
class A:
    VAR = 1


class B(A):
    VAR1 = self.VAR

b = B()
```

Junchao-Mellanox: Got the same:

```python
class A:
    VAR = 1


class B(A):
    VAR1 = self.VAR

b = B()
```

Output:
```
  File "D:/code/test/test1.py", line 5, in <module>
    class B(A):
  File "D:/code/test/test1.py", line 6, in B
    VAR1 = self.VAR
NameError: name 'self' is not defined
```

jleveque: Ah, because `self` is only available in a function. Can you set `psu_master_led_color = None` here and add `self.psu_master_led_color = self.STATUS_LED_COLOR_OFF` in the constructor?

Junchao-Mellanox: sure

jleveque: Actually, you can probably omit the `psu_master_led_color = None` line.

Junchao-Mellanox: Fixed

sonic_platform_base/psu_base.py
```diff
@@ -27,7 +27,7 @@ class PsuBase(device_base.DeviceBase):
     _thermal_list = []
 
     # Status of Master LED
-    psu_master_led_color = STATUS_LED_COLOR_OFF
+    psu_master_led_color = None
```
jleveque, line 30: I think we should remove this, as it makes it a class variable, which is shared between all instances. I also think we should do the same for `_fan_list` and `_thermal_list`, moving the comments into the constructor. Thoughts?

Junchao-Mellanox: I agree for _fan_list and _thermal_list. But I am not so sure about psu_master_led_color, it looks like a class level variable to me.

jleveque: I don't think we want `psu_master_led_color` to be class-level; each PSU instance should have it's own LED status.

Junchao-Mellanox: Understood.

Junchao-Mellanox: I saw some comments for the class level _thermal_list. So if vendor doesn't call PsuBase.__init__, it is an issue of vendor side, right?

```python
    # available on the PSU. Put a class level _thermal_list here to 
    # avoid an exception when call get_num_thermals, get_all_thermals
    # and get_thermal if vendor does not call PsuBase.__init__ in concrete
    # PSU class
    _thermal_list = []
```

jleveque: Hmm... Maybe we should leave this for now, then to prevent breaking anything. However, I do agree that it should be a vendor issue. We can look into doing this in a separate PR.

Junchao-Mellanox: It seems psu_master_led_color should be a class level variable because it is used in a classmethod. See: 

```python
    @classmethod
    def get_status_master_led(cls):
        """
        Gets the state of the Master status LED for a given device-type

        Returns:
            A string, one of the predefined STATUS_LED_COLOR_* strings.
        """
        return cls.psu_master_led_color
```

jleveque: It looks like related changes are also being made in https://github.com/Azure/sonic-platform-daemons/pull/119/files. For now, let's leave this alone and revisit it in a separate PR.

Junchao-Mellanox: Should I revert psu_base.py?

jleveque: No I think this is fine. We just won't worry about changing from class vars to instance vars in this PR.

jleveque (approved):

PR #157, opened 2021-01-04
Make sonic_sfp Python2 and Python3 compatible

msosyak:
Make `_read_eeprom_specific_bytes` Python3 and Python2 compatible

### DIFF
--- a/sonic_platform_base/sonic_sfp/sfputilbase.py
+++ b/sonic_platform_base/sonic_sfp/sfputilbase.py
@@ -338,8 +338,15 @@ class SfpUtilBase(object):
             return None
 
         try:
-            for n in range(0, num_bytes):
-                eeprom_raw[n] = hex(ord(raw[n]))[2:].zfill(2)
+            # in case raw is bytes (python3 is used) raw[n] will return int,
+            # and in case raw is str(python2 is used) raw[n] will return str,
+            # so for python3 the are no need to call ord to convert str to int.
+            if type(raw) == bytes:
+                for n in range(0, num_bytes):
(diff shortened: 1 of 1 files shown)


sonic_platform_base/sonic_sfp/sfputilbase.py
```diff
@@ -338,8 +338,15 @@ def _read_eeprom_specific_bytes(self, sysfsfile_eeprom, offset, num_bytes):
             return None
 
         try:
-            for n in range(0, num_bytes):
-                eeprom_raw[n] = hex(ord(raw[n]))[2:].zfill(2)
+            # in case raw is bytes (python3 is used) raw[n] will return int,
+            # and in case raw is str(python2 is used) raw[n] will return str,
+            # so for python3 the are no need to call ord to convert str to int.
```
jleveque, line 343: Thanks for this fix. Can you also please add a `#TODO: Remove this check once we no longer support Python 2` comment here for posterity?

jleveque (changes requested):

jleveque (changes requested):
I noticed there is also a call to `self.eep_dict.iteritems()` on line 746. Can you also please change this to `self.eep_dict.items()` to make it compatible with both Python 2 and 3?

msosyak:
> I noticed there is also a call to `self.eep_dict.iteritems()` on line 746. Can you also please change this to `self.eep_dict.items()` to make it compatible with both Python 2 and 3?

Done

sonic_platform_base/sonic_sfp/sfputilbase.py
msosyak: Done

jleveque (approved):

PR #161, opened 2021-02-04
[sonic-pcie] clean up on the message

sujinmkang:
Clean up the wrong pcieutil cli command information from sonic-pcie platform layer


jleveque:
Was the old message technically wrong, or was it just wordy?

sujinmkang:
@jleveque actually it's changed to 'pcieutil generate' from 'pcieutil pcie_generate' so it contains wrong information. And I don't think it's good to have the platform api layer inform about the utility cli command.  let's hold this PR. I will have a better solution than just removing the current message.

PR #162, opened 2021-02-04
[sonic_y_cable] add support for retrieving firmware info for Y cable, internal and nic temperature and voltage

vdahiya12:
Summary:
This PR provides the necessary infrastructure to add support for getting firmware information running on the three ends of Y cable. Also we get internal and nic side temperature and voltage 
### Type of change

- [ ] Bug fix
- [ ] Test case(new/improvement)

### Approach
Added changes in the sonic_y_cable package

To add the necessary support for  fetching firmware info for Y cable, internal and nic temperature and voltage

#### How did you do it?
Added the changes in sonic-platform-common module in the y_cable.py file

opened a python shell and ran the API's manually and test verified
the values are correct.

### DIFF
--- a/sonic_y_cable/y_cable.py
+++ b/sonic_y_cable/y_cable.py
@@ -852,7 +852,6 @@ def get_ber_info(physical_port, target):
     buffer = bytearray([target])
     curr_offset = OFFSET_TARGET
 
-
     ber_result = []
 
     if platform_chassis is not None:
@@ -1171,15 +1170,46 @@ def get_firmware_version(physical_port, target):
     if (rev_major_slot1 == 0 and rev_minor_slot1 == 0 and rev_build_lsb_slot1 == 0 and rev_build_msb_slot1 == 0 and rev_major_slot2 == 0 and rev_minor_slot2 == 0 and rev_build_lsb_slot2 == 0 and rev_build_msb_slot2 == 0):
         return None
     else:
-        build_slot1 = chr(rev_build_lsb_slot1)+ chr(rev_build_msb_slot1)
-        version_slot1 = str(rev_major_slot1) + "."+ str(rev_minor_slot1)
-        build_slot2 = chr(rev_build_lsb_slot2)+ chr(rev_build_msb_slot2)
-        version_slot2 = str(rev_major_slot2) + "."+ str(rev_minor_slot2)
-        result["build_slot1"] =  build_slot1
-        result["version_slot1"] =  version_slot1
-        result["build_slot2"] =  build_slot2
-        result["version_slot2"] =  version_slot2
+        build_slot1 = chr(rev_build_lsb_slot1) + chr(rev_build_msb_slot1)
+        version_slot1 = str(rev_major_slot1) + "." + str(rev_minor_slot1)
+        build_slot2 = chr(rev_build_lsb_slot2) + chr(rev_build_msb_slot2)
+        version_slot2 = str(rev_major_slot2) + "." + str(rev_minor_slot2)
+        if slot_status & 0x01:
+            run_slot1 = True
+        else:
+            run_slot1 = False
+        if slot_status & 0x02:
+            commit_slot1 = True
+        else:
+            commit_slot1 = False
+        if slot_status & 0x04:
+            empty_slot1 = True
+        else:
+            empty_slot1 = False
+        if slot_status & 0x10:
+            run_slot2 = True
+        else:
+            run_slot2 = False
+        if slot_status & 0x20:
+            commit_slot2 = True
+        else:
+            commit_slot2 = False
+        if slot_status & 0x40:
+            empty_slot2 = True
+        else:
+            empty_slot2 = False
+
+        result["build_slot1"] = build_slot1
+        result["version_slot1"] = version_slot1
+        result["build_slot2"] = build_slot2
+        result["version_slot2"] = version_slot2
         result["slot_status"] = slot_status
+        result["run_slot1"] = run_slot1
+        result["run_slot2"] = run_slot2
+        result["commit_slot1"] = commit_slot1
+        result["commit_slot2"] = commit_slot2
+        result["empty_slot1"] = empty_slot1
+        result["empty_slot2"] = empty_slot2
 
     return result
 
--- a/sonic_y_cable/y_cable.py
+++ b/sonic_y_cable/y_cable.py
@@ -1157,8 +1157,7 @@ def get_firmware_version(physical_port, target):
 
     result = {}
     NUM_MCU_SIDE = 3
-    addr = target * (FIRMWARE_INFO_PAYLOAD_SIZE / NUM_MCU_SIDE)
-    base_addr = int(addr)
+    base_addr = int(target * (FIRMWARE_INFO_PAYLOAD_SIZE / NUM_MCU_SIDE))
     rev_major_slot1 = struct.unpack_from('<B', data[0 + base_addr:  1 + base_addr])[0]
     rev_minor_slot1 = struct.unpack_from('<B', data[2 + base_addr:  3 + base_addr])[0]
     rev_build_lsb_slot1 = struct.unpack_from('<B', data[4 + base_addr:  5 + base_addr])[0]
@@ -1172,14 +1171,14 @@ def get_firmware_version(physical_port, target):
     if (rev_major_slot1 == 0 and rev_minor_slot1 == 0 and rev_build_lsb_slot1 == 0 and rev_build_msb_slot1 == 0 and rev_major_slot2 == 0 and rev_minor_slot2 == 0 and rev_build_lsb_slot2 == 0 and rev_build_msb_slot2 == 0):
         return None
     else:
-        result["rev_major_slot1"] = rev_major_slot1
-        result["rev_minor_slot1"] = rev_minor_slot1
-        result["rev_build_lsb_slot1"] = rev_build_lsb_slot1
-        result["rev_build_msb_slot1"] = rev_build_msb_slot1
-        result["rev_major_slot2"] = rev_major_slot2
-        result["rev_minor_slot2"] = rev_minor_slot2
-        result["rev_build_lsb_slot2"] = rev_build_lsb_slot2
-        result["rev_build_msb_slot2"] = rev_build_msb_slot2
+        build_slot1 = chr(rev_build_lsb_slot1)+ chr(rev_build_msb_slot1)
+        version_slot1 = str(rev_major_slot1) + "."+ str(rev_minor_slot1)
+        build_slot2 = chr(rev_build_lsb_slot2)+ chr(rev_build_msb_slot2)
+        version_slot2 = str(rev_major_slot2) + "."+ str(rev_minor_slot2)
+        result["build_slot1"] =  build_slot1
+        result["version_slot1"] =  version_slot1
+        result["build_slot2"] =  build_slot2
+        result["version_slot2"] =  version_slot2
         result["slot_status"] = slot_status
 
     return result
--- a/sonic_y_cable/y_cable.py
+++ b/sonic_y_cable/y_cable.py
@@ -87,7 +87,10 @@ EEPROM_READ_DATA_INVALID = -1
 EEPROM_ERROR = -1
 EEPROM_TIMEOUT_ERROR = -1
 
-WAIT_TIMEOUT_FOR_BER_EYE = 1
+BER_TIMEOUT_SECS = 1
+EYE_TIMEOUT_SECS = 1
+
+MAX_NUM_LANES = 4
 
 SYSLOG_IDENTIFIER = "sonic_y_cable"
 
@@ -874,13 +877,12 @@ def get_ber_info(physical_port, target):
             time_diff = time_now - time_start
             if done[0] == 1:
                 break
-            elif time_diff >= WAIT_TIMEOUT_FOR_BER_EYE:
+            elif time_diff >= BER_TIMEOUT_SECS:
                 return EEPROM_TIMEOUT_ERROR
 
         idx = 0
-        maxLane = 4
         curr_offset = OFFSET_LANE_1_BER_RESULT
-        for lane in range(maxLane):
+        for lane in range(MAX_NUM_LANES):
             msb_result = platform_chassis.get_sfp(physical_port).read_eeprom(curr_offset+idx, 1)
             if y_cable_validate_read_data(msb_result, 1, physical_port, "BER data msb result") == EEPROM_READ_DATA_INVALID:
                 return EEPROM_ERROR
@@ -952,12 +954,11 @@ def get_eye_info(physical_port, target):
             time_diff = time_now - time_start
             if done[0] == 1:
                 break
-            elif time_diff >= WAIT_TIMEOUT_FOR_BER_EYE:
+            elif time_diff >= EYE_TIMEOUT_SECS:
                 return EEPROM_TIMEOUT_ERROR
 
         idx = 0
-        maxLane = 4
-        for lane in range(maxLane):
+        for lane in range(MAX_NUM_LANES):
             curr_offset = OFFSET_LANE_1_EYE_RESULT
             msb_result = platform_chassis.get_sfp(physical_port).read_eeprom(curr_offset+idx, 1)
             if y_cable_validate_read_data(msb_result, 1, physical_port, "EYE data msb result") == EEPROM_READ_DATA_INVALID:
@@ -1144,12 +1145,12 @@ def get_firmware_version(physical_port, target):
     data = bytearray(FIRMWARE_INFO_PAYLOAD_SIZE)
 
     if platform_chassis is not None:
-        for byteIdx in range(0, FIRMWARE_INFO_PAYLOAD_SIZE):
-            curr_offset = 0xfc * 128 + 128 + byteIdx
-            readOut = platform_chassis.get_sfp(physical_port).read_eeprom(curr_offset, 1)
-            if y_cable_validate_read_data(readOut, 1, physical_port, "firmware info") == EEPROM_READ_DATA_INVALID:
+        for byte_idx in range(0, FIRMWARE_INFO_PAYLOAD_SIZE):
+            curr_offset = 0xfc * 128 + 128 + byte_idx
+            read_out = platform_chassis.get_sfp(physical_port).read_eeprom(curr_offset, 1)
+            if y_cable_validate_read_data(read_out, 1, physical_port, "firmware info") == EEPROM_READ_DATA_INVALID:
                 return EEPROM_ERROR
-            data[byteIdx] = readOut[0]
+            data[byte_idx] = read_out[0]
     else:
         helper_logger.log_error("platform_chassis is not loaded, failed to get NIC lanes active")
         return -1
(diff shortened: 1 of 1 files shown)


sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,104 @@ def check_if_nic_lanes_active(physical_port):
...
+    rev_build_lsb_slot1 = struct.unpack_from('<B', data[4 + baseAddr:  5 + baseAddr])[0]
+    rev_build_msb_slot1 = struct.unpack_from('<B', data[5 + baseAddr:  6 + baseAddr])[0]
+    rev_major_slot2 = struct.unpack_from('<B', data[1 + baseAddr:  2 + baseAddr])[0]
+    rev_minor_slot2 = struct.unpack_from('<B', data[3 + baseAddr:  4 + baseAddr])[0]
+    rev_build_lsb_slot2 = struct.unpack_from('<B', data[6 + baseAddr:  7 + baseAddr])[0]
+    rev_build_msb_slot2 = struct.unpack_from('<B', data[7 + baseAddr:  8 + baseAddr])[0]
+    slotStatus = struct.unpack_from('<B', data[8 + baseAddr:  9 + baseAddr])[0]
+
+    if (rev_major_slot1 == 0 and rev_minor_slot1 == 0 and rev_build_lsb_slot1 == 0 and rev_build_msb_slot1 == 0 and rev_major_slot2 == 0 and rev_minor_slot2 == 0 and rev_build_lsb_slot2 == 0 and rev_build_msb_slot2 == 0):
+        return -1
```
jleveque, line 1173: Prefer not to have different return value types (here list/int). Suggest either returning `None` on failure.

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,104 @@ def check_if_nic_lanes_active(physical_port):
...
+    else:
+        result.append(rev_major_slot1)
+        result.append(rev_minor_slot1)
+        result.append(rev_build_lsb_slot1)
```
jleveque, line 1177: Where is `rev_build_msb_slot1`?

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,104 @@ def check_if_nic_lanes_active(physical_port):
...
+    baseAddr = int(Addr)
+    rev_major_slot1 = struct.unpack_from('<B', data[0 + baseAddr:  1 + baseAddr])[0]
+    rev_minor_slot1 = struct.unpack_from('<B', data[2 + baseAddr:  3 + baseAddr])[0]
+    rev_build_lsb_slot1 = struct.unpack_from('<B', data[4 + baseAddr:  5 + baseAddr])[0]
+    rev_build_msb_slot1 = struct.unpack_from('<B', data[5 + baseAddr:  6 + baseAddr])[0]
+    rev_major_slot2 = struct.unpack_from('<B', data[1 + baseAddr:  2 + baseAddr])[0]
+    rev_minor_slot2 = struct.unpack_from('<B', data[3 + baseAddr:  4 + baseAddr])[0]
+    rev_build_lsb_slot2 = struct.unpack_from('<B', data[6 + baseAddr:  7 + baseAddr])[0]
+    rev_build_msb_slot2 = struct.unpack_from('<B', data[7 + baseAddr:  8 + baseAddr])[0]
+    slotStatus = struct.unpack_from('<B', data[8 + baseAddr:  9 + baseAddr])[0]
```
jleveque, line 1170: Suggest renaming `slot_status`.

Also, what is this value? Does it represent the active slot?

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,104 @@ def check_if_nic_lanes_active(physical_port):
...
+            curr_offset = 0xfc * 128 + 128 + byteIdx
+            readOut = platform_chassis.get_sfp(physical_port).read_eeprom(curr_offset, 1)
+            if y_cable_validate_read_data(readOut, 1, physical_port, "firmware info") == EEPROM_READ_DATA_INVALID:
+                return EEPROM_ERROR
+            data[byteIdx] = readOut[0]
+    else:
+        helper_logger.log_error("platform_chassis is not loaded, failed to get NIC lanes active")
+        return -1
+
+    result = []
```
jleveque, line 1158: A list of all these values will be difficult/confusing to consume. Suggest a dict instead with descriptive keys.

Also, I wonder if we should return all of these values individually like this, or concatenate them into strings?

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,104 @@ def check_if_nic_lanes_active(physical_port):
...
+    NUM_MCU_SIDE = 3
+    Addr = target * (FIRMWARE_INFO_PAYLOAD_SIZE / NUM_MCU_SIDE)
+    baseAddr = int(Addr)
```
jleveque, line 1161: Rename `base_addr`

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,104 @@ def check_if_nic_lanes_active(physical_port):
...
+            if y_cable_validate_read_data(readOut, 1, physical_port, "firmware info") == EEPROM_READ_DATA_INVALID:
+                return EEPROM_ERROR
+            data[byteIdx] = readOut[0]
+    else:
+        helper_logger.log_error("platform_chassis is not loaded, failed to get NIC lanes active")
+        return -1
+
+    result = []
+    NUM_MCU_SIDE = 3
+    Addr = target * (FIRMWARE_INFO_PAYLOAD_SIZE / NUM_MCU_SIDE)
```
jleveque, line 1160: Rename `addr`

jleveque (changes requested):

sonic_y_cable/y_cable.py
vdahiya12: added, fixed

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
vdahiya12: actually slot_status would be used by api caller to determine which slot has run, commit and empty. Something like this. 
`if slot_status & 0x01: runSlot1    = '[*]'
        else:                 runSlot1    = '[ ]'
        if slot_status & 0x02: commitSlot1 = '[*]'
        else:                 commitSlot1 = '[ ]'
        if slot_status & 0x04: emptySlot1  = '[*]'
        else:                 emptySlot1  = '[ ]'
        if slot_status & 0x10: runSlot2    = '[*]'
        else:                 runSlot2    = '[ ]'
        if slot_status & 0x20: commitSlot2 = '[*]'
        else:                 commitSlot2 = '[ ]'
        if slot_status & 0x40: emptySlot2  = '[*]'
        else:                 emptySlot2  = '[ ]'`

sonic_y_cable/y_cable.py
vdahiya12: changed it into dictionary. 
 strings would be better less interpretation by the caller

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
jleveque: > strings would be better less interpretation by the caller

Exactly. We just need to determine if the caller would ever need this granularity. If not, then we could return strings.

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,105 @@ def check_if_nic_lanes_active(physical_port):
...
+                return EEPROM_ERROR
+            data[byteIdx] = readOut[0]
+    else:
+        helper_logger.log_error("platform_chassis is not loaded, failed to get NIC lanes active")
+        return -1
+
+    result = {}
+    NUM_MCU_SIDE = 3
+    addr = target * (FIRMWARE_INFO_PAYLOAD_SIZE / NUM_MCU_SIDE)
+    base_addr = int(addr)
```
jleveque, line 1161: Since addr is not used elsewhere, I suggest combining these lines:

```python
    base_addr = int(target * (FIRMWARE_INFO_PAYLOAD_SIZE / NUM_MCU_SIDE))
```

jleveque (changes requested):

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
vdahiya12: fixed
sample output dict
`{'build_slot1': 'MS', 'version_slot1': '0.6', 'build_slot2': 'MS', 'version_slot2': '0.6', 'slot_status': 3}`

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,104 @@ def check_if_nic_lanes_active(physical_port):
...
+    else:
+        build_slot1 = chr(rev_build_lsb_slot1)+ chr(rev_build_msb_slot1)
+        version_slot1 = str(rev_major_slot1) + "."+ str(rev_minor_slot1)
+        build_slot2 = chr(rev_build_lsb_slot2)+ chr(rev_build_msb_slot2)
+        version_slot2 = str(rev_major_slot2) + "."+ str(rev_minor_slot2)
+        result["build_slot1"] =  build_slot1
+        result["version_slot1"] =  version_slot1
+        result["build_slot2"] =  build_slot2
+        result["version_slot2"] =  version_slot2
+        result["slot_status"] = slot_status
```
jleveque, line 1182: Does it make more sense to translate this value here? Something like:

```python
result["active_slot1"] = True/False
result["committed_slot1"] = True/False
...
```

jleveque (changes requested):

sonic_y_cable/y_cable.py
vdahiya12: fixed with the way intended

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,104 @@ def check_if_nic_lanes_active(physical_port):
...
+    rev_build_msb_slot2 = struct.unpack_from('<B', data[7 + base_addr:  8 + base_addr])[0]
+    slot_status = struct.unpack_from('<B', data[8 + base_addr:  9 + base_addr])[0]
+
+    if (rev_major_slot1 == 0 and rev_minor_slot1 == 0 and rev_build_lsb_slot1 == 0 and rev_build_msb_slot1 == 0 and rev_major_slot2 == 0 and rev_minor_slot2 == 0 and rev_build_lsb_slot2 == 0 and rev_build_msb_slot2 == 0):
+        return None
+    else:
+        build_slot1 = chr(rev_build_lsb_slot1)+ chr(rev_build_msb_slot1)
+        version_slot1 = str(rev_major_slot1) + "."+ str(rev_minor_slot1)
+        build_slot2 = chr(rev_build_lsb_slot2)+ chr(rev_build_msb_slot2)
+        version_slot2 = str(rev_major_slot2) + "."+ str(rev_minor_slot2)
```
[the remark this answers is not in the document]
vdahiya12: fixed

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1128,135 @@ def check_if_nic_lanes_active(physical_port):
...
+    base_addr = int(target * (FIRMWARE_INFO_PAYLOAD_SIZE / NUM_MCU_SIDE))
+    rev_major_slot1 = struct.unpack_from('<B', data[0 + base_addr:  1 + base_addr])[0]
+    rev_minor_slot1 = struct.unpack_from('<B', data[2 + base_addr:  3 + base_addr])[0]
+    rev_build_lsb_slot1 = struct.unpack_from('<B', data[4 + base_addr:  5 + base_addr])[0]
+    rev_build_msb_slot1 = struct.unpack_from('<B', data[5 + base_addr:  6 + base_addr])[0]
+    rev_major_slot2 = struct.unpack_from('<B', data[1 + base_addr:  2 + base_addr])[0]
+    rev_minor_slot2 = struct.unpack_from('<B', data[3 + base_addr:  4 + base_addr])[0]
+    rev_build_lsb_slot2 = struct.unpack_from('<B', data[6 + base_addr:  7 + base_addr])[0]
+    rev_build_msb_slot2 = struct.unpack_from('<B', data[7 + base_addr:  8 + base_addr])[0]
+    slot_status = struct.unpack_from('<B', data[8 + base_addr:  9 + base_addr])[0]
```
jleveque, line 1168: Regarding style, there should be no spaces after `:` for list slicing. Suggest the following:

```python
rev_major_slot1 = struct.unpack_from('<B', data[(0 + base_addr):(1 + base_addr)])[0]
rev_minor_slot1 = struct.unpack_from('<B', data[(2 + base_addr):(3 + base_addr)])[0]
...
```

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1128,135 @@ def check_if_nic_lanes_active(physical_port):
...
+        result["version_slot1"] = version_slot1
+        result["build_slot2"] = build_slot2
+        result["version_slot2"] = version_slot2
+        result["slot_status"] = slot_status
+        result["run_slot1"] = run_slot1
+        result["run_slot2"] = run_slot2
+        result["commit_slot1"] = commit_slot1
+        result["commit_slot2"] = commit_slot2
+        result["empty_slot1"] = empty_slot1
+        result["empty_slot2"] = empty_slot2
```
jleveque, line 1212: Suggest reducing number of lines here:

```python
        result["build_slot1"] = chr(rev_build_lsb_slot1) + chr(rev_build_msb_slot1)
        result["version_slot1"] = str(rev_major_slot1) + "." + str(rev_minor_slot1)
        result["build_slot2"] = chr(rev_build_lsb_slot2) + chr(rev_build_msb_slot2)
        result["version_slot2"] = str(rev_major_slot2) + "." + str(rev_minor_slot2)
        result["run_slot1"] = True if slot_status & 0x01 else False
        result["commit_slot1"] = True if slot_status & 0x02 else False
        ...
```

jleveque (changes requested):

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
jleveque: Thanks. I see you didn't add parentheses around the slice values. While not necessary, I added them in my suggestion because I felt they made it more readable. What do you think?

sonic_y_cable/y_cable.py
jleveque: Thanks. I see you retained the intermediate variables (e.g., `build_slot1`). They only get assigned and then are assigned to `result["xxxx"]`, so they seem unnecessary and just double the lines of code needed. I suggest following my example above and assign the values directly into the respective `result` element.

sonic_y_cable/y_cable.py
vdahiya12: fixed, its more readable now

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
```diff
@@ -75,6 +80,15 @@
...
 SWITCH_COUNT_AUTO = "auto"
 
+FIRMWARE_INFO_PAYLOAD_SIZE = 48
+NUM_MCU_SIDE = 3
+
+EEPROM_READ_DATA_INVALID = -1
+EEPROM_ERROR = -1
+EEPROM_TIMEOUT_ERROR = -1
+
+WAIT_TIMEOUT_FOR_BER_EYE = 1
```
jleveque, line 90: Suggest breaking this into two, one for BER, one for eye, and also add the units in the name (e.g., `BER_TIMEOUT_SECS`, `EYE_TIMEOUT_SECS`).

sonic_y_cable/y_cable.py
```diff
@@ -856,20 +865,28 @@ def get_ber_info(physical_port, target):
...
+            time_now = time.time()
+            time_diff = time_now - time_start
             if done[0] == 1:
                 break
+            elif time_diff >= WAIT_TIMEOUT_FOR_BER_EYE:
+                return EEPROM_TIMEOUT_ERROR
 
         idx = 0
-        maxLane = 2
+        maxLane = 4
```
jleveque, line 881: Suggest making this a constant and defining at top of file, since it is used in multiple places.

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1128,112 @@ def check_if_nic_lanes_active(physical_port):
...
 
     return result
+
+
+def get_firmware_version(physical_port, target):
+
+    data = bytearray(FIRMWARE_INFO_PAYLOAD_SIZE)
+
+    if platform_chassis is not None:
+        for byteIdx in range(0, FIRMWARE_INFO_PAYLOAD_SIZE):
```
jleveque, line 1147: s/byteIdx/byte_idx/

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1128,112 @@ def check_if_nic_lanes_active(physical_port):
...
+            curr_offset = 0xfc * 128 + 128 + byteIdx
+            readOut = platform_chassis.get_sfp(physical_port).read_eeprom(curr_offset, 1)
```
jleveque, line 1149: s/readOut/read_out/

jleveque (changes requested):

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
```diff
@@ -1097,11 +1129,112 @@ def check_if_nic_lanes_active(physical_port):
...
+        build_slot1 = chr(rev_build_lsb_slot1) + chr(rev_build_msb_slot1)
+        version_slot1 = str(rev_major_slot1) + "." + str(rev_minor_slot1)
+        build_slot2 = chr(rev_build_lsb_slot2) + chr(rev_build_msb_slot2)
+        version_slot2 = str(rev_major_slot2) + "." + str(rev_minor_slot2)
+
+        result["build_slot1"] = build_slot1
+        result["version_slot1"] = version_slot1
+        result["build_slot2"] = build_slot2
+        result["version_slot2"] = version_slot2
+        result["slot_status"] = slot_status
```
jleveque, line 1184: Do we need `result["slot_status"]` anymore, since we are breaking it apart below?

jleveque (changes requested):

sonic_y_cable/y_cable.py
vdahiya12: removed it for now, will update again if required

jleveque (approved):

PR #163, opened 2021-02-05
[sonic_sfp/qsfp_dd.py] Update DOM capability method name to align with other drivers

jleveque:
Update method name to align with other drivers. Change from `parse_qsfp_dom_capability` to `parse_dom_capability`. Other drivers were aligned via https://github.com/Azure/sonic-mgmt/pull/2937.

When this submodule is updated in sonic-buildimage, all callers will need to be updated to reflect the new nomenclature.

### DIFF
--- a/sonic_platform_base/sonic_sfp/qsfp_dd.py
+++ b/sonic_platform_base/sonic_sfp/qsfp_dd.py
@@ -223,8 +223,8 @@ class qsfp_dd_InterfaceId(sffbase):
     def parse_cable_len(self, cable_len_data, start_pos):
         return sffbase.parse(self, self.cable_len, cable_len_data, start_pos)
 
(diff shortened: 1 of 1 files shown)


jleveque:
@keboliu, @shlomibitton: Please review.

shlomibitton:
@jleveque Need to update the mlnx-platform-api with this change
https://github.com/Azure/sonic-buildimage/blob/834347b8f793ca60e3b0b6a91654259006c4df7b/platform/mellanox/mlnx-platform-api/sonic_platform/sfp.py#L485

jleveque:
@shlomibitton: Yes, I am aware of this. Another contributor made a similar modification in https://github.com/Azure/sonic-mgmt/pull/2937, and I needed to do the same. Since that PR has merged, I believe we should align the method name everywhere. I will make the related changes with the submodule update.

Issue #160, opened 2021-01-28
sffbase.py convert_hex_to_string method is throwing exception on str.strip operation after Python 3 migration


snider-nokia:
**return str.strip(binascii.unhexlify(ret_str))** is causing an exception such that SFP eeprom fields are not being populated properly.  This occurs because the str.strip exception is being returned from convert_hex_to_string instead of the properly converted value.

jleveque:
@snider-nokia: I just found this issue myself a few days ago and opened [this PR](https://github.com/Azure/sonic-platform-common/pull/165) to fix. Please help test and review it.

PR #165, opened 2021-02-06
[sffbase] Make convert_hex_to_string() compatible with both Python 2 and Python 3

jleveque:
When called with Python 3, `str.strip(binascii.unhexlify(ret_str))` will raise an exception: ` descriptor 'strip' requires a 'str' object but received a 'bytes'`. This patch (`binascii.unhexlify(ret_str).decode("utf-8").strip()`) will work properly with both Python 2 and Python 3.

Fixes https://github.com/Azure/sonic-platform-common/issues/160


jleveque:
@snider-nokia: Please test and review.

PR #173, opened 2021-02-23
[sonic_platform_base] Proper use of class and instance attributes

jleveque:
#### Description
- Use class and instance attributes properly. Only use class attributes for data which should be shared among all instances.
- Import all modules in `__init__.py`
- Remove unused imports
- Clean up whitespace

#### Motivation and Context
Proper object-oriented programming. This could prevent issues where attributes are shared between instances when they were not meant to be.

This also fixes a bug with `PsuBase.psu_master_led_color`, where it is defined as both a class attribute and an instance attribute.

With this change, some vendors' APIs can be cleaned up, because they redefined class attributes as instance attributes. That code can be removed.

However, note that vendor APIs MUST call the base class initializer in their initializer methods, or the instance attributes will not be available and will raise exceptions. 

Tested on one vendor's platform


jleveque:
@Staphylo, @zzhiyuan: FYI. Please make sure Arista concrete platform API classes all call the base (super) class initializer. I went through and took care of a few other platforms in the sonic-buildimage repo:

- https://github.com/Azure/sonic-buildimage/pull/6852
- https://github.com/Azure/sonic-buildimage/pull/6853
- https://github.com/Azure/sonic-buildimage/pull/6854

sonic_platform_base/psu_base.py
```diff
@@ -15,28 +15,22 @@ class PsuBase(device_base.DeviceBase):
...
+        # List of FanBase-derived objects representing all fans
+        # available on the PSU
         self._fan_list = []
 
         # List of ThermalBase-derived objects representing all thermals
         # available on the PSU
         self._thermal_list = []
 
-        self.psu_master_led_color = self.STATUS_LED_COLOR_OFF
+        PsuBase._psu_master_led_color = self.STATUS_LED_COLOR_OFF
```
mprabhu-nokia, line 33: @jleveque Thanks for addressing this!

Staphylo:
@jleveque thanks for the heads up. We'll make the changes.

PR #176, opened 2021-03-11
[y_cable] add support for enable/disable autoswitch feature on Y cable

vdahiya12:
This PR adds functions definition with details for enabling/disabling Y cable auto-switch feature

#### Description
Added functions for enabling/disabling auto-switch on Y cable
```
def set_switching_mode(physical_port, mode)
def get_switching_mode(physical_port)
```

#### Motivation and Context

We need to have this feature within host cli so that a user can enable/disable the autoswitch on Y cable

opened a python shell and executed API's for correctness

### DIFF
--- a/sonic_y_cable/y_cable.py
+++ b/sonic_y_cable/y_cable.py
@@ -92,9 +92,9 @@ EYE_TIMEOUT_SECS = 1
 
 MAX_NUM_LANES = 4
 
-# definitions for switching modes inside muxcable
-SWITCHING_MODE_AUTO = 1
+# switching modes inside muxcable
 SWITCHING_MODE_MANUAL = 0
+SWITCHING_MODE_AUTO = 1
 
 # Valid return codes for upgrade firmware routine steps
 FIRMWARE_DOWNLOAD_SUCCESS = 0
--- a/sonic_y_cable/y_cable.py
+++ b/sonic_y_cable/y_cable.py
@@ -52,6 +52,7 @@ OFFSET_INTERNAL_TEMPERATURE = 22
 OFFSET_INTERNAL_VOLTAGE = 26
 OFFSET_NIC_TEMPERATURE = 727
 OFFSET_NIC_VOLTAGE = 729
+OFFSET_ENABLE_AUTO_SWITCH = 651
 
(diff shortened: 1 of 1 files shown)


sonic_y_cable/y_cable.py
```diff
@@ -1401,3 +1408,84 @@ def rollback_firmware(physical_port):
     """
 
     return FIRMWARE_ROLLBACK_SUCCESS
+
+
+def enable_auto_switch(physical_port, mode):
```
jleveque, line 1413: Wondering whether we should call this something like `set_switching_mode()` and pass in a constant like `SWITCHING_MODE_AUTO` or `SWITCHING_MODE_MANUAL`. We would then change `check_if_auto_switch_enabled()` to `get_switching_mode()` and it would return one of these constants.

vdahiya12: fixed

sonic_y_cable/y_cable.py
```diff
@@ -92,9 +92,9 @@
 
 MAX_NUM_LANES = 4
 
-# definitions for enable/disable mode for autoswitch feature
-ENABLE_AUTO_SWITCH = 1
-DISABLE_AUTO_SWITCH = 0
+# definitions for switching modes inside muxcable
```
jleveque, line 95: Remove "definitions for"; unnecessarily verbose

sonic_y_cable/y_cable.py
```diff
@@ -92,9 +92,9 @@
...
+SWITCHING_MODE_AUTO = 1
+SWITCHING_MODE_MANUAL = 0
```
jleveque, line 97: Suggest arranging these in numerical order

jleveque (changes requested):

sonic_y_cable/y_cable.py
vdahiya12: fixed

sonic_y_cable/y_cable.py
vdahiya12: fixed

jleveque (approved):

PR #177, opened 2021-03-11
Change import order in Ycable helper and EEPROM read bytearray change in SFP plugin

vpsubramaniam:
#### Description

- Change import order in Ycable helper
- Change data read from EEPROM to bytearray in _read_eeprom_specific_bytes for python 2 and 3 compatibility.

#### Motivation and Context

https://github.com/Azure/sonic-buildimage/issues/6820
- Import of sonic_platform before logger when platform API is not available leads to non import of logger resulting in xcvrd crash
- Since string and byte comparison returns true in python2 current check for differentiating python2 and 3 in EEPROM read fails.

Verified that python 2 xcvrd runs without exiting in DellEMC S5248F platform.
Logs: [Unit testing logs.txt](https://github.com/Azure/sonic-platform-common/files/6123278/Unit.testing.logs.txt)

### DIFF
--- a/sonic_y_cable/y_cable.py
+++ b/sonic_y_cable/y_cable.py
@@ -10,8 +10,8 @@ try:
     import struct
     from ctypes import c_int8
 
-    import sonic_platform.platform
     from sonic_py_common import logger
+    import sonic_platform.platform
 except ImportError as e:
     # When build python3 xcvrd, it tries to do basic check which will import this file. However,
     # not all platform supports python3 API now, so it could cause an issue when importing
--- a/sonic_platform_base/sonic_sfp/sfputilbase.py
+++ b/sonic_platform_base/sonic_sfp/sfputilbase.py
@@ -344,16 +344,10 @@ class SfpUtilBase(object):
             return None
 
         try:
-            # in case raw is bytes (python3 is used) raw[n] will return int,
-            # and in case raw is str(python2 is used) raw[n] will return str,
-            # so for python3 the are no need to call ord to convert str to int.
-            # TODO: Remove this check once we no longer support Python 2
-            if type(raw) == bytes:
-                for n in range(0, num_bytes):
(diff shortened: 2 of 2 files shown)


sonic_y_cable/y_cable.py
```diff
@@ -10,8 +10,8 @@
     import struct
     from ctypes import c_int8
 
-    import sonic_platform.platform
     from sonic_py_common import logger
+    import sonic_platform.platform
```
jleveque, line 14: I don't understand why this order makes a difference. You state in the PR description,

>Import of sonic_platform before logger when platform API is not available leads to non import of logger resulting in xcvrd crash

Where does the "non import of logger" occur?

ArunSaravananBalachandran: In case platform API is not available, it results in the import of logger in the subsequent line not being performed and thus accessing it later as below raises an Exception:   
https://github.com/Azure/sonic-platform-common/blob/10dc16f40baa0867da21fb31c089b73d33cc4107/sonic_y_cable/y_cable.py#L110

vdahiya12 (approved):

sonic_y_cable/y_cable.py
jleveque: Under what circumstances is the platform API not available? I think maybe we need to rethink/refactor something here, rather than simply change the order of imports.

ArunSaravananBalachandran: This issue is seen in platforms, in which the platform API is currently not available/implemented.

jleveque: I see. I was hoping this was the answer. As of now, we expect all platforms to have sonic-platform package support, rather than the old plugins. So I'm OK with taking this change for now to give vendors a bit more time to move over, but this is not the standard import ordering, so I intend to change it back in the future.

jleveque (approved):

PR #182, opened 2021-04-02
[Y-Cable] refactor get_firmware_version to comply with all vendors

vdahiya12:
This PR refactors the get_firmware_version output to return the following fields for each target. 
```
{
"version_active": "",
"version_inactive": "",
"version_next": "",
```
So by calling this for all the 3 MCU's TOR1, TOR2 and NIC we can get the below result. 

```
{
"version_nic_active": "0.6MS",
"version_nic_inactive": "0.5MS",
"version_nic_next": "0.6MS",
"version_self_active": "0.5MS",
"version_self_inactive": "0.6MS",
"version_self_next": "0.6MS",
"version_peer_active": "0.6MS",
"version_peer_inactive": "0.6MS",
"version_peer_next": "0.6MS",
} 
```

#### Motivation and Context
Required to refactor get_firmware_version output to comply with all vendors output.

### DIFF
--- a/sonic_y_cable/y_cable.py
+++ b/sonic_y_cable/y_cable.py
@@ -1240,8 +1240,7 @@ def get_firmware_version(physical_port, target):
 
         result["version_active"] = version_build_slot1 if slot_status & 0x01 else version_build_slot2
         result["version_inactive"] = version_build_slot2 if slot_status & 0x01 else version_build_slot1
-        result["version_next"] = version_build_slot1 if slot_status & 0x02
-        result["version_next"] = version_build_slot2 if slot_status & 0x20
+        result["version_next"] = version_build_slot1 if slot_status & 0x02 else version_build_slot2
 
 
     return result
--- a/sonic_y_cable/y_cable.py
+++ b/sonic_y_cable/y_cable.py
@@ -1224,6 +1224,10 @@ def get_firmware_version(physical_port, target):
         build_slot2 = chr(rev_build_lsb_slot2) + chr(rev_build_msb_slot2)
         version_slot2 = str(rev_major_slot2) + "." + str(rev_minor_slot2)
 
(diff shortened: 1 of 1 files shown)


sonic_y_cable/y_cable.py
```diff
@@ -1235,6 +1235,14 @@ def get_firmware_version(physical_port, target):
         result["empty_slot1"] = True if slot_status & 0x04 else False
         result["empty_slot2"] = True if slot_status & 0x40 else False
 
+        version_build_slot1 = version_slot1 + build_slot1
+        version_build_slot2 = version_slot2 + build_slot2
+
+        result["version_active"] = version_build_slot1 if slot_status & 0x01 else version_build_slot2
+        result["version_inactive"] = version_build_slot2 if slot_status & 0x01 else version_build_slot1
+        result["version_next"] = version_build_slot1 if slot_status & 0x02 else version_build_slot2
```
jleveque, line 1243: These new fields are intended to supplant the fields on lines 1227-1236, simplifying the resulting structure. Is there a reason you did not delete lines 1227-1236?

vdahiya12: just thinking to keep them for now, until xcvrd has the change where it does not look for old fields and fetches these new fields, otherwise we might see regression, if we remove the old fields without making the change in xcvrd. Once xcvrd is updated I will have another PR to remove these old fields. 
The other approach is to make the change in xcvrd and y_cable.py and make the submodule ptr update together in one PR for both sonic-platform-common and sonic-platform-daemons, so that nothing breaks. 
What do you suggest ?

jleveque: Makes sense to allow for a smooth transition. Maybe add a TODO comment to remove them eventually?

jleveque (approved):

PR #184, opened 2021-04-15
 [DeviceBase] Added hardware revision number to generic device properties

robocoder99:
#### Description
Added a `get_revision()` method to the generic platform device class which returns the hardware revision number of the device as a string. 

#### Motivation and Context
Exposing the hardware revision number to the platform API allows the user to precisely determine the FRU hardware components installed for debugging and inventory tracking purposes.

### DIFF
--- a/sonic_platform_base/device_base.py
+++ b/sonic_platform_base/device_base.py
@@ -55,6 +55,15 @@ class DeviceBase(object):
         """
         raise NotImplementedError
 
+    def get_revision(self):
+        """
+        Retrieves the hardware revision number of the device
+
+        Returns:
+            string: Revision number of device
+        """
+        raise NotImplementedError
+
     def get_status(self):
         """
         Retrieves the operational status of the device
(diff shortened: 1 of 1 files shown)


sonic_platform_base/device_base.py
```diff
@@ -55,6 +55,15 @@ def get_serial(self):
         """
         raise NotImplementedError
 
+    def get_revision(self):
+        """
+        Retrieves the hardware revision number of the device
```
jleveque, line 60: Suggest rewording:

`Retrieves the hardware revision value of the device`

sonic_platform_base/device_base.py
```diff
@@ -55,6 +55,15 @@ def get_serial(self):
...
+
+        Returns:
+            string: Revision number of device
```
jleveque, line 63: Suggest rewording:

`string: Hardware revision value of device`

jleveque (changes requested):

jleveque (approved):

PR #186, opened 2021-05-05
[sonic_y_cable] add abstract class YCableBase required for Y-cable API support for multiple vendors

vdahiya12:
#### Description
For multiple Y-Cable vendor support once we do have a mapping from vendor/part number to the appropriate Y-Cable module to load, we need to map appropriate port to a module as well. This PR adds definition for base abstract class.

#### Motivation and Context
- Basically, the key idea is once we have a port identified as being of a certain vendor and it has been identified to load a certain module, it is then needed to call the correct module/API on each port each time we call the API on the port

- it is required to maintain this mapping in memory since xcvrd does not want to read/parse this y_cable_vendor_mapping.py file again and again each time we call the Y-Cable API

- Also note that the module loaded might change during xcvrd running lifetime since cable/transceiver can be changed from one vendor to another. So we need to take this into consideration as well

Proposed Solution for this
Each module of the Y-Cable vendor can be a class (of each transceiver type) and all we need to do is instantiate the objects of these classes as class instances and these objects will provide the interface of calling the API's for the appropriate vendor Y-Cable.

This instantiation will be done inside xcvrd, when xcvrd starts

- These objects basically emulate Y-Cable instances and whatever action/interaction needs to be done with the Y-Cable the methods of these objects would provide that

- each vendor in their implementation can inherit from a base class where there will be definitions for all the supported capabilities of the Y-Cable.

- for vendors the recommended approach in case their subclass implementation does not support a method, is to set the method equal to None. This differentiates it from a method they forgot to implement. Then, the calling code should first check if the method is None before attempting to call it.

Design document for the support is [here](https://github.com/Azure/SONiC/pull/757)

Not tested, since this PR just defines the abstract base class implementation.

### DIFF
--- a/sonic_y_cable/y_cable_base.py
+++ b/sonic_y_cable/y_cable_base.py
@@ -1,0 +1,577 @@
+"""
+    y_cable_base.py
+
+    Base class for implementing Y-Cable specific functionality in SONiC.
+    This is the base class of sonic_y_cable implentation. A vendor specific 
+    implementation of YCable needs to inherit from this class.
+    Classes derived from this class provide the ability to interact
+    with a vendor specific Y-Cable
+"""
+
+
+class YCableBase(object):
+
+    # definitions of targets for getting the cursor
+    # equalization parameters from the register spec
+    # the name of the target denotes which side cursor
+    # values will be retreived
+
+    TARGET_NIC = 0
+    TARGET_TOR1 = 1
+    TARGET_TOR2 = 2
+
+    # definitions of targets for getting the EYE/BER
+    # and initiating PRBS/Loopback on the Y cable
+    # the name of the target denotes which side values
+    # will be retreived/initiated
+
+    EYE_PRBS_TARGET_LOCAL = 0
+    EYE_PRBS_TARGET_TOR1 = 1
+    EYE_PRBS_TARGET_TOR2 = 2
+    EYE_PRBS_TARGET_NIC = 3
+
+    # definitions of switch counter types
+    # to be entered by the user in get_switch_count api
+    # for retreiving the counter values
+
+    SWITCH_COUNT_MANUAL = "manual"
+    SWITCH_COUNT_AUTO = "auto"
+
+    # switching modes inside muxcable
+    SWITCHING_MODE_MANUAL = 0
+    SWITCHING_MODE_AUTO = 1
+
+    # Valid return codes for upgrade firmware routine steps
+    FIRMWARE_DOWNLOAD_SUCCESS = 0
+    FIRMWARE_DOWNLOAD_FAILURE = 1
+    FIRMWARE_ACTIVATE_SUCCESS = 0
+    FIRMWARE_ACTIVATE_FAILURE = 1
+    FIRMWARE_ROLLBACK_SUCCESS = 0
+    FIRMWARE_ROLLBACK_FAILURE = 1
+
+    def __init__(self, port):
+        self.port = port
+
+    def toggle_mux_to_torA(self, port):
+        """
+        This API specifically does a hard switch toggle of the Y cable's MUX regardless of link state to
+        TOR A. This means if the Y cable is actively routing, the "check_active_linked_tor_side"
+        API will now return Tor A. It also implies that if the link is actively routing on this port, Y cable
+        MUX will start forwarding packets from TOR A to NIC, and drop packets from TOR B to NIC
+        regardless of previous forwarding state.
+
+        Args:
+            port:
+            an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            a Boolean, true if the toggle succeeded and false if it did not succeed.
+        """
+
+        raise NotImplementedError
+
+    def toggle_mux_to_torB(self, port):
+        """
+        This API specifically does a hard switch toggle of the Y cable's MUX regardless of link state to
+        TOR B. This means if the Y cable is actively routing, the "check_active_linked_tor_side"
+        API will now return Tor B. It also implies that if the link is actively routing on this port, Y cable
+        MUX will start forwarding packets from TOR B to NIC, and drop packets from TOR A to NIC
+        regardless of previous forwarding state.
+
+        Args:
+            port:
+                an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            a Boolean, true if the toggle succeeded and false if it did not succeed.
+        """
+        raise NotImplementedError
+
+    def check_read_side(self, port):
+        """
+        This API specifically checks which side of the Y cable the reads are actually getting performed
+        from, either TOR A or TOR B or NIC and returns the value.
+
+        Args:
+            port:
+                an Integer, the actual physical port connected to Y end of a Y cable which can which side reading the MUX from
+
+        Returns:
+            an Integer, 1 if reading the Y cable from TOR A side(TOR 1).
+                      , 2 if reading the Y cable from TOR B side(TOR 2).
+                      , 0 if reading the Y cable from NIC side.
+                      , -1 if reading the Y cable API fails.
+        """
+
+        raise NotImplementedError
+
+    def check_mux_direction(self, port):
+        """
+        This API specifically checks which side of the Y cable mux is currently point to
+        and returns either TOR A or TOR B. Note that this API should return mux-direction
+        regardless of whether the link is active/routing state or not.
+
+        Args:
+            port:
+                an Integer, the actual physical port connected to a Y cable
+
+        Returns:
+            an Integer, 1 if the mux is pointing to TOR A .
+                      , 2 if the mux is pointing to TOR B.
+                      , -1 if checking which side mux is pointing to API fails.
+        """
+
+        raise NotImplementedError
+
+    def check_active_linked_tor_side(self, port):
+        """
+        This API specifically checks which side of the Y cable is actively linked and routing
+        and returns either TOR A or TOR B.
+
+        Args:
+            port:
+                an Integer, the actual physical port connected to a Y cable
+
+        Returns:
+            an Integer, 1 if TOR A is actively linked and routing(TOR 1).
+                      , 2 if TOR B is actively linked and routing(TOR 2).
+                      , 0 if nothing linked and actively routing
+                      , -1 if checking which side linked for routing API fails.
+        """
+
+        raise NotImplementedError
+
+    def check_if_link_is_active_for_NIC(self, port):
+        """
+        This API specifically checks if NIC side of the Y cable's link is active
+
+        Args:
+            port:
+                 an Integer, the actual physical port connected to a Y cable
+
+        Returns:
+            a boolean, true if the link is active
+                     , false if the link is not active
+        """
+
+        raise NotImplementedError
+
+    def check_if_link_is_active_for_torA(self, port):
+        """
+        This API specifically checks if tor A side of the Y cable's link is active
+
+        Args:
+            port:
+                 an Integer, the actual physical port connected to a Y cable
+
+        Returns:
+            a boolean, true if the link is active
+                     , false if the link is not active
+        """
+
+        raise NotImplementedError
+
+    def check_if_link_is_active_for_torB(self, port):
+        """
+        This API specifically checks if tor B side of the Y cable's link is active
+
+        Args:
+            port:
+                 an Integer, the actual physical port connected to a Y cable
+
+        Returns:
+            a boolean, true if the link is active
+                     , false if the link is not active
+        """
+
+        raise NotImplementedError
+
+    def get_eye_info(self, port):
+        """
+        This API specifically returns the EYE height value for a specfic port.
+        The target could be local side, TOR1, TOR2, NIC etc.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
+                             EYE_PRBS_TARGET_LOCAL -> local side,
+                             EYE_PRBS_TARGET_TOR1 -> TOR 1
+                             EYE_PRBS_TARGET_TOR2 -> TOR 2
+                             EYE_PRBS_TARGET_NIC -> NIC
+        Returns:
+            a list, with EYE values of lane 0 lane 1 lane 2 lane 3 with corresponding index
+        """
+
+        raise NotImplementedError
+
+    def get_ber_info(self, port):
+        """
+        This API specifically returns the BER (Bit error rate) value for a specfic port.
+        The target could be local side, TOR1, TOR2, NIC etc.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
+                             EYE_PRBS_TARGET_LOCAL -> local side,
+                             EYE_PRBS_TARGET_TOR1 -> TOR 1
+                             EYE_PRBS_TARGET_TOR2 -> TOR 2
+                             EYE_PRBS_TARGET_NIC -> NIC
+        Returns:
+            a list, with BER values of lane 0 lane 1 lane 2 lane 3 with corresponding index
+        """
+
+        raise NotImplementedError
+
+    def get_vendor(self, port):
+        """
+        This API specifically returns the vendor name of the Y cable for a specfic port.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+        Returns:
+            a string, with vendor name
+        """
+
+        raise NotImplementedError
+
+    def get_part_number(self, port):
+        """
+        This API specifically returns the part number of the Y cable for a specfic port.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+        Returns:
+            a string, with part number
+        """
+
+        raise NotImplementedError
+
+    def get_switch_count(self, port, count_type):
+        """
+        This API specifically returns the switch count to change the Active TOR which has
+        been done manually by the user.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            count_type:
+                 a string, for getting the count type
+                          "manual" -> manual switch count
+                          "auto" -> automatic switch count
+        Returns:
+            an integer, the number of times manually the Y-cable has been switched
+        """
+
+        raise NotImplementedError
+
+    def get_target_cursor_values(self, port, lane, target):
+        """
+        This API specifically returns the cursor equalization parameters for a target(NIC, TOR1, TOR2).
+        This includes pre one, pre two , main, post one, post two cursor values
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            lane:
+                 an Integer, the lane on which to collect the cursor values
+                             1 -> lane 1,
+                             2 -> lane 2
+                             3 -> lane 3
+                             4 -> lane 4
+            target:
+                 an Integer, the actual target to get the cursor values on
+                             TARGET_NIC -> NIC,
+                             TARGET_TOR1-> TOR1,
+                             TARGET_TOR2 -> TOR2
+        Returns:
+            a list, with  pre one, pre two , main, post one, post two cursor values in the order
+        """
+
+        raise NotImplementedError
+
+    def get_firmware_version(self, port, target):
+        """ This routine should return the active, inactive and next (committed)
+        firmware running on the target. Each of the version values in this context
+        could be a string with a major and minor number and a build value.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the actual target to get the cursor values on
+                             TARGET_NIC -> NIC,
+                             TARGET_TOR1-> TOR1,
+                             TARGET_TOR2 -> TOR2
+        Returns:
+            a Dictionary:
+                 with version_active, version_inactive and version_next keys
+                 and their corresponding values
+
+        """
+
+        raise NotImplementedError
+
+    def download_firmware(self, port, fwfile):
+        """ This routine should download and store the firmware on all the
+        components of the Y cable of the port specified.
+        This should include any internal transfers, checksum validation etc.
+        from TOR to TOR or TOR to NIC side of the firmware specified by the fwfile.
+        This basically means that the firmware which is being downloaded should be
+        available to be activated (start being utilized by the cable) once this API is
+        successfully executed.
+        Note that this API should ideally not require any rollback even if it fails
+        as this should not interfere with the existing cable functionality because
+        this has not been activated yet.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            fwfile:
+                 a string, a path to the file which contains the firmware image.
+                 Note that the firmware file can be in the format of the vendor's
+                 choosing (binary, archive, etc.). But note that it should be one file
+                 which contains firmware for all components of the Y-cable
+        Returns:
+            an Integer:
+                 a predefined code stating whether the firmware download was successful
+                 or an error code as to what was the cause of firmware download failure
+        """
+
+        raise NotImplementedError
+
+    def activate_firmware(self, port, fwfile):
+        """ This routine should activate the downloaded firmware on all the
+        components of the Y cable of the port specified.
+        This API is meant to be used in conjunction with download_firmware API, and
+        should be called once download_firmware API is succesful.
+        This means that the firmware which has been downloaded should be
+        activated (start being utilized by the cable) once this API is
+        successfully executed.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+        Returns:
+            an Integer:
+                 a predefined code stating whether the firmware activate was successful
+        """
+
+        raise NotImplementedError
+
+    def rollback_firmware(self, port):
+        """ This routine should rollback the firmware to the previous version
+        which was being used by the cable. This API is intended to be called when the
+        user either witnesses an activate_firmware API failure or sees issues with
+        newer firmware in regards to stable cable functioning.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+        Returns:
+            an Integer:
+                 a predefined code stating whether the firmware rollback was successful
+                 or an error code as to what was the cause of firmware rollback failure
+        """
+
+        raise NotImplementedError
+
+    def set_switching_mode(self, port, mode):
+        """
+        This API specifically enables the auto switching or manual switching feature on the muxcable,
+        depending upon the mode entered by the user.
+        Autoswitch feature if enabled actually does an automatic toggle of the mux in case the active
+        side link goes down and basically points the mux to the other side.
+
+        Args:
+             physical_port:
+                 an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+             mode:
+                 an Integer, specifies which type of switching mode we set the muxcable to
+                 either SWITCHING_MODE_AUTO or SWITCHING_MODE_MANUAL
+
+        Returns:
+            a Boolean, true if the switch succeeded and false if it did not succeed.
+        """
+
+        raise NotImplementedError
+
+    def get_switching_mode(self, port):
+        """
+        This API specifically returns which type of switching mode the cable is set to auto/manual
+
+        Args:
+             physical_port:
+                 an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            an Integer, SWITCHING_MODE_AUTO if auto switch is enabled.
+                        SWITCHING_MODE_MANUAL if manual switch is enabled.
+        """
+
+        raise NotImplementedError
+
+    def get_nic_temperature(self, port):
+        """
+        This API specifically returns nic temperature of the physical port specified
+
+        Args:
+             physical_port:
+                 an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            an Integer, the temperature of the NIC MCU
+        """
+
+        raise NotImplementedError
+
+    def get_local_temperature(self, port):
+        """
+        This API specifically returns local ToR temperature of the physical port specified
+
+        Args:
+             physical_port:
+                 an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            an Integer, the temperature of the local MCU
+        """
+
+        raise NotImplementedError
+
+    def get_nic_voltage(self, port):
+        """
+        This API specifically returns nic voltage of the physical port specified
+
+        Args:
+             physical_port:
+                 an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            a float, the voltage of the NIC MCU
+        """
+
+        raise NotImplementedError
+
+    def get_local_voltage(self, port):
+        """
+        This API specifically returns local ToR voltage of the physical port specified
+
+        Args:
+             physical_port:
+                 an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            a float, the voltage of the local MCU
+        """
+
+        raise NotImplementedError
+
+    def enable_prbs_mode(self, port, target, mode_value, lane_map):
+        """
+        This API specifically configures and enables the PRBS mode/type depending upon the mode_value the user provides.
+        The mode_value configures the PRBS Type for generation and BER sensing on a per side basis.
+        Target is an integer for selecting which end of the Y cable we want to run PRBS on.
+        LaneMap specifies the lane configuration to run the PRBS on.
+        Note that this is a diagnostic mode command and must not run during normal traffic/switch operation
+
+        Args:
+            physical_port:
+                an Integer, the actual physical port connected to a Y cable
+            target:
+                an Integer, the target on which to enable the PRBS
+                             EYE_PRBS_TARGET_LOCAL -> local side,
+                             EYE_PRBS_TARGET_TOR1 -> TOR 1
+                             EYE_PRBS_TARGET_TOR2 -> TOR 2
+                             EYE_PRBS_TARGET_NIC -> NIC
+            mode_value:
+                 an Integer, the mode/type for configuring the PRBS mode.
+
+            lane_map:
+                 an Integer, representing the lane_map to be run PRBS on
+                 0bit for lane 0, 1bit for lane1 and so on.
+                 for example 3 -> 0b'0011 , means running on lane0 and lane1
+
+        Returns:
+            a boolean, true if the enable is successful
+                     , false if the enable failed
+
+        """
+
+        raise NotImplementedError
+
+    def disable_prbs_mode(self, port, target):
+        """
+        This API specifically disables the PRBS mode on the physical port.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
+                             EYE_PRBS_TARGET_LOCAL -> local side,
+                             EYE_PRBS_TARGET_TOR1 -> TOR 1
+                             EYE_PRBS_TARGET_TOR2 -> TOR 2
+                             EYE_PRBS_TARGET_NIC -> NIC
+
+        Returns:
+            a boolean, true if the disable is successful
+                     , false if the disable failed
+        """
+
+        raise NotImplementedError
+
+    def enable_loopback_mode(self, port, target, lane_map):
+        """
+        This API specifically configures and enables the Loopback mode on the port user provides.
+        Target is an integer for selecting which end of the Y cable we want to run loopback on.
+        LaneMap specifies the lane configuration to run the loopback on.
+        Note that this is a diagnostic mode command and must not run during normal traffic/switch operation
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
+                             EYE_PRBS_TARGET_LOCAL -> local side,
+                             EYE_PRBS_TARGET_TOR1 -> TOR 1
+                             EYE_PRBS_TARGET_TOR2 -> TOR 2
+                             EYE_PRBS_TARGET_NIC -> NIC
+            lane_map:
+                 an Integer, representing the lane_map to be run PRBS on
+                 0bit for lane 0, 1bit for lane1 and so on.
+                 for example 3 -> 0b'0011 , means running on lane0 and lane1
+
+        Returns:
+            a boolean, true if the enable is successful
+                     , false if the enable failed
+        """
+
+        raise NotImplementedError
+
+    def disable_loopback_mode(self, port, target):
+        """
+        This API specifically disables the Loopback mode on the port user provides.
+        Target is an integer for selecting which end of the Y cable we want to run loopback on.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
+                             EYE_PRBS_TARGET_LOCAL -> local side,
+                             EYE_PRBS_TARGET_TOR1 -> TOR 1
+                             EYE_PRBS_TARGET_TOR2 -> TOR 2
+                             EYE_PRBS_TARGET_NIC -> NIC
+
+        Returns:
+            a boolean, true if the disable is successful
+                     , false if the disable failed
+        """
+
+        raise NotImplementedError
--- a/sonic_y_cable/y_cable_base.py
+++ b/sonic_y_cable/y_cable_base.py
@@ -54,8 +54,9 @@ class YCableBase():
 
     # definitions of the modes to be run for loopback mode
     # on the port/cable
-    LOOPBACK_MODE_NEAR_END = 0
-    LOOPBACK_MODE_FAR_END = 1
+    LOOPBACK_MODE_NONE = 0
+    LOOPBACK_MODE_NEAR_END = 1
+    LOOPBACK_MODE_FAR_END = 2
 
     # Valid return codes for upgrade firmware routine steps
     FIRMWARE_DOWNLOAD_SUCCESS = 0
@@ -1104,6 +1105,7 @@ class YCableBase():
         Returns:
             mode_value:
                 One of the following predefined constants, the mode to be run for loopback:
+                    LOOPBACK_MODE_NONE
                     LOOPBACK_MODE_NEAR_END
                     LOOPBACK_MODE_FAR_END
         """
--- a/sonic_y_cable/y_cable_base.py
+++ b/sonic_y_cable/y_cable_base.py
@@ -43,6 +43,17 @@ class YCableBase(object):
     SWITCHING_MODE_MANUAL = 0
     SWITCHING_MODE_AUTO = 1
 
+    # definitions of the fec mode to be configured
+    # on a port/cable
+    FEC_MODE_NONE,
+    FEC_MODE_RS,
+    FEC_MODE_FC
+
+    # definitions of the modes to be run for loopback mode
+    # on the port/cable
+    LOOPBACK_MODE_NEAR_END
+    LOOPBACK_MODE_FAR_END
+
     # Valid return codes for upgrade firmware routine steps
     FIRMWARE_DOWNLOAD_SUCCESS = 0
     FIRMWARE_DOWNLOAD_FAILURE = 1
@@ -61,7 +72,7 @@ class YCableBase(object):
 
     def toggle_mux_to_tor_a(self):
         """
-        This API specifically does a hard switch toggle of the Y cable's MUX regardless of link state to
+        This API does a hard switch toggle of the Y cable's MUX regardless of link state to
         TOR A on the port this is called for. This means if the Y cable is actively sending traffic,
         the "get_active_linked_tor_side" API will now return Tor A.
         It also implies that if the link is actively sending traffic on this port,
@@ -79,7 +90,7 @@ class YCableBase(object):
 
     def toggle_mux_to_tor_b(self):
         """
-        This API specifically does a hard switch toggle of the Y cable's MUX regardless of link state to
+        This API does a hard switch toggle of the Y cable's MUX regardless of link state to
         TOR B. This means if the Y cable is actively sending traffic, the "get_active_linked_tor_side"
         API will now return Tor B. It also implies that if the link is actively sending traffic on this port,
         Y cable. MUX will start forwarding packets from TOR B to NIC, and drop packets from TOR A to NIC
@@ -95,7 +106,7 @@ class YCableBase(object):
 
     def get_read_side(self):
         """
-        This API specifically checks which side of the Y cable the reads are actually getting performed
+        This API checks which side of the Y cable the reads are actually getting performed
         from, either TOR A or TOR B or NIC and returns the value.
         The port on which this API is called for can be referred using self.port.
 
@@ -113,7 +124,7 @@ class YCableBase(object):
 
     def get_mux_direction(self):
         """
-        This API specifically checks which side of the Y cable mux is currently point to
+        This API checks which side of the Y cable mux is currently point to
         and returns either TOR A or TOR B. Note that this API should return mux-direction
         regardless of whether the link is active and sending traffic or not.
         The port on which this API is called for can be referred using self.port.
@@ -131,7 +142,7 @@ class YCableBase(object):
 
     def get_active_linked_tor_side(self):
         """
-        This API specifically checks which side of the Y cable is actively linked and sending traffic
+        This API checks which side of the Y cable is actively linked and sending traffic
         and returns either TOR A or TOR B.
         The port on which this API is called for can be referred using self.port.
         This is different from get_mux_direction in a sense it also implies the link on the side
@@ -151,7 +162,7 @@ class YCableBase(object):
 
     def is_link_active(self, target):
         """
-        This API specifically checks if NIC, TOR_A and TOR_B  of the Y cable's link is active.
+        This API checks if NIC, TOR_A and TOR_B  of the Y cable's link is active.
         The target specifies which link is supposed to be checked
         The port on which this API is called for can be referred using self.port.
 
@@ -171,7 +182,7 @@ class YCableBase(object):
 
     def get_eye_info(self, target):
         """
-        This API specifically returns the EYE height value for a specfic port.
+        This API returns the EYE height value for a specfic port.
         The target could be local side, TOR_A, TOR_B, NIC etc.
         The port on which this API is called for can be referred using self.port.
 
@@ -190,7 +201,7 @@ class YCableBase(object):
 
     def get_ber_info(self, target):
         """
-        This API specifically returns the BER (Bit error rate) value for a specfic port.
+        This API returns the BER (Bit error rate) value for a specfic port.
         The target could be local side, TOR_A, TOR_B, NIC etc.
         The port on which this API is called for can be referred using self.port.
 
@@ -209,7 +220,7 @@ class YCableBase(object):
 
     def get_vendor(self):
         """
-        This API specifically returns the vendor name of the Y cable for a specfic port.
+        This API returns the vendor name of the Y cable for a specfic port.
         The port on which this API is called for can be referred using self.port.
 
         Args:
@@ -222,7 +233,7 @@ class YCableBase(object):
 
     def get_part_number(self):
         """
-        This API specifically returns the part number of the Y cable for a specfic port.
+        This API returns the part number of the Y cable for a specfic port.
         The port on which this API is called for can be referred using self.port.
 
         Args:
@@ -233,17 +244,80 @@ class YCableBase(object):
 
         raise NotImplementedError
 
-    def get_switch_count(self, count_type):
+    def get_switch_count_total(self, switch_count_type, clear_on_read=False):
         """
-        This API specifically returns the switch count to change the Active TOR which has
-        been done manually by the user.
+        This API returns the total switch count to change the Active TOR which has
+        been done manually/automatic by the user.
         The port on which this API is called for can be referred using self.port.
 
         Args:
-            count_type:
+            switch_count_type:
                 One of the following predefined constants, for getting the count type:
                     SWITCH_COUNT_MANUAL -> manual switch count
                     SWITCH_COUNT_AUTO -> automatic switch count
+            clear_on_read:
+                a boolean, True if the count has to be reset after read to zero
+                         , False if the count is not to be reset after read
+            Returns:
+                an integer, the number of times the Y-cable has been switched
+        """
+
+        raise NotImplementedError
+
+    def get_switch_count_tor_a(self, clear_on_read=False):
+        """
+        This API returns the switch count to change the Active TOR which has
+        been done manually by the user initiated from ToR A
+        The port on which this API is called for can be referred using self.port.
+
+        Args:
+            clear_on_read:
+                a boolean, True if the count has to be reset after read to zero
+                         , False if the count is not to be reset after read
+
+            Returns:
+                an integer, the number of times the Y-cable has been switched from ToR A
+        """
+
+        raise NotImplementedError
+
+    def get_switch_count_tor_b(self, clear_on_read=False):
+        """
+        This API returns the switch count to change the Active TOR which has
+        been done manually by the user initiated from ToR B
+        The port on which this API is called for can be referred using self.port.
+
+        Args:
+            clear_on_read:
+                a boolean, True if the count has to be reset after read to zero
+                         , False if the count is not to be reset after read
+
+            Returns:
+                an integer, the number of times the Y-cable has been switched from ToR B
+        """
+
+        raise NotImplementedError
+
+    def get_switch_count_target(self, switch_count_type, target, clear_on_read=False):
+        """
+        This API returns the total number of times the Active TOR has
+        been done manually/automaticlly toggled towards a target.
+        For example, TARGET_TOR_A as target would imply
+        how many times the mux has been toggled towards TOR A.
+        The port on which this API is called for can be referred using self.port.
+
+        Args:
+            switch_count_type:
+                One of the following predefined constants, for getting the count type:
+                    SWITCH_COUNT_MANUAL -> manual switch count
+                    SWITCH_COUNT_AUTO -> automatic switch count
+            target:
+                One of the following predefined constants, the actual target to check the link on:
+                     TARGET_TOR_A-> TORA,
+                     TARGET_TOR_B -> TORB
+            clear_on_read:
+                a boolean, True if the count has to be reset after read to zero
+                         , False if the count is not to be reset after read
             Returns:
                 an integer, the number of times manually the Y-cable has been switched
         """
@@ -252,7 +326,7 @@ class YCableBase(object):
 
     def get_target_cursor_values(self, lane, target):
         """
-        This API specifically returns the cursor equalization parameters for a target(NIC, TOR_A, TOR_B).
+        This API returns the cursor equalization parameters for a target(NIC, TOR_A, TOR_B).
         This includes pre one, pre two , main, post one, post two cursor values
         The port on which this API is called for can be referred using self.port.
 
@@ -270,6 +344,33 @@ class YCableBase(object):
                      TARGET_TOR_B -> TORB
         Returns:
             a list, with  pre one, pre two , main, post one, post two cursor values in the order
+        """
+
+        raise NotImplementedError
+
+    def set_target_cursor_values(self, lane, cursor_values, target):
+        """
+        This API sets the cursor equalization parameters for a target(NIC, TOR_A, TOR_B).
+        This includes pre one, pre two , main, post one, post two etc. cursor values
+        The port on which this API is called for can be referred using self.port.
+
+        Args:
+            lane:
+                 an Integer, the lane on which to collect the cursor values
+                             1 -> lane 1,
+                             2 -> lane 2
+                             3 -> lane 3
+                             4 -> lane 4
+            cursor_values:
+                a list, with  pre one, pre two , main, post one, post two cursor etc. values in the order
+            target:
+                One of the following predefined constants, the actual target to get the cursor values on:
+                     TARGET_NIC -> NIC,
+                     TARGET_TOR_A-> TORA,
+                     TARGET_TOR_B -> TORB
+        Returns:
+            a boolean, True if cursor values setting is successful
+                     , False if cursor values setting is not successful
         """
 
         raise NotImplementedError
@@ -379,7 +480,7 @@ class YCableBase(object):
 
     def set_switching_mode(self, mode):
         """
-        This API specifically enables the auto switching or manual switching feature on the Y-Cable,
+        This API enables the auto switching or manual switching feature on the Y-Cable,
         depending upon the mode entered by the user.
         Autoswitch feature if enabled actually does an automatic toggle of the mux in case the active
         side link goes down and basically points the mux to the other side.
@@ -402,7 +503,7 @@ class YCableBase(object):
 
     def get_switching_mode(self):
         """
-        This API specifically returns which type of switching mode the cable is set to auto/manual
+        This API returns which type of switching mode the cable is set to auto/manual
         The port on which this API is called for can be referred using self.port.
 
         Args:
@@ -417,7 +518,7 @@ class YCableBase(object):
 
     def get_nic_temperature(self):
         """
-        This API specifically returns nic temperature of the physical port for which this API is called.
+        This API returns nic temperature of the physical port for which this API is called.
         The port on which this API is called for can be referred using self.port.
 
         Args:
@@ -430,7 +531,7 @@ class YCableBase(object):
 
     def get_local_temperature(self):
         """
-        This API specifically returns local ToR temperature of the physical port for which this API is called.
+        This API returns local ToR temperature of the physical port for which this API is called.
         The port on which this API is called for can be referred using self.port.
 
         Args:
@@ -443,7 +544,7 @@ class YCableBase(object):
 
     def get_nic_voltage(self):
         """
-        This API specifically returns nic voltage of the physical port for which this API is called.
+        This API returns nic voltage of the physical port for which this API is called.
         The port on which this API is called for can be referred using self.port.
 
         Args:
@@ -456,7 +557,7 @@ class YCableBase(object):
 
     def get_local_voltage(self):
         """
-        This API specifically returns local ToR voltage of the physical port for which this API is called.
+        This API returns local ToR voltage of the physical port for which this API is called.
         The port on which this API is called for can be referred using self.port.
 
         Args:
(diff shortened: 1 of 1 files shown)


sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+    def get_eye_info(self, port):
+        """
+        This API specifically returns the EYE height value for a specfic port.
+        The target could be local side, TOR1, TOR2, NIC etc.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
```
jleveque, line 198: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

I would prefer using an Enum class here. However, there was a big change between Python 2 and 3, and sonic-platform-common still needs to support Python 2 for a little while longer, so I think we avoid Enums for now, and maybe change over after the package is Python 3-only.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+    def get_ber_info(self, port):
+        """
+        This API specifically returns the BER (Bit error rate) value for a specfic port.
+        The target could be local side, TOR1, TOR2, NIC etc.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
```
jleveque, line 218: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            lane:
+                 an Integer, the lane on which to collect the cursor values
+                             1 -> lane 1,
+                             2 -> lane 2
+                             3 -> lane 3
+                             4 -> lane 4
+            target:
+                 an Integer, the actual target to get the cursor values on
```
jleveque, line 288: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+    def get_firmware_version(self, port, target):
+        """ This routine should return the active, inactive and next (committed)
+        firmware running on the target. Each of the version values in this context
+        could be a string with a major and minor number and a build value.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the actual target to get the cursor values on
```
jleveque, line 307: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            fwfile:
+                 a string, a path to the file which contains the firmware image.
+                 Note that the firmware file can be in the format of the vendor's
+                 choosing (binary, archive, etc.). But note that it should be one file
+                 which contains firmware for all components of the Y-cable
+        Returns:
+            an Integer:
```
jleveque, line 341: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

Also, please explicitly list the acceptable constants this function should return.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        should be called once download_firmware API is succesful.
+        This means that the firmware which has been downloaded should be
+        activated (start being utilized by the cable) once this API is
+        successfully executed.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+        Returns:
+            an Integer:
```
jleveque, line 361: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

Also, please explicitly list the acceptable constants this function should return.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        """ This routine should rollback the firmware to the previous version
+        which was being used by the cable. This API is intended to be called when the
+        user either witnesses an activate_firmware API failure or sees issues with
+        newer firmware in regards to stable cable functioning.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+        Returns:
+            an Integer:
```
jleveque, line 377: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

Also, please explicitly list the acceptable constants this function should return.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        """
+        This API specifically returns which type of switching mode the cable is set to auto/manual
+
+        Args:
+             physical_port:
+                 an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            an Integer, SWITCHING_MODE_AUTO if auto switch is enabled.
+                        SWITCHING_MODE_MANUAL if manual switch is enabled.
```
jleveque, line 414: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        The mode_value configures the PRBS Type for generation and BER sensing on a per side basis.
+        Target is an integer for selecting which end of the Y cable we want to run PRBS on.
+        LaneMap specifies the lane configuration to run the PRBS on.
+        Note that this is a diagnostic mode command and must not run during normal traffic/switch operation
+
+        Args:
+            physical_port:
+                an Integer, the actual physical port connected to a Y cable
+            target:
+                an Integer, the target on which to enable the PRBS
```
jleveque, line 487: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+
+    def disable_prbs_mode(self, port, target):
+        """
+        This API specifically disables the PRBS mode on the physical port.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
```
jleveque, line 516: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        This API specifically configures and enables the Loopback mode on the port user provides.
+        Target is an integer for selecting which end of the Y cable we want to run loopback on.
+        LaneMap specifies the lane configuration to run the loopback on.
+        Note that this is a diagnostic mode command and must not run during normal traffic/switch operation
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
```
jleveque, line 540: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+    def disable_loopback_mode(self, port, target):
+        """
+        This API specifically disables the Loopback mode on the port user provides.
+        Target is an integer for selecting which end of the Y cable we want to run loopback on.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 an Integer, the target on which to enable the PRBS
```
jleveque, line 566: Maybe rather than explicitly stating "An Integer", we simply state "One of the following predefined constants:". This would also allow us to change the underlying type transparently if desired.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        API will now return Tor A. It also implies that if the link is actively routing on this port, Y cable
+        MUX will start forwarding packets from TOR A to NIC, and drop packets from TOR B to NIC
+        regardless of previous forwarding state.
+
+        Args:
+            port:
+            an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            a Boolean, true if the toggle succeeded and false if it did not succeed.
```
jleveque, line 68: Suggest capitalizing `True` and `False` to align with Python syntax. Please do this throughout the file.

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+class YCableBase(object):
+
+    # definitions of targets for getting the cursor
+    # equalization parameters from the register spec
+    # the name of the target denotes which side cursor
+    # values will be retreived
+
+    TARGET_NIC = 0
+    TARGET_TOR1 = 1
+    TARGET_TOR2 = 2
```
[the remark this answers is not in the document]
theasianpianist: The 'A/B' convention is already pretty widely adopted in sonic-mgmt and the existing Y cable codebase, I'd suggest sticking with that.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+    FIRMWARE_ROLLBACK_FAILURE = 1
+
+    def __init__(self, port):
+        self.port = port
+
+    def toggle_mux_to_torA(self, port):
+        """
+        This API specifically does a hard switch toggle of the Y cable's MUX regardless of link state to
+        TOR A. This means if the Y cable is actively routing, the "check_active_linked_tor_side"
+        API will now return Tor A. It also implies that if the link is actively routing on this port, Y cable
```
theasianpianist, line 59: nit: maybe change 'routing' to 'sending', since 'routing' tends to imply some decision-making capability which is not the case for the cable.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+            a Boolean, true if the toggle succeeded and false if it did not succeed.
+        """
+
+        raise NotImplementedError
+
+    def toggle_mux_to_torB(self, port):
+        """
+        This API specifically does a hard switch toggle of the Y cable's MUX regardless of link state to
+        TOR B. This means if the Y cable is actively routing, the "check_active_linked_tor_side"
+        API will now return Tor B. It also implies that if the link is actively routing on this port, Y cable
```
theasianpianist, line 77: Same as above

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+
+        Returns:
+            an Integer, 1 if the mux is pointing to TOR A .
+                      , 2 if the mux is pointing to TOR B.
+                      , -1 if checking which side mux is pointing to API fails.
+        """
+
+        raise NotImplementedError
+
+    def check_active_linked_tor_side(self, port):
```
theasianpianist, line 126: How is this different from `check_mux_direction`?

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+                      , 2 if TOR B is actively linked and routing(TOR 2).
+                      , 0 if nothing linked and actively routing
+                      , -1 if checking which side linked for routing API fails.
+        """
+
+        raise NotImplementedError
+
+    def check_if_link_is_active_for_NIC(self, port):
+        """
+        This API specifically checks if NIC side of the Y cable's link is active
```
theasianpianist, line 146: What is the definition we are using for 'active'? Does that mean the cable is plugged in, traffic is flowing through this link, etc.

sonic_y_cable/y_cable_base.py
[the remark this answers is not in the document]
vdahiya12: Fixed, throughout the file

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        This API specifically checks which side of the Y cable mux is currently point to
+        and returns either TOR A or TOR B. Note that this API should return mux-direction
+        regardless of whether the link is active/routing state or not.
+
+        Args:
+            port:
+                an Integer, the actual physical port connected to a Y cable
+
+        Returns:
+            an Integer, 1 if the mux is pointing to TOR A .
```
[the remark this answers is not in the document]
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+
+        Args:
+            port:
+                an Integer, the actual physical port connected to Y end of a Y cable which can which side reading the MUX from
+
+        Returns:
+            an Integer, 1 if reading the Y cable from TOR A side(TOR 1).
+                      , 2 if reading the Y cable from TOR B side(TOR 2).
+                      , 0 if reading the Y cable from NIC side.
+                      , -1 if reading the Y cable API fails.
```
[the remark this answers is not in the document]
vdahiya12: fixed, aligned as suggested

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        regardless of whether the link is active/routing state or not.
+
+        Args:
+            port:
+                an Integer, the actual physical port connected to a Y cable
+
+        Returns:
+            an Integer, 1 if the mux is pointing to TOR A .
+                      , 2 if the mux is pointing to TOR B.
+                      , -1 if checking which side mux is pointing to API fails.
```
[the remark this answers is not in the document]
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,577 @@
...
+        This API specifically returns the switch count to change the Active TOR which has
+        been done manually by the user.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            count_type:
+                 a string, for getting the count type
+                          "manual" -> manual switch count
+                          "auto" -> automatic switch count
```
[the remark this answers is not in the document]
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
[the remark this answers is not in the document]
jleveque: There are still a couple on lines 293-294. Please make another pass.

sonic_y_cable/y_cable_base.py
```diff
@@ -97,10 +98,11 @@ def check_read_side(self, port):
                 an Integer, the actual physical port connected to Y end of a Y cable which can which side reading the MUX from
 
         Returns:
-            an Integer, 1 if reading the Y cable from TOR A side(TOR 1).
-                      , 2 if reading the Y cable from TOR B side(TOR 2).
-                      , 0 if reading the Y cable from NIC side.
-                      , -1 if reading the Y cable API fails.
+            One of the following predefined constants:
+                TARGET_TORA, if reading the Y cable from TOR A side(TOR A).
+                TARGET_TORB, if reading the Y cable from TOR B side(TOR B).
```
jleveque, line 103: Seems like the parenthesized text is redundant now. Suggest removing.

sonic_y_cable/y_cable_base.py
```diff
@@ -116,9 +118,10 @@ def check_mux_direction(self, port):
                 an Integer, the actual physical port connected to a Y cable
 
         Returns:
-            an Integer, 1 if the mux is pointing to TOR A .
-                      , 2 if the mux is pointing to TOR B.
-                      , -1 if checking which side mux is pointing to API fails.
+            One of the following predefined constants:
+                TARGET_TORA, if mux is pointing to TOR A side(TOR A).
+                TARGET_TORB, if mux is pointing to TOR B side(TOR B).
```
jleveque, line 123: Seems like the parenthesized text is redundant now. Suggest removing.

sonic_y_cable/y_cable_base.py
```diff
@@ -133,10 +136,10 @@ def check_active_linked_tor_side(self, port):
                 an Integer, the actual physical port connected to a Y cable
 
         Returns:
-            an Integer, 1 if TOR A is actively linked and routing(TOR 1).
-                      , 2 if TOR B is actively linked and routing(TOR 2).
-                      , 0 if nothing linked and actively routing
-                      , -1 if checking which side linked for routing API fails.
+            One of the following predefined constants:
+                TARGET_TORA, if TOR A is actively linked and routing(TOR A).
+                TARGET_TORB, if TOR B is actively linked and routing(TOR B).
```
jleveque, line 141: Seems like the parenthesized text is redundant now. Suggest removing.

sonic_y_cable/y_cable_base.py
```diff
@@ -17,17 +17,18 @@ class YCableBase(object):
     # values will be retreived
 
     TARGET_NIC = 0
-    TARGET_TOR1 = 1
-    TARGET_TOR2 = 2
+    TARGET_TORA = 1
+    TARGET_TORB = 2
+    TARGET_UNKNOWN = -1
```
jleveque, line 22: Suggest moving above `TARGET_NIC` to align in numerical order.

sonic_y_cable/y_cable_base.py
```diff
@@ -262,8 +265,8 @@ def get_switch_count(self, port, count_type):
                  an Integer, the actual physical port connected to a Y cable
             count_type:
                  a string, for getting the count type
```
jleveque, line 267: Suggest changing to `One of the following predefined constants:`

sonic_y_cable/y_cable_base.py
```diff
@@ -180,26 +183,26 @@ def check_if_link_is_active_for_torB(self, port):
...
         Args:
             physical_port:
                  an Integer, the actual physical port connected to a Y cable
             target:
-                 an Integer, the target on which to enable the PRBS
-                             EYE_PRBS_TARGET_LOCAL -> local side,
-                             EYE_PRBS_TARGET_TOR1 -> TOR 1
-                             EYE_PRBS_TARGET_TOR2 -> TOR 2
-                             EYE_PRBS_TARGET_NIC -> NIC
+                 One of the following predefined constant, the target on which to get the eye:
```
jleveque, line 201: s/constant/constants/

sonic_y_cable/y_cable_base.py
```diff
@@ -209,17 +212,17 @@ def get_eye_info(self, port):
...
         Args:
             physical_port:
                  an Integer, the actual physical port connected to a Y cable
             target:
-                 an Integer, the target on which to enable the PRBS
-                             EYE_PRBS_TARGET_LOCAL -> local side,
-                             EYE_PRBS_TARGET_TOR1 -> TOR 1
-                             EYE_PRBS_TARGET_TOR2 -> TOR 2
-                             EYE_PRBS_TARGET_NIC -> NIC
+                 One of the following predefined constant, the target on which to get the ber:
```
jleveque, line 221: s/constant/constants/

sonic_y_cable/y_cable_base.py
```diff
@@ -262,8 +265,8 @@ def get_switch_count(self, port, count_type):
                  an Integer, the actual physical port connected to a Y cable
             count_type:
                  a string, for getting the count type
-                          "manual" -> manual switch count
-                          "auto" -> automatic switch count
+                          SWITCH_COUNT_MANUAL -> manual switch count
+                          SWITCH_COUNT_MANUAL -> automatic switch count
```
jleveque, line 269: s/SWITCH_COUNT_MANUAL/SWITCH_COUNT_AUTO/

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
vdahiya12: actually this takes into the account that link is active as well, if link is not active this will reflect as unknown, mux-direction on the other hand does not care about link-status

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: yes basically link is up means cable is up and configured correctly, show int status should show "oper" up

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
jleveque: Doesn't appear to be fixed.

sonic_y_cable/y_cable_base.py
jleveque: Doesn't appear to be fixed.

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
theasianpianist: Looks like it still says 'routing'

sonic_y_cable/y_cable_base.py
theasianpianist: same as above

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,589 @@
...
+        and returns either TOR A or TOR B.
+
+        Args:
+            port:
+                an Integer, the actual physical port connected to a Y cable
+
+        Returns:
+            One of the following predefined constants:
+                TARGET_TORA, if TOR A is actively linked and routing/sending traffic.
+                TARGET_TORB, if TOR B is actively linked and routin/sending traffic.
```
theasianpianist, line 141: s/routin/routing/

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,589 @@
...
+        Returns:
+            One of the following predefined constants:
+                FIRMWARE_ACTIVATE_SUCCESS
+                FIRMWARE_ACTIVATE_FAILURE
+        """
+
+        raise NotImplementedError
+
+    def rollback_firmware(self, port):
+        """ This routine should rollback the firmware to the previous version
```
theasianpianist, line 375: Opening triple quote should be on its own line
```suggestion
        """
        This routine should rollback the firmware to the previous version
```

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,589 @@
...
+
+    # definitions of targets for getting the various fields/cursor
+    # equalization parameters from the register spec
+    # the name of the target denotes which side MCU
+    # values will be retreived on the Y-Cable
+
+    TARGET_UNKNOWN = -1
+    TARGET_NIC = 0
+    TARGET_TORA = 1
+    TARGET_TORB = 2
```
theasianpianist, line 22: May be better for readability if these are `TARGET_TOR_A` and `TARGET_TOR_B` instead

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,589 @@
...
+
+    # definitions of targets for getting the EYE/BER
+    # and initiating PRBS/Loopback on the Y cable
+    # the name of the target denotes which side values
+    # will be retreived/initiated
+
+    EYE_PRBS_TARGET_LOCAL = 0
+    EYE_PRBS_TARGET_TORA = 1
+    EYE_PRBS_TARGET_TORB = 2
```
theasianpianist, line 31: same as above

sonic_y_cable/y_cable_base.py
theasianpianist: Is this definition commonly known? It might be worth including this info in the docstring.

sonic_y_cable/y_cable_base.py
theasianpianist: On second thought, it may be better to have it only as 'sending', as per https://github.com/Azure/sonic-platform-common/pull/186/files#r626912311

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,589 @@
...
+    def get_ber_info(self, port):
+        """
+        This API specifically returns the BER (Bit error rate) value for a specfic port.
+        The target could be local side, TORA, TORB, NIC etc.
+
+        Args:
+            physical_port:
+                 an Integer, the actual physical port connected to a Y cable
+            target:
+                 One of the following predefined constants, the target on which to get the ber:
```
jleveque, line 221: s/ber/BER/

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: Hmm maybe I can elaborate a little. I think "oper" up on Cli show interfaces status might confuse the vendors, but objectively  this API it's just their capability to check if all the parameters for transceiver initialization happened correctly, and the link is active. Actually when the lanes come up they send some packets internally on Serdes to check if lanes are functioning properly and a bunch of other things to check if link is healthy. While this generally translates to link is up, they cant tell if the link will be on SONiC side. So I think we can make this definition more concrete but maybe we should not put "show int status" output a cross check for them. Thoughts ?

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
theasianpianist: I think as long as the vendors know what this means, it's fine as is. I'll defer to your judgement here.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,593 @@
...
+        from TOR to TOR or TOR to NIC side of the firmware specified by the fwfile.
+        This basically means that the firmware which is being downloaded should be
+        available to be activated (start being utilized by the cable) once this API is
+        successfully executed.
+        Note that this API should ideally not require any rollback even if it fails
+        as this should not interfere with the existing cable functionality because
+        this has not been activated yet.
+
+        Args:
+            physical_port:
```
jleveque, line 338: Argument name (`physical_port`) doesn't match that in the signature (`port`). This is true a number of functions in the file. Please go over all and make sure they align.

Also, there is no description of the `fwfile` argument. Please add.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,593 @@
...
+                FIRMWARE_DOWNLOAD_SUCCESS
+                FIRMWARE_DOWNLOAD_FAILURE
+
+                a predefined code stating whether the firmware download was successful
+                or an error code as to what was the cause of firmware download failure
+        """
+
+        raise NotImplementedError
+
+    def activate_firmware(self, port, fwfile):
```
jleveque, line 356: In this function, `fwfile` is optional. Suggest defaulting it to `fwfile=None`

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,602 @@
...
+    FIRMWARE_DOWNLOAD_FAILURE = 1
+    FIRMWARE_ACTIVATE_SUCCESS = 0
+    FIRMWARE_ACTIVATE_FAILURE = 1
+    FIRMWARE_ROLLBACK_SUCCESS = 0
+    FIRMWARE_ROLLBACK_FAILURE = 1
+
+    def __init__(self, port):
+        self.port = port
+
+    def toggle_mux_to_torA(self, port):
```
jleveque, line 56: Do we need to pass `port` to all these functions? The port is passed in in the initializer; therefore the vendor can simply refer to `self.port`.

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,602 @@
...
+                 an Integer, the actual physical port connected to a Y cable
+            fwfile:
+                 a string, a path to the file which contains the firmware image.
+                 Note that the firmware file can be in the format of the vendor's
+                 choosing (binary, archive, etc.). But note that it should be one file
+                 which contains firmware for all components of the Y-cable. In case the
+                 vendor chooses to pass this file in activate_firmware, the API should
+                 have the logic to retreive the firmware version from this file
+                 which has to be activated on the componenets of the Y-Cable
+                 this API has been called for.
```
jleveque, line 377: Should mention that `fwfile` is optional. If `None`, the cable should activate whatever firmware is marked to be activated next. If provided, it should retreive the firmware version(s) from this file, ensure they are downloaded on the cable, then activate them.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,602 @@
...
+            port:
+                 an Integer, the actual physical port connected to a Y cable
+            fwfile:
+                 a string, a path to the file which contains the firmware image.
+                 Note that the firmware file can be in the format of the vendor's
+                 choosing (binary, archive, etc.). But note that it should be one file
+                 which contains firmware for all components of the Y-cable. In case the
+                 vendor chooses to pass this file in activate_firmware, the API should
+                 have the logic to retreive the firmware version from this file
+                 which has to be activated on the componenets of the Y-Cable
```
jleveque, line 376: s/componenets/components/

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,602 @@
...
+        Args:
+            port:
+                an Integer, the actual physical port connected to Y end of a Y cable which can toggle the MUX
+
+        Returns:
+            a Boolean, True if the toggle succeeded and False if it did not succeed.
+        """
+        raise NotImplementedError
+
+    def check_read_side(self, port):
```
jleveque, line 91: Maybe rename `get_read_side`

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,602 @@
...
+            One of the following predefined constants:
+                TARGET_TOR_A, if reading the Y cable from TOR A side.
+                TARGET_TOR_B, if reading the Y cable from TOR B side.
+                TARGET_NIC, if reading the Y cable from NIC side.
+                TARGET_UNKNOWN, if reading the Y cable API fails.
+        """
+
+        raise NotImplementedError
+
+    def check_mux_direction(self, port):
```
jleveque, line 110: Maybe rename `get_mux_direction`

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,602 @@
...
+        Returns:
+            One of the following predefined constants:
+                TARGET_TOR_A, if mux is pointing to TOR A side.
+                TARGET_TOR_B, if mux is pointing to TOR B side.
+                TARGET_UNKNOWN, if mux direction API fails.
+        """
+
+        raise NotImplementedError
+
+    def check_active_linked_tor_side(self, port):
```
jleveque, line 129: Maybe rename `get_active_linked_tor_side`

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,602 @@
...
+        Returns:
+            One of the following predefined constants:
+                TARGET_TOR_A, if TOR A is actively linked and sending traffic.
+                TARGET_TOR_B, if TOR B is actively linked and sending traffic.
+                TARGET_UNKNOWN, if checking which side is linked and sending traffic API fails.
+        """
+
+        raise NotImplementedError
+
+    def check_if_link_is_active_for_NIC(self, port):
```
jleveque, line 147: Maybe rename `is_nic_link_active`. Or better yet, maybe `is_link_active(target)`, and pass in the side to check. With the latter, you could also remove `check_if_link_is_active_for_torA` and `check_if_link_is_active_for_torB`

vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,602 @@
...
+    # Valid return codes for upgrade firmware routine steps
+    FIRMWARE_DOWNLOAD_SUCCESS = 0
+    FIRMWARE_DOWNLOAD_FAILURE = 1
+    FIRMWARE_ACTIVATE_SUCCESS = 0
+    FIRMWARE_ACTIVATE_FAILURE = 1
+    FIRMWARE_ROLLBACK_SUCCESS = 0
+    FIRMWARE_ROLLBACK_FAILURE = 1
+
+    def __init__(self, port):
+        self.port = port
```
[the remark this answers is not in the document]
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed, removed from all the API's

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,618 @@
...
+
+        Args:
+
+        Returns:
+            a Boolean, True if the toggle succeeded and False if it did not succeed.
+        """
+
+        raise NotImplementedError
+
+    def toggle_mux_to_torB(self):
```
theasianpianist, line 79: Same as above

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,636 @@
...
+
+        Args:
+
+        Returns:
+            a float, the voltage of the local MCU
+        """
+
+        raise NotImplementedError
+
+    def enable_prbs_mode(self, target, mode_value, lane_map):
```
jleveque, line 469: Everything below here appears to be debug functions, correct? I suggest adding a comment to visibly separate this section. Something like:

```python
#############################################################################################
###                                  Debug Functionality                                  ### 
#############################################################################################
```

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,636 @@
...
+        The port on which this API is called for can be referred using self.port.
+
+        Args:
+
+        Returns:
+            a boolean, True if debug mode enabled
+                     , False if debug mode not enabled
+        """
+
+        raise NotImplementedError
```
jleveque, line 618: Suggest condensing these into two functions:

`set_debug_mode(bool)`
`get_debug_mode()`

Also suggest moving them to the top of the debug functionality section.

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,636 @@
...
+    with a vendor specific Y-Cable
+"""
+
+
+class YCableBase(object):
+
+    # definitions of targets for getting the various fields/cursor
+    # equalization parameters from the register spec
+    # the name of the target denotes which side MCU
+    # values will be retreived on the Y-Cable
```
jleveque, line 17: s/retreived/retrieved from or written to/

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,636 @@
...
+    EYE_PRBS_TARGET_NIC = 3
+
+    # definitions of switch counter types
+    # to be entered by the user in get_switch_count api
+    # for retreiving the counter values
+
+    SWITCH_COUNT_MANUAL = "manual"
+    SWITCH_COUNT_AUTO = "auto"
+
+    # switching modes inside muxcable
```
jleveque, line 41: More inconsistent nomenclature: `muxcable`. I think we should stick with `Y-cable` everywhere at this point.

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,618 @@
...
+
+    def __init__(self, port):
+        """
+        Args:
+            port:
+                 an Integer, the actual physical port connected to a Y cable
+        """
+        self.port = port
+
+    def toggle_mux_to_torA(self):
```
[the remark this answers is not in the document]
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: added

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,636 @@
+"""
+    y_cable_base.py
+
+    Base class for implementing Y-Cable specific functionality in SONiC.
+    This is the base class of sonic_y_cable implentation. A vendor specific 
+    implementation of YCable needs to inherit from this class.
```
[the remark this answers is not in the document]
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,636 @@
+"""
+    y_cable_base.py
+
+    Base class for implementing Y-Cable specific functionality in SONiC.
+    This is the base class of sonic_y_cable implentation. A vendor specific
```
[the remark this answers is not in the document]
jleveque: Doesn't appear to be fixed.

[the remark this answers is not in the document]
vdahiya12: there happened to be more than one instance of `vendor specific`, changed them to `vendor-specific`. fixed

jleveque (changes requested):

xinyulin (commented):
comment for get_ber_info() and get_debug_mode()

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1013 @@
...
+                FIRMWARE_DOWNLOAD_SUCCESS
+                FIRMWARE_DOWNLOAD_FAILURE
+
+                a predefined code stating whether the firmware download was successful
+                or an error code as to what was the cause of firmware download failure
+        """
+
+        raise NotImplementedError
+
+    def activate_firmware(self, fwfile=None):
```
xinyulin, line 431: Is it necessary to add additional input parameter 'hitless' to enable/disable hitless upgrade 
```suggestion
    def activate_firrmware(self, fwfile=None, hitless=False):
```

- Hitless upgrade: it will backup/restore the current state (ex. variables of link status, API attributes...etc.) before and after firmware upgrade.

- Non-hitless upgrade: it will update the firmware regardless the current status, a link flip can be observed during the upgrade.

Ideally, the vendor should handle the hitless/non-hitless upgrade in their firmware or python API automatically, but this optional could be useful if we would like to do troubleshooting or reset the module reinitialize the chip especially for the y-cable which is not designed to issue a module reset (write lower page, byte93, bit7 to 1) from NIC end

sonic_y_cable/y_cable_base.py
```diff
@@ -327,7 +336,8 @@ def get_switch_count_target(self, switch_count_type, target, clear_on_read=False
     def get_target_cursor_values(self, lane, target):
         """
         This API returns the cursor equalization parameters for a target(NIC, TOR_A, TOR_B).
-        This includes pre one, pre two , main, post one, post two cursor values
+        This includes pre one, pre two , main, post one, post two , post three cursor values
```
jleveque, line 339: Remove spaces before commas

sonic_y_cable/y_cable_base.py
```diff
@@ -343,7 +353,7 @@ def get_target_cursor_values(self, lane, target):
                      TARGET_TOR_A -> TORA,
                      TARGET_TOR_B -> TORB
         Returns:
-            a list, with  pre one, pre two , main, post one, post two cursor values in the order
+            a list, with  pre one, pre two , main, post one, post two , post three cursor values in the order
```
jleveque, line 356: Remove spaces before commas

sonic_y_cable/y_cable_base.py
```diff
@@ -362,7 +372,7 @@ def set_target_cursor_values(self, lane, cursor_values, target):
                              3 -> lane 3
                              4 -> lane 4
             cursor_values:
-                a list, with  pre one, pre two , main, post one, post two cursor etc. values in the order
+                a list, with  pre one, pre two , main, post one, post two cursor, post three etc. values in the order
```
jleveque, line 375: Remove spaces before commas

sonic_y_cable/y_cable_base.py
```diff
@@ -285,6 +292,8 @@ def get_switch_count_tor_b(self, clear_on_read=False):
         """
         This API returns the switch count to change the Active TOR which has
         been done manually by the user initiated from ToR B
+        This is essentially all the successful switches initiated from ToR B. Toggles which were
+        initiated to toggle from ToR B and did not succed do not count.
```
jleveque, line 296: s/succed/succeed/

sonic_y_cable/y_cable_base.py
```diff
@@ -616,15 +643,35 @@ def create_port(self, speed, fec_mode=FEC_MODE_NONE, anlt=False):
                 50000 -> 50G
                 100000 -> 100G
 
-            fec_mode:
-                One of the following predefined constants, the actual fec mode for the port to be configured:
+            fec_mode_tor_a:
+                One of the following predefined constants, the actual fec mode for the tor A to be configured:
```
jleveque, line 647: s/tor/ToR/

Please make this change throughout all docstrings for consistency

jleveque, line 647: s/fec/FEC/

Please make this change throughout all docstrings for consistency

sonic_y_cable/y_cable_base.py
```diff
@@ -706,12 +768,17 @@ def set_anlt(self, enable):
 
         raise NotImplementedError
 
-    def get_anlt(self):
+    def get_anlt(self, target):
         """
         This API gets the mode of the cable for corresponding lane configuration.
```
jleveque, line 773: s/mode/auto-negotiation + link training (AN/LT) mode/
s/lane configuration/port/

sonic_y_cable/y_cable_base.py
```diff
@@ -706,12 +768,17 @@ def set_anlt(self, enable):
...
         The port on which this API is called for can be referred using self.port.
 
         Args:
+            target:
+                One of the following predefined constants, the actual target to get the anlt on:
```
jleveque, line 778: Replace "anlt" with "AN/LT mode"

sonic_y_cable/y_cable_base.py
```diff
@@ -803,6 +880,11 @@ def restart_anlt(self):
         The port on which this API is called for can be referred using self.port.
 
         Args:
+            target:
+                One of the following predefined constants, the actual target to restart anlt on:
```
jleveque, line 884: Replace "anlt" with "AN/LT". Do this throughout all docstrings for consistency.

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,631 @@
...
+                     EYE_PRBS_TARGET_TOR_A -> TOR A
+                     EYE_PRBS_TARGET_TOR_B -> TOR B
+                     EYE_PRBS_TARGET_NIC -> NIC
+        Returns:
+            a list, with EYE values of lane 0 lane 1 lane 2 lane 3 with corresponding index
+        """
+
+        raise NotImplementedError
+
+    def get_ber_info(self, target):
```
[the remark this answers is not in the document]
vdahiya12: I think we can move this to debug section, I will update thanks

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,631 @@
...
+
+
+        Returns:
+            a boolean, True if the enable is successful
+                     , False if the enable failed
+        """
+
+        raise NotImplementedError
+
+    def get_debug_mode(self):
```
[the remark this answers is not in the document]
vdahiya12: Is it possible to have this synchronized across both ToR's? 
If not we might have to make this API as optional/or else remove it.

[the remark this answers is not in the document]
vdahiya12: @jleveque

sonic_y_cable/y_cable_base.py
vdahiya12: I will put in a param for hitless. 
Thanks

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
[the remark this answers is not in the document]
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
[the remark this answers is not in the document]
jleveque: This was intended to be a software mechanism which would allow debug functionality if enabled and stop/disable debug functionality if disabled. However, these actions were all considered to be performed from the same ToR. Synchronization across ToRs was not considered.

@xinyulin: Do you foresee problems with this approach? Does the other ToR side need to be aware that debugging is occurring on the other side?

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1130 @@
...
+                    PRBS_DIRECTION_CHECKER
+
+        Returns:
+            a boolean, True if the disable is successful
+                     , False if the disable failed
+        """
+
+        raise NotImplementedError
+
+    def enable_loopback_mode(self, target, mode=NEAR_END_LOOPBACK, lane_mask):
```
xinyulin, line 1012:
```suggestion
    def enable_loopback_mode(self, target, lane_mask, mode=LOOPBACK_MODE_NEAR_END):
```
Suggest to swap the order of lane_mask and mode to fix 'non-default argument follows default argument' syntax error and correct the variable name with 'LOOPBACK_MODE_NEAR_END'

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1130 @@
...
+                a boolean, True if the count has to be reset after read to zero
+                         , False if the count is not to be reset after read
+            Returns:
+                an integer, the number of times the Y-cable has been switched
+        """
+
+        raise NotImplementedError
+
+    def get_switch_count_tor_a(self, clear_on_read=False):
+        """
```
xinyulin, line 255: Both get_switch_count_tor_a() and get_switch_count_tor_b() are not supported in current MCU firmware. The cable can only return the total auto/manual switch count by reading the upper page 0x5, byte 141 to 148. If this information is useful for diagnostics, I will update the API document and  MCU FW to support these two APIs from next firmware release.

sonic_y_cable/y_cable_base.py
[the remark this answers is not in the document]
xinyulin: @jleveque: The only concern is the debug mode status is not synchronized. In application, the other ToR side don't need to be aware that debugging mode of the other side. As you mentioned, it should be fine if all the debug actions were being performed from the same ToR.

@vdahiya12: I will consider to find a simple way to synchronize debug mode across ToRs for next firmware update

sonic_y_cable/y_cable_base.py
vdahiya12: sounds good

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1142 @@
...
+
+        Returns:
+            a boolean, True if the cable is target reset
+                     , False if the cable target is not reset
+        """
+
+        raise NotImplementedError
+
+    def create_port(self, speed, fec_mode_tor_a=FEC_MODE_NONE, fec_mode_tor_b=FEC_MODE_NONE,
+                    fec_mode_nic=FEC_MODE_NONE, anlt_tor_a=False, anlt_tor_b= False, anlt_nic=False):
```
xinyulin, line 631: Is it make sense to change the input parameter as suggestion?
For current y_cable firmware implementation, both TOR A and TOR B must be operated in the same mode. it is not allowed to configure AN/LT, FEC on TOR A and B individually.

```suggestion
                    fec_mode_nic=FEC_MODE_NONE, anlt_tor=False, anlt_nic=False):
```

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1142 @@
...
+                     TARGET_TOR_B -> TORB
+
+        Returns:
+            a boolean, True if the cable is target reset
+                     , False if the cable target is not reset
+        """
+
+        raise NotImplementedError
+
+    def create_port(self, speed, fec_mode_tor_a=FEC_MODE_NONE, fec_mode_tor_b=FEC_MODE_NONE,
```
xinyulin, line 630:
```suggestion
    def create_port(self, speed, fec_mode_tor=FEC_MODE_NONE, 
```

sonic_y_cable/y_cable_base.py
xinyulin: Originally, we defined a total manual switching count in upper page 0x4, byte 141-144. In order to support these two APIs

- get_switch_count_tor_a()
- get_switch_count_tor_b()

I'd like to change the definition as below

- **upper page 0x4, byte 141-144**: number of manual switching count initiated from **TOR A**
- **upper page 0x4, byte 225-229**: number of manual switching count initiated from **TOR B**

and the get_switch_count_total() will return the sum of TOR A (byte 141-144) and TOR B (byte 225 - 228).

sonic_y_cable/y_cable_base.py
```diff
@@ -79,14 +72,20 @@ def __init__(self, log_identifier):
...
         Args:
             port:
                  an Integer, the actual physical port connected to a Y cable
         """
         self.port = port
 
+        super(YCableBase, self).__init__(
+        log_identifier=log_identifier,
+        log_facility=Logger.LOG_FACILITY_DAEMON,
+        log_option=(Logger.LOG_OPTION_NDELAY | Logger.LOG_OPTION_PID)
```
jleveque, line 86: These lines should be indented.

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
```diff
@@ -72,19 +70,33 @@ class YCableBase(Logger):
...
 
-    def __init__(self, port, log_identifier):
+    def __init__(self, port, main_logger):
         """
         Args:
             port:
                  an Integer, the actual physical port connected to a Y cable
+            main_logger:
+                 a logging instance object, the main logging instance which can be used to log messages into syslog
+                 which can be called from derived classes using functions defined as log_info, log_warning etc.
```
jleveque, line 80: I think we can just call it `logger` instead of `main_logger`. The docstring should mention that it must be an instance of the Logger class from sonic_py_common instantiated by the program which is instantiating this class.

sonic_y_cable/y_cable_base.py
```diff
@@ -72,19 +70,33 @@ class YCableBase(Logger):
...
+
         """
         self.port = port
+        self.main_logger = main_logger
```
jleveque, line 84: Again, suggest simplifying to `self.logger`. Can also name it `self._logger` to imply that it is private, so that no one tries to call it directly and only calls the `log_warning()`, etc. methods.

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -70,33 +70,33 @@ class YCableBase():
...
+                 by the program which is instantiating this class.
 
         """
         self.port = port
-        self.main_logger = main_logger
+        self._logger = logger
 
     def log_warning(self, msg):
-        self.main_logger.log_warning("y_cable_port {} {}".format(self.port, msg))
+        self._logger.log_warning("y_cable_port {} {}".format(self.port, msg))
```
jleveque, line 87: Suggest to add something like a colon as a delimiter. E.g, `y_cable_port {}: {}`. Do this for all log functions.

jleveque (changes requested):

sonic_y_cable/y_cable_base.py
vdahiya12: fixed

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1142 @@
...
+        Returns:
+           a dictionary:
+               a detailed format agreed upon by vendors
+        """
+
+        raise NotImplementedError
+
+#############################################################################################
+###                                  Debug Functionality                                  ###
+#############################################################################################
```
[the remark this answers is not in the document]
vdahiya12: I dont think there is any other way for now, other than to reload the confg_db.json

sonic_y_cable/y_cable_base.py
vdahiya12: fixed, can you please check

sonic_y_cable/y_cable_base.py
vdahiya12: fixed, can you please check

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1157 @@
...
+    SWITCHING_MODE_AUTO = 1
+
+    # definitions of the FEC mode to be configured
+    # on a port/cable
+    FEC_MODE_NONE = 0
+    FEC_MODE_RS = 1
+    FEC_MODE_FC = 2
+
+    # definitions of the modes to be run for loopback mode
+    # on the port/cable
```
xinyulin, line 56: Shall we add LOOPBACK_MODE_NONE if the loopback mode is not enabled?

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1157 @@
...
+                    EYE_PRBS_LOOPBACK_TARGET_LOCAL -> local side,
+                    EYE_PRBS_LOOPBACK_TARGET_TOR_A -> TOR A
+                    EYE_PRBS_LOOPBACK_TARGET_TOR_B -> TOR B
+                    EYE_PRBS_LOOPBACK_TARGET_NIC -> NIC
+
+        Returns:
+            mode_value:
+                One of the following predefined constants, the mode to be run for loopback:
+                    LOOPBACK_MODE_NEAR_END
+                    LOOPBACK_MODE_FAR_END
```
xinyulin, line 1108: Shall we return LOOPBACK_MODE_NONE if this feature is not enabled?

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1157 @@
...
+                         , False if the log is not to be cleared after read
+
+        Returns:
+           list:
+              a list of strings which correspond to the event logs of the cable
+        """
+
+        raise NotImplementedError
+
+    def get_pcs_stats(self, target):
```
xinyulin, line 822:
```suggestion
    def get_pcs_stats(self, target, clear_on_read=False):
```
Is it necessary to add clear_on_read to reset the pcs statistics?

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1157 @@
...
+                     TARGET_TOR_B -> TORB
+
+        Returns:
+           a dictionary:
+               a detailed format agreed upon by vendors
+        """
+
+        raise NotImplementedError
+
+    def get_fec_stats(self, target):
```
xinyulin, line 841:
```suggestion
    def get_fec_stats(self, target, clear_on_read=False):
```
Is it necessary to add clear_on_read to reset the fec statistics?

sonic_y_cable/y_cable_base.py
vdahiya12: I will update the PR

vdahiya12: fixed

sonic_y_cable/y_cable_base.py
vdahiya12: if the feature cannot be supported, you can return an error when the caller tries to enable clear_on_read= True;
I believe other vendors might be able to support this, hence it is kept as an argument.

sonic_y_cable/y_cable_base.py
vdahiya12: if the feature cannot be supported, you can return an error when the caller tries to enable clear_on_read= True;
I believe other vendors might be able to support this, hence it is kept as an argument.

lguohan (approved):
:shipit:

prgeor (changes requested):

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1157 @@
...
+                 they are downloaded on the cable, then activate them.
+
+            hitless (optional):
+                a boolean, True, Hitless upgrade: it will backup/restore the current state
+                                 (ex. variables of link status, API attributes...etc.) before
+                                 and after firmware upgrade.
+                a boolean, False, Non-hitless upgrade: it will update the firmware regardless
+                                  the current status, a link flip can be observed during the upgrade.
+        Returns:
+            One of the following predefined constants:
```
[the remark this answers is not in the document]
vdahiya12: Currently all vendors need to support hitless by default. The firmware error codes are going to be in a future PR as they are under discussion by vendors.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1157 @@
...
+                     TARGET_TOR_A -> TORA,
+                     TARGET_TOR_B -> TORB
+        Returns:
+            a boolean, True if cursor values setting is successful
+                     , False if cursor values setting is not successful
+        """
+
+        raise NotImplementedError
+
+    def get_firmware_version(self, target):
```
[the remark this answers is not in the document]
vdahiya12: Yes, depends on vendor. Some vendors have a unique version on all MCU's whereas some could have different versions.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1159 @@
...
+        Returns:
+            One of the following predefined constants:
+                TARGET_TOR_A, if TOR A is actively linked and sending traffic.
+                TARGET_TOR_B, if TOR B is actively linked and sending traffic.
+                TARGET_UNKNOWN, if checking which side is linked and sending traffic API fails.
+        """
+
+        raise NotImplementedError
+
+    def is_link_active(self, target):
```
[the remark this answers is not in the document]
vdahiya12: currently by default if an API fails we return a generic error EEPROM_FAILED. This works well since if the RC is this code, we put an N/A inside the value, otherwise True or False indicates the actual return value.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1159 @@
...
+                 either SWITCHING_MODE_AUTO or SWITCHING_MODE_MANUAL
+
+        Returns:
+            a Boolean, True if the switch succeeded and False if it did not succeed.
+        """
+
+        raise NotImplementedError
+
+    def get_switching_mode(self):
+        """
```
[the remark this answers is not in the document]
vdahiya12: by default the mode has to be manual. There can never be a no mode value. it is either auto or manual

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1159 @@
...
+
+        Args:
+
+        Returns:
+            a float, the voltage of the local MCU
+        """
+
+        raise NotImplementedError
+
+    def get_alive_status(self):
```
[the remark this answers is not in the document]
vdahiya12: No link active is different from get_alive_status. get_alive_status just checks whether all sides of the cable are plugged in or not.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1159 @@
...
+                an Integer, the time value for hysteresis to be set in milliseconds
+
+        Returns:
+            a boolean, True if the time is configured
+                     , False if the time is not configured
+        """
+
+        raise NotImplementedError
+
+    def get_autoswitch_hysteresis_timer(self):
```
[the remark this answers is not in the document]
vdahiya12: this should return the value corresponding to the eeprom offset where the definition is. All vendors could have different hysteresis_timer values but default is 5 ms

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1157 @@
...
+            clear_on_read:
+                a boolean, True if the count has to be reset after read to zero
+                         , False if the count is not to be reset after read
+            Returns:
+                an integer, the number of times manually the Y-cable has been switched
+        """
+
+        raise NotImplementedError
+
+    def get_target_cursor_values(self, lane, target):
```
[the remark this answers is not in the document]
vdahiya12: Let me make a note for this, all vendors will also need to be notfied for this. We can put this in a future PR.

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1157 @@
...
+                     TARGET_NIC -> NIC,
+                     TARGET_TOR_A -> TORA,
+                     TARGET_TOR_B -> TORB
+        Returns:
+            a list, with  pre one, pre two, main, post one, post two, post three cursor values in the order
+        """
+
+        raise NotImplementedError
+
+    def set_target_cursor_values(self, lane, cursor_values, target):
```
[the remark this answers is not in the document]
vdahiya12: Let me make a note for this, all vendors will also need to be notfied for this. We can put this in a future PR

sonic_y_cable/y_cable_base.py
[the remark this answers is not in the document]
vdahiya12: Following on that all API's if they fail can just return a negative error code, which if not defined in base class.
can be subsitituted by EEPROM_ERROR=-1

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1159 @@
...
+        Args:
+
+        Returns:
+            a string, with part number
+        """
+
+        raise NotImplementedError
+
+    def get_switch_count_total(self, switch_count_type, clear_on_read=False):
+        """
```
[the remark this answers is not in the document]
vdahiya12: noted for a further enhancement PR

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1159 @@
...
+                a boolean, True if the count has to be reset after read to zero
+                         , False if the count is not to be reset after read
+
+            Returns:
+                an integer, the number of times the Y-cable has been switched from ToR B
+        """
+
+        raise NotImplementedError
+
+    def get_switch_count_target(self, switch_count_type, target, clear_on_read=False):
```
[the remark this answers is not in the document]
vdahiya12: noted for a further enhancement PR

sonic_y_cable/y_cable_base.py
```diff
@@ -0,0 +1,1159 @@
...
+        Args:
+
+        Returns:
+            a boolean, True if the cable is alive
+                     , False if the cable is not alive
+        """
+
+        raise NotImplementedError
+
+    def reset(self, target):
```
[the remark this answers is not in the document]
vdahiya12: if target is NIC this will reset.

PR #187, opened 2021-05-07
[thermal_manager_base] Add a stop function to thermal manager

Junchao-Mellanox:
#### Description

Add a stop function to thermal manager

#### Motivation and Context

Allow thermalctld to call the stop function to stop run_policy function fast

sonic-mgmt regression


liat-grozovik:
@Junchao-Mellanox LGTM but can you explain why you need it?  is there a flow we wish to fix as part of this faster approach?
Can you please add unit test to cover this flow?

Junchao-Mellanox:
Unit test case added.

PR #190, opened 2021-05-30
Update eeprom_base.py

### LINKED ISSUE #170: [sonic_eeprom] Class methods shouldn't take the EEPROM data as a parameter
EEPROM class should obtain EEPROM data when instantiated and store as a class member, then use that data in methods rather than having to pass the raw data into the methods. It defeats the purpose of object-orientation.

Piyushagg19:

### DIFF
--- a/sonic_platform_base/sonic_eeprom/eeprom_base.py
+++ b/sonic_platform_base/sonic_eeprom/eeprom_base.py
@@ -29,6 +29,7 @@ class EepromDecoder(object):
         self.r = readonly
         self.cache_name = None
         self.cache_update_needed = False
+        self.eeprom_file_handle = None
         self.lock_file = None
 
     def check_status(self):
@@ -220,7 +221,8 @@ class EepromDecoder(object):
         except Exception:
             pass
         self.cache_update_needed = using_eeprom
-        return io.open(eeprom_file, "rb")
+        self.eeprom_file_handle = io.open(eeprom_file, "rb")
+        return using_eeprom
 
     def read_eeprom(self):
         sizeof_info = 0
@@ -231,8 +233,9 @@ class EepromDecoder(object):
 
     def read_eeprom_bytes(self, byteCount, offset=0):
         F = None
+        eeprom_flag = False
         try:
-            F = self.open_eeprom()
+            F = self.eeprom_file_handle
             F.seek(self.s + offset)
             o = F.read(byteCount)
 
@@ -243,9 +246,11 @@ class EepromDecoder(object):
                 os.remove(self.cache_name)
                 self.cache_update_needed = True
                 F.close()
-                F = self.open_eeprom()
-                F.seek(self.s + offset)
-                o = F.read(byteCount)
+                eeprom_flag = self.open_eeprom()
+                if (eeprom_flag == self.cache_update_needed):
+                    F = self.eeprom_file_handle
+                    F.seek(self.s + offset)
+                    o = F.read(byteCount)
 
             if len(o) != byteCount:
                 raise RuntimeError("Expected to read %d bytes from %s, "


Piyushagg19:
@jleveque: Can you please do the review?

sonic_platform_base/sonic_eeprom/eeprom_base.py
```diff
@@ -220,7 +221,8 @@ def open_eeprom(self):
         except Exception:
             pass
         self.cache_update_needed = using_eeprom
-        return io.open(eeprom_file, "rb")
+        self.eeprom_file_handle = io.open(eeprom_file, "rb")
+        return using_eeprom
```
jleveque, line 225: Unfortunately, changing the return value will break all existing code which is already using this library. These changes will need to be implemented incrementally. This PR should store the EEPROM file handle in a member variable as you have done, but it should not break existing functionality. All methods which currently take the EEPROM file handle as a parameter should be modified such that the parameter defaults to `None`, and if the handle is `None`, then the function should use `self.eeprom_file_handle`. This will allow vendors to update their existing code to stop passing the handle to methods. Once all vendors have transitioned, then we can change this return value to a boolean.

jleveque (changes requested):

sonic_platform_base/sonic_eeprom/eeprom_base.py
Piyushagg19: @jleveque: Does it mean that we have to store the file handle again in a temporary variable in the `read_eeprom_bytes` function but use the `eeprom_file_handle` after the function call to update the `eeprom_file` file path?

jleveque: `read_eeprom_bytes` does not take the EEPROM file handle as a parameter. However, other methods do.

Piyushagg19:
@jleveque: Can you please check if the update suffices all the requirements?

sonic_platform_base/sonic_eeprom/eeprom_base.py
```diff
@@ -256,8 +261,8 @@ def read_eeprom_bytes(self, byteCount, offset=0):
         finally:
             if F is not None:
                 F.close()
-
-        return bytearray(o)
+        self.byte_array = bytearray(o)
```
jleveque, line 264: We don't need to store this as a class member. It should be local only.

sonic_platform_base/sonic_eeprom/eeprom_base.py
```diff
@@ -29,6 +29,9 @@ def __init__(self, path, format, start, status, readonly):
         self.r = readonly
         self.cache_name = None
         self.cache_update_needed = False
+        self.eeprom_file_handle = None
+        self.o = None
```
jleveque, line 33: Prefer using a more descriptive name, like `eeprom_raw_bytes` or similar.

jleveque (changes requested):

jleveque:
> @jleveque: From what I understood, I modified the code to meet the requirements. Can you please check if the update suffices all the requirements? If not, please let me know the rest of the areas that require changes.

All methods which expect the full raw EEPROM bytes to be passed in should also be modified to default the parameter to `None` and then check if it is `None`, in which case it will use the new class member.

Piyushagg19:
@jleveque: Sorry for pestering you again but this is my first contribution to the open-source community. Hope you understand! I'm not able to understand what methods are you referring to when you say 'methods which expect the full raw EEPROM bytes to be passed in'. If you can please provide me with an example or mention the methods that require the changes, I'll be able to understand better and get the rest of the work done. Thank you.

Piyushagg19:
@jleveque: The update consists of two changes.
1. Reverting the `read_eeprom_bytes()` return statement back to its original state.
2. Using the more descriptive names of variables.

PR #196, opened 2021-06-09
Fix typo for midplane APIs.

caizhenghui-juniper:
There is typo "return" vs "raise" in midplane APIs.

#### Description
In midplane API, default method should be raise NotImplementedError not return NotImplementedError

#### Motivation and Context
In case of platforms don't have the implementation for midplane API, the default value will not be set since there is no exception raised by this function.

Run the test on pmon without implementing the method and chassisd won't complain about line:
266                 # Update db with midplane information
267                 fvs = swsscommon.FieldValuePairs([(CHASSIS_MIDPLANE_INFO_IP_FIELD, midplane_ip),
268 B->                                               (CHASSIS_MIDPLANE_INFO_ACCESS_FIELD, str(midplane_access))])

### DIFF
--- a/sonic_platform_base/module_base.py
+++ b/sonic_platform_base/module_base.py
@@ -451,7 +451,7 @@ class ModuleBase(device_base.DeviceBase):
             A string, the IP-address of the module reachable over the midplane
 
         """
-        return NotImplementedError
+        raise NotImplementedError
 
     def is_midplane_reachable(self):
         """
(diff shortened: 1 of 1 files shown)


anshuv-mfst:
Hi @jleveque - Could you please take a look, minor change in platform for chassis. Thanks!

jleveque (approved):

PR #204, opened 2021-07-06
[sfp] Update lpmode API and add get_power_set

andywongarista:
#### Description
- Add get_power_set() method for retrieving the power-set status of the SFP
- Update lpmode accessors to indicate raising NotImplementedError as expected behaviour if not supported on a platform

#### Motivation and Context

Some platforms do not have a controllable lpmode pin that can be used to access the lpmode status of a xcvr. However, the xcvr itself may have a controllable lpmode mechanism via its EEPROM. To make use of this, platforms that don't have a controllable lpmode pin should raise NotImplementedError. Common code (like sfputil) will be updated in a subsequent PR to use the EEPROM lpmode mechanism instead when NotImplementedError is raised from using the lpmode API.

Adding a new get_power_set method because it will be used to determine the lpmode status of a xcvr when making use of the EEPROM lpmode mechanism (i.e. if power_override is true, then lpmode == power_set).

Nothing to test as this is just adding a new abstract method.

### DIFF
--- a/sonic_platform_base/sfp_base.py
+++ b/sonic_platform_base/sfp_base.py
@@ -248,6 +248,8 @@ class SfpBase(device_base.DeviceBase):
         """
         Retrieves the lpmode (low power mode) status of this SFP
 
+        Raises NotImplementedError if the lpmode pin is not available for this SFP
+
         Returns:
             A Boolean, True if lpmode is enabled, False if disabled
         """
@@ -259,6 +261,15 @@ class SfpBase(device_base.DeviceBase):
 
         Returns:
             A Boolean, True if power-override is enabled, False if disabled
+        """
+        raise NotImplementedError
+
+    def get_power_set(self):
+        """
+        Retrieves the power-set status of this SFP
+
+        Returns:
+            A Boolean, True if power-set is enabled, False if disabled
         """
         raise NotImplementedError
 
@@ -354,6 +365,8 @@ class SfpBase(device_base.DeviceBase):
     def set_lpmode(self, lpmode):
         """
         Sets the lpmode (low power mode) of SFP
+
+        Raises NotImplementedError if the lpmode pin is not available for this SFP
 
         Args:
             lpmode: A Boolean, True to enable lpmode, False to disable it


prgeor:
<<< Some platforms do not have a controllable lpmode pin that can be used to access the lpmode status of a xcvr
Do you mean, that the QSFP LPmode pin is missing from the module or the platform choose not to expose this pin for software control?

sonic_platform_base/sfp_base.py
```diff
@@ -262,6 +264,15 @@ def get_power_override(self):
         """
         raise NotImplementedError
 
+    def get_power_set(self):
```
prgeor, line 267: I think the "Power set" wording is taken from SFF 8636. What does get_power_set() mean for QSFP-DD or modules that follow CMIS spec?

sonic_platform_base/sfp_base.py
```diff
@@ -248,6 +248,8 @@ def get_lpmode(self):
         """
         Retrieves the lpmode (low power mode) status of this SFP
```
prgeor, line 249: can you modify the description to "Sets the lpmode via LPMode pin of this SFP"

sonic_platform_base/sfp_base.py
```diff
@@ -248,6 +248,8 @@ def get_lpmode(self):
         """
```
prgeor, line 248: Does this API retrieves LPMode signal status or whether the module is in low power mode?

andywongarista: I believe this API retrieves the hardware LPMode signal status; the corresponding set_lpmode API seems to imply this since it notes that it can be overridden by set_power_override.

sonic_platform_base/sfp_base.py
andywongarista: Yes, from what I can tell the original lpmode/power override APIs were designed only with SFF 8436/8636 in mind, and get_power_set would go with them. It doesn't really a have a direct equivalent in CMIS. My interpretation would be that LowPwrRequestSW in CMIS rev 5 (ForceLowPwr in rev 4) essentially behaves like power_set AND power_override, i.e.:
- If LowPwrRequestSW is 1, then power_override is 1 and power_set is 1
- If LowPwrRequestSW is 0, then power_override is 0 and power_set is X
For CMIS we may want to design new APIs for power-related functionality. For now this is sufficient since the only (Arista) platforms we're aware of that don't support the lpmode pin are using 8436/8636-compliant xcvrs and not CMIS-compliant xcvrs.

andywongarista:
> <<< Some platforms do not have a controllable lpmode pin that can be used to access the lpmode status of a xcvr
> Do you mean, that the QSFP LPmode pin is missing from the module or the platform choose not to expose this pin for software control?

The latter.

sonic_platform_base/sfp_base.py
prgeor: my only concern here is that these APIs will be used by same port that support optics following two different spec and api wording doesn't seem meaning for CMIS. Rest of the (get/set) APIs in sfp_base.py is applicable for almost all optics.  How is get_power_set() going to be used. ? Can't this be absorbed in platform specific sfp.py?

prgeor:
@andywongarista should this draft be now closed? if so, please do so.

PR #212, opened 2021-08-12
[sonic_platform_base/sonic_sfp] Fix QSFP-DD power class mask

snider-nokia:
#### Description
    Power class most significant bit is being improperly masked off

#### Motivation and Context
    QSFP-DD Power class is being set incorrectly 

    I verified that Power class is now being set correctly


snider-nokia:
@lguohan @judyjoseph @abhisekb Can you please review?

snider-nokia:
@prgeor, Can you merge this PR ?

PR #221, opened 2021-10-15
Add retry reading/setting mux status to simulated y-cable driver

wangxin:
#### Description

Add retry reading/setting mux status to simulated y-cable driver

#### Motivation and Context

When DUT is rebooted, xcvrd may call the simulated y-cable driver to get mux
direction before the mgmt interface is up. The simulated y-cable driver needs
to send HTTP request to mux simulator server to read mux status. It has
dependency on the mgmt interface. So, this could result in below error:
```
Oct 16 03:15:44.029933 sonic-dut ERR pmon#xcvrd[34]: y_cable_port 1: GET http://192.168.1.33:8082/mux/vms21-6/0 for physical_port 1 failed with URLError(OSError(113, 'No route to host'))
Oct 16 03:15:44.030306 sonic-dut ERR pmon#xcvrd[34]: Error: Could not establish the active side for  Y cable port Ethernet0 to perform read_y_cable update state db
```

This would cause other problem and may result in same interface on both upper
tor and lower tor remains in "standby" state.

The fix is to add retry to the simulated y-cable driver for reading or setting
mux status. The retry interval is 1 second. Retry timeout is 30 seconds.

The issue can be reliably reproduced on a 7260 dualtor testbed after run the test_acl::TestAclWithReboot cases.
With this fix, the issue cannot be reproduced.
Also tested the `config mux mode active` command.

### DIFF
--- a/sonic_y_cable/microsoft/y_cable_simulated.py
+++ b/sonic_y_cable/microsoft/y_cable_simulated.py
@@ -35,6 +35,7 @@ class YCable(YCableBase):
 
     POLL_TIMEOUT = 30
     POLL_INTERVAL = 1
+    URLOPEN_TIMEOUT = 5
 
     def __init__(self, port, logger):
         YCableBase.__init__(self, port, logger)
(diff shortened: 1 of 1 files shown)


bingwang-ms (approved):

sonic_y_cable/microsoft/y_cable_simulated.py
```diff
@@ -85,22 +89,30 @@ def _get(self, url=None):
...
+        while True:
             try:
-                req = urllib.request.Request(get_url)
-                with urllib.request.urlopen(req) as resp:
-                    return json.loads(resp.read().decode('utf-8'))
-            except urllib.error.HTTPError as e:
-                self.log_error('GET {} for physical_port {} failed with {}, detail: {}'.format(
+                try:
+                    req = urllib.request.Request(get_url)
+                    with urllib.request.urlopen(req) as resp:
```
vdahiya12, line 97: Should we add a timeout in the urlopen request as well ?
https://stackoverflow.com/questions/51227425/change-or-create-a-timeout-with-urllib-request

sonic_y_cable/microsoft/y_cable_simulated.py
```diff
@@ -85,22 +89,30 @@ def _get(self, url=None):
...
-                    e.read()))
-        except (urllib.error.URLError, json.decoder.JSONDecodeError, Exception) as e:
-            self.log_error('GET {} for physical_port {} failed with {}'.format(
-                get_url,
-                self.port,
-                repr(e)))
+                    repr(e)))
+
+            # Retry in case of exception, to workaround 'no route to host' issue after pmon restart
+            if (time.time() - start_time) > self.POLL_TIMEOUT:
```
vdahiya12, line 112: Should we add an error log here ?

vdahiya12 (commented):
Added a few comments

sonic_y_cable/microsoft/y_cable_simulated.py
wangxin: Good catch! Thanks!
Updated.

vdahiya12 (approved):

PR #222, opened 2021-10-22
[Y-Cable][Credo] add theading locker to support thread-safe calling, add SKU check for download_firmware API. 

xinyulin:
#### Description
Add threading locker to protect VSC protocol
This PR also adds SKU check for firmware download in the API.
so 50G or 100G cables wont be allowed to the alternate firmware versions, meaning 50G wont be allowed to download 100G and 100G wont be allowed 50G
#### Motivation and Context
To solve the racing condition when multiple threads access the same port at the same time

Tested on Arista 7050

### DIFF
--- a/sonic_y_cable/credo/y_cable_credo.py
+++ b/sonic_y_cable/credo/y_cable_credo.py
@@ -1381,6 +1381,14 @@ class YCable(YCableBase):
             fwImage = bytearray(inFile.read())
             inFile.close()
 
+            bin_pid = struct.unpack_from('>B', fwImage[5 : 6])[0]
+            mcu_pid = self.read_mmap(0xFB, 187)
+
+            if bin_pid != mcu_pid:
+                self.log_error('Firmware binary ID Mismatched Bin[%d] MCU[%d]' % (bin_pid, mcu_pid))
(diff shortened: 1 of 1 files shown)


vdahiya12:
@xinyulin Can we add more methods which require locking during firmware update to be added inside this.

xinyulin:
@vdahiya12 Is it acceptable to get ack_error=1 during firmware_activate()? If not, I think I'd need to add thread lock on most of APIs.
As I explained in the mail thread, the module is unable to response any i2c transaction during firmware activation.
If the y_cable_helper call any eeprom read/write related API (ex. get_read_side()) during firmware activation, it will get nothing but a NACK error.

vdahiya12:
> @vdahiya12 Is it acceptable to get ack_error=1 during firmware_activate()? If not, I think I'd need to add thread lock on most of APIs. As I explained in the mail thread, the module is unable to response any i2c transaction during firmware activation. If the y_cable_helper call any eeprom read/write related API (ex. get_read_side()) during firmware activation, it will get nothing but a NACK error.

Okay lets add the thread locker on API's that we can get rid of the NACK errors

PR #244, opened 2021-11-24
Firmware upgrade CLI support for QSFP-DD transceivers

prgeor:
#### Description
Changes to support firmware upgrade CLI for QSFP-DD transceivers

#### Motivation and Context
1. Refactored some of the APIs so as to show firmware download/upgrade progress
2. Fixed few crashes

1. Ran unit test
2. Firmware upgrade done successfully on Inphi, Molex and Centera QSFP-DD modules

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -23,6 +23,8 @@ class CmisApi(XcvrApi):
 
     def __init__(self, xcvr_eeprom):
         super(CmisApi, self).__init__(xcvr_eeprom)
+        self.vdm = CmisVdmApi(xcvr_eeprom)
+        self.cdb = CmisCdbApi(xcvr_eeprom)
 
     def get_model(self):
         '''
@@ -684,13 +686,9 @@ class CmisApi(XcvrApi):
         '''
         This function returns the application select code that each host lane has
         '''
-        apsel_dict = {}
-        if self.is_flat_memory():
-            for lane in range(1, self.NUM_CHANNELS+1):
-                apsel_dict["%s%d" % (consts.ACTIVE_APSEL_HOSTLANE, lane)] = 'N/A'
-        else:
-            apsel_dict = self.xcvr_eeprom.read(consts.ACTIVE_APSEL_CODE)
-        return apsel_dict
+        if (self.is_flat_memory()):
+            return 'N/A'
+        return self.xcvr_eeprom.read(consts.ACTIVE_APSEL_CODE)
 
     def get_tx_config_power(self):
         '''
@@ -764,7 +762,7 @@ class CmisApi(XcvrApi):
             'low warn' : 'N/A'
         }
 
-        if self.is_flat_memory():
+        if (self.is_flat_memory()):
             return laser_temp_dict
 
         try:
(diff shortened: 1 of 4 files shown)


sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -664,6 +666,8 @@ def get_active_apsel_hostlane(self):
         '''
         This function returns the application select code that each host lane has
         '''
+        if (self.is_flat_memory()):
+            return 'N/A'
```
qinchuanares, line 670: Example from a non-flat memory result would be 
{'ActiveAppSelLane1': 1, 'ActiveAppSelLane2': 1, 'ActiveAppSelLane3': 1, 'ActiveAppSelLane4': 1, 'ActiveAppSelLane5': 1, 'ActiveAppSelLane6': 1, 'ActiveAppSelLane7': 1, 'ActiveAppSelLane8': 1}.

Could you keep all the keys in the dictionary and make values 'N/A'?

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1327,16 +1335,16 @@ def module_fw_upgrade(self, imagepath):
...
         result = self.get_module_fw_info()
         try:
             _, _, _, _, _, _, _, _ = result['result']
         except (ValueError, TypeError):
             return result['status'], result['info']
-        result = self.get_module_fw_upgrade_feature()
-        try: 
+        result = self.get_module_fw_mgmt_feature()
+        try:
             startLPLsize, maxblocksize, lplonly_flag, autopaging_flag, writelength = result['result']
```
qinchuanares, line 1348: get_module_fw_mgmt_feature() return changed the key from 'result' to 'feature'. Need to change here too.

qinchuanares (commented):
Why remove test_get_module_fw_mgmt_feature()?

sonic_platform_base/sonic_xcvr/api/public/cmis.py
prgeor: done. good catch.

prgeor:
> test_get_module_fw_mgmt_feature

all hw reads are mocked and its just some string comparisons. it needs lot of effort to understand the parametrized mock input arguments which is quite long!

sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
```diff
@@ -128,6 +128,15 @@ def set_power_override(self, power_override, power_set):
     def get_eeprom_path(self):
         raise NotImplementedError
 
+    def set_optoe_write_max(self, write_max):
```
andywongarista, line 131: We should define an abstract `set_write_max(self, write_max)` method in SfpBase and then use this implementation for SfpOptoeBase. Platform vendors that don't use optoe would then implement set_write_max.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1119,16 +1116,23 @@ def get_module_fw_info(self):
             elif ImageBCommitted == 1:
                 CommittedImage = 'B'
             else:
-                CommittedImage = 'N/A'            
-            txt += 'Running Image: %s; Committed Image: %s\n' %(RunningImage, CommittedImage)
+                CommittedImage = 'N/A'
+            txt += 'Running Image: %s\n' % (RunningImage)
+            txt += 'Committed Image: %s\n' %(CommittedImage)
```
andywongarista, line 1121: Add space after %

andywongarista (changes requested):

lguohan (dismissed):

sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
prgeor: Sfpoptoe driver keeping the default write max to '1' is done for historical reasons. I don't see any reason why non-optoe drivers should keep the default write max other than SMBUS block write size of 32 bytes. Just wanted to avoid yet another non-implemented API in SfpBase. Do you see any platform that needs this change? if not, we can revisit later.

lguohan (approved):

PR #252, opened 2021-12-09
[ssd] Allow individual vendor parsers to handle errors

alexrallen:
#### Description

Removed a `return None` in the command execution helper function `_execute_shell` in the SSD class so that command output is still parsed in the event of a non-zero exit code. 

#### Motivation and Context

If there is an issue found with a SSD SMART utilities may return a non-zero error code and we would still want to parse the output so we should not bail.

In addition, with the `return None` the parsing functions throw stack traces during execution. When removed, the parsing functions are able to elegantly handle missing information and the fields are left as `N/A` so there is no need for handling here. 

Tested manually on a SN2010 with a malfunctioning SSD which was initially throwing stack traces and now handles the error elegantly reporting the health as `N/A` a much more informative error than a python stack trace.

### DIFF
--- a/sonic_platform_base/sonic_ssd/ssd_generic.py
+++ b/sonic_platform_base/sonic_ssd/ssd_generic.py
@@ -63,9 +63,6 @@ class SsdUtil(SsdBase):
     def _execute_shell(self, cmd):
         process = subprocess.Popen(cmd.split(), universal_newlines=True, stdout=subprocess.PIPE)
         output, error = process.communicate()
-        exit_code = process.returncode
-        if exit_code:
-            return None
         return output
 
     def _parse_re(self, pattern, buffer):


sujinmkang:
@alexrallen What is the type of the disk? Can you please share how to test and what is the output? do we have any subsequent changes following this change?

alexrallen:
The code I am changing is related to the "generic" disk code. 

To test this you will need to have a switch where SMART is throwing a non-zero return code. Then just attempt to get the ssd health through the "show" command. 

No related changes, this code I am removing was redundant, It does not need to be replaced as the error handling is elsewhere.

sujinmkang:
@alexrallen The previous change actually make the underlying problem come out to the surface. And we noticed the ssdutil throw exceptions when the disk type is hdd. Instead of hiding the problem, it could be better to find the root cause of the exceptions. 
Please refer to https://github.com/Azure/sonic-utilities/pull/1957. I am still working on it.

alexrallen:
@sujinmkang I am not sure what you are asking here, can you clarify?

The change I am proposing here is so that a non-zero return code from `smartctl` does not keep the system from parsing the output which could still be valid and usable. 

The precise case this is an issue with is when the SSD fails a smartctl check, the user should still know about it. 

I am not seeing how your change is related, can you elaborate a bit further?

alexrallen:
The exit code is non-zero and smartctl throws an error that it cannot read specific health metrics from the SSD (I don't have the specific message available) 

Otherwise smartctl gives all of the normal output which can be parsed.

alexrallen:
@sujinmkang I looked over your PR closely. Your PR will only cover instances when the device is of type SSD. My change will cover cases in which it IS an SSD but there is a SMART error with it. 

Both of these changes should be able to be implemented I believe.

sujinmkang (approved):

PR #254, opened 2021-12-21
[component] Fix typo for FW_AUTO_ERR_UNKNOWN

alexrallen:
#### Description
Fixed typo FW_AUTO_ERR_UKNOWN --> FW_AUTO_ERR_UNKNOWN

#### Motivation and Context
This is a typo and in the interest of a readable code base should not remain. There are also very few references to this as of now so changing it now before more references are added is a priority.  

Unit tests passing


liat-grozovik:
@alexrallen is this type used only by Mellanox and thus no need to change other implementations?

liat-grozovik:
@sujinmkang  fyi

alexrallen:
That is correct @liat-grozovik

sujinmkang (approved):

Issue #255, opened 2021-12-22
xcvrd crashes in SFP refactored code.


aravindmani-1:
In the latest master image, xcvrd crashes while trying to get transceiver info for 400G DAC.
  root@sonic:~# show ver
  
  SONiC Software Version: SONiC.master.60307-dirty-20211220.184919
  Distribution: Debian 11.2
  Kernel: 5.10.0-8-2-amd64
  Build commit: 0d327abe9
  Build date: Mon Dec 20 18:56:46 UTC 2021
  Built by: AzDevOps@sonic-build-workers-00101A
  
  Platform: x86_64-dellemc_z9332f_d1508-r0
  HwSKU: DellEMC-Z9332f-O32
  ASIC: broadcom
  ASIC Count: 1
  Serial Number: 1QHXCW2
  Model Number: 04CN21
  Hardware Revision: A00
  Uptime: 09:35:14 up 24 min,  1 user,  load average: 0.68, 0.63, 0.54
  
root@sonic:~# python3
Python 3.9.2 (default, Feb 28 2021, 17:03:44)
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>>
>>> from sonic_platform.chassis import Chassis
>>> c=Chassis()
>>> c.get_sfp(1).get_transceiver_info()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.9/dist-packages/sonic_platform_base/sonic_xcvr/sfp_optoe_base.py", line 24, in get_transceiver_info
    return api.get_transceiver_info() if api is not None else None
  File "/usr/local/lib/python3.9/dist-packages/sonic_platform_base/sonic_xcvr/api/public/cmis.py", line 121, in get_transceiver_info
    admin_info = self.xcvr_eeprom.read(consts.ADMIN_INFO_FIELD)
  File "/usr/local/lib/python3.9/dist-packages/sonic_platform_base/sonic_xcvr/xcvr_eeprom.py", line 30, in read
    return field.decode(raw_data, **decoded_deps)
  File "/usr/local/lib/python3.9/dist-packages/sonic_platform_base/sonic_xcvr/fields/xcvr_field.py", line 252, in decode
    result[field.name] = field.decode(raw_data[offset - start: offset + field.get_size() - start],
[...]

PR #256, opened 2022-01-04
split i2c read across pages

prgeor:
Signed-off-by: Prince George <prgeor@microsoft.com>

#### Description
Fix crash in get_transceiver_info() issue #255 

#### Motivation and Context
Reading register fields across multiple page can result in crash. So, we split the read such that each field grouped are within the same page. This way we can skip reading page >= 01h for passive modules like DAC cable.

Verified on 400G DAC cable and AOC cable


vdahiya12 (dismissed):

aravindmani-1:
@prgeor can you check build failure?.

aravindmani-1 (dismissed):

aravindmani-1 (approved):

vdahiya (approved):

vdahiya12 (approved):

PR #258, opened 2022-01-14
[202012][sonic_sfp]: prefer Config DB port index mapping over platform.json/port_config.ini

nazariig:
#### Description

**Approach:**
1. Always use Config DB to get port mapping when it's available
2. When port mapping is missing in Config DB, use either `platform.json` or `port_config.ini`

On `multiasic` platforms the relevant Config DB instance will be used

#### Motivation and Context

* Always read port mapping from Config DB when it's available
* Fallback to `platform.json` or `port_config.ini` if needed

#### How Has This Been Tested?

* manually tested

#### Additional Information (Optional)
* N/A

PR #264, opened 2022-02-24
[sfputil]Refactoring read_porttab_mappings

dgsudharsan:
#### Motivation and Context

Refactored sfputilhelper API read_porttab_mappings. This API has duplicate logic as done inside get_port_config.
The get_port_config has logic to first look into config_db, fallback to hwsku/platform.json, fallback to port_config.ini. Additionally added another fallback to look into the porttabfile to allow custom porttabfile to be used.

Ran some basic platform tests. Existing UT should also cover these changes

This is to address review comment in https://github.com/Azure/sonic-platform-common/pull/258#discussion_r786128567

### DIFF
--- a/sonic_platform_base/sonic_sfp/sfputilhelper.py
+++ b/sonic_platform_base/sonic_sfp/sfputilhelper.py
@@ -49,139 +49,46 @@ class SfpUtilHelper(object):
         logical = []
         logical_to_physical = {}
         physical_to_logical = {}
-        last_fp_port_index = 0
-        last_portname = ""
-        first = 1
-        port_pos_in_file = 0
-        parse_fmt_port_config_ini = False
-        parse_fmt_platform_json = False
-
-        parse_fmt_port_config_ini = (os.path.basename(porttabfile) == PORT_CONFIG_INI)
-        parse_fmt_platform_json = (os.path.basename(porttabfile) == PLATFORM_JSON)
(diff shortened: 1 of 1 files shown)


Junchao-Mellanox (approved):

dgsudharsan:
@prgeor Can you please review?

prgeor (approved):

mlok-nokia:
Based on the current multiasic platform behaviours, It looks like this PR breaks the "sudo sfputil show eeprom" and "sudo sfputil show presence" on multiasic platform.  On the multiasic platform, it looks like a partial port info is inserted to a "PORT" table in CONFIG_DB in the host database if a port has a SFP and link up.    The Function get_port_config(hwsku, platform) gets "PORT|*" with partial info from the host database.  That info doesn't contain physical info which is required for the sfputilhelper.

prgeor:
> Based on the current multiasic platform behaviours, It looks like this PR breaks the "sudo sfputil show eeprom" and "sudo sfputil show presence" on multiasic platform. On the multiasic platform, it looks like a partial port info is inserted to a "PORT" table in CONFIG_DB in the host database if a port has a SFP and link up. The Function get_port_config(hwsku, platform) gets "PORT|*" with partial info from the host database. That info doesn't contain physical info which is required for the sfputilhelper.

@mlok-nokia please raise a Git issue?

PR #271, opened 2022-04-01
enhance ssd_generic with more error handling to avoid python crash

keboliu:
Signed-off-by: Kebo Liu <kebol@nvidia.com>

#### Description
1. Judge the `self._parse_re` return value, if it's `N/A` then stop further string slice handling to avoid a crash.

2. Update the regular expression pattern for Innodisk SSD health, to handle the case that the output of the health section is different when its lifetime reaches the end. 
#### Motivation and Context
1. Original code doesn't handle the case that `self._parse_re` returns `N/A`, it's assuming that `self._parse_re` will always return a `none N/A` value thus further handling the result w/o judge, which could in a crash.

4.  On Innodisk SSD, when the SSD remaining lifetime reaches the end, the output of the `Health` section will be a number w/o `%`, e.g. `Health: 0.00` instead of `Health: 95.0%` in the normal case,  need to update the regular expression to handle this case.

1. UT test has been added.
2. Tested the change on platforms with different types of SSD.

### DIFF
--- a/sonic_platform_base/sonic_ssd/ssd_generic.py
+++ b/sonic_platform_base/sonic_ssd/ssd_generic.py
@@ -76,18 +76,40 @@ class SsdUtil(SsdBase):
     def parse_generic_ssd_info(self):
         if "nvme" in self.dev:
             self.model = self._parse_re('Model Number:\s*(.+?)\n', self.ssd_info)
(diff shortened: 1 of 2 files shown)


stephenxs (approved):
LGTM

sujinmkang (approved):

PR #280, opened 2022-05-02
[Credo][Ycable] changes for synchronizing executing Telemetry API's when mux toggle is inprogress

vdahiya12:
In this PR, there is a support for adding a mux_toggle_status variable inside the base class for mux_cable.
Using this variable the Derived classes for mux_cable can check this and return in case of a mux_toggle_status is in progress.
From the higher layer this allows ycabled to synchronize the calls and not let mux_cable toggle to go in conjunction with some of the Telemetry calls.

#### Motivation and Context
To get the toggle time to a minimum/ not allow i2c to transactions on the cable to collide with each other

Ran the changes on 7050cx3 arista testbed

### DIFF
--- a/sonic_y_cable/credo/y_cable_credo.py
+++ b/sonic_y_cable/credo/y_cable_credo.py
@@ -7,9 +7,6 @@
 import math
 import time
 import struct
-import threading
-from contextlib import contextmanager
-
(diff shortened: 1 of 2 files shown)


zjswhhh (approved):
LGTM.

PR #288, opened 2022-07-04
Support get_port_or_cage_type

stephenxs:
The API returns the masks of all types of port or cage that can be supported on the port
All the types are defined in sfp_base.SfpBase
Currently, the API is used to check whether a port is an `RJ45` port so we suggest supporting it at least for `RJ45` ports

Signed-off-by: Stephen Sun <stephens@nvidia.com>

### DIFF
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -7,7 +7,7 @@
 
 import sys
 from . import device_base
-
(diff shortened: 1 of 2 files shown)


liat-grozovik (dismissed):

liat-grozovik:
@prgeor could you please help to review?

PR #289, opened 2022-07-06
[xcvr] xcvrd would dead when cable_type is None

chiourung:
#### Description

swsscommon.FieldValuePairs can't accept None and cause xcvrd dead.

Signed-off-by: chiourung_huang <chiourung_huang@edge-core.com>


prgeor:
@chiourung could add more details to what is this crash? Can you attach backtrace of the crash?

chiourung:
> @chiourung could add more details to what is this crash? Can you attach backtrace of the crash?

For xcvrd, it would enter line 401
```
        except NotImplementedError:
            helper_logger.log_error("This functionality is currently not implemented for this platform")
            sys.exit(NOT_IMPLEMENTED_ERROR)
```
There is no backtrace.

PR #300, opened 2022-08-15
Added WDC ssd health

deran1980:
- What I did
Added support for a new SSD appliance from WDC 

- How I did it

Modify ssd_genric.py to extend support of a new SSD 

- How to verify it

Verify on Accton platform (based on Xsight ASIC)  with WDC SSD

- Description for the changelog

- A picture of a cute animal (not mandatory but encouraged)


deran1980:
@prgeor can you please assist in reviewing this PR?

deran1980:
Hi @stephenxs  - can you help with review of this PR?

prgeor (dismissed):

prgeor:
@deran1980 could you add code coverage?

tests/ssd_generic_test.py
```diff
@@ -948,3 +1038,13 @@ def test_transcend_ssd(self, mock_exec):
...
         assert transcend_ssd.get_serial() == "F318410080"
+
+    @mock.patch('sonic_platform_base.sonic_ssd.ssd_generic.SsdUtil._execute_shell')
+    def test_wdc_ssd(self, mock_exec):
+        mock_exec.return_value = output_wdc_vendor
+        wdc_ssd = SsdUtil('/dev/sda')
+        assert wdc_ssd.get_health() == 98
+        assert wdc_ssd.get_model() == 'WDC PC SA530 SDASN8Y1T00'
+        assert wdc_ssd.get_firmware() == "40103000"
+        assert wdc_ssd.get_temperature() == '23 (Min/Max 16/59)'
```
bluecmd, line 1049: Are you sure you want the temperature to be a complex string instead of a scalar value? The other devices seem to use a scalar as a string, so this one stands out.

deran1980: thanks for the feedback.
this is actually from working with real ssd and this is the value I got and passed so this is the way I took it to the test case.
at the end this is a mockup, but I prefer to keep it close to what I actually saw on real HW.

bluecmd: Of course, but what I am suggesting is that instead of extracting the string `23 (Min/Max 16/59)` using the call `self._parse_re('\d*\s\(Min.*', celsius_str)` you instead use a regex group to match only the `23`, e.g. by using `(\d*)\s\(Min.*`.

```
>>> a = '077 059 --- Old_age Always - 23 (Min/Max 16/59)'
>>> import re
>>> re.findall(r'\d*\s\(Min.*', a)
['23 (Min/Max 16/59)']
>>> re.findall(r'(\d*)\s\(Min.*', a)
['23']
```

PR #301, opened 2022-08-16
[ycable] add definitions of some new API's for Y-Cable infrastructure

vdahiya12:
This PR adds the following API's useful for muxcable MCU's debug, these are added as base class for muxcable API's and implemented by vendor
```
    def queue_info(self):

        This API should dump all the meaningful data from the eeprom which can
        help vendor debug the queue info currently relevant to the MCU
        using this API the vendor could check how many txns are currently in the queue etc
        for debugging purposes


    def reset_cause(self):

        This API should return the reset cause for the NIC MCU.
        This should help ascertain whether a reset was caused by soft reboot or
        cable poweroff

   

    def operation_time(self):
        This API should return the time since the cable is powered on from NIC MCU side
        This should be helpful in debugging purposes as to if/when the cable has been powered on

 

    def mem_read(self):

        This API should return the memory contents/as well as pointers/counters for DMA or hardware 
        FIFO's which could be useful for debugging the state of the MCU
```

### DIFF
--- a/sonic_y_cable/y_cable_base.py
+++ b/sonic_y_cable/y_cable_base.py
@@ -1234,7 +1234,7 @@ class YCableBase():
 
         Returns:
             a Bytearray:
-                 the contenets of the memory inside the MCU firmware
+                 the contents of the memory inside the MCU firmware
                  which would help diagnose the cable for proper functioning
         """
 
--- a/sonic_y_cable/y_cable_base.py
+++ b/sonic_y_cable/y_cable_base.py
@@ -1203,7 +1203,7 @@ class YCableBase():
         """
         raise NotImplementedError
 
-    def opertion_time(self):
+    def operation_time(self):
         """
         This API should return the time since the cable is powered on from NIC MCU side
         This should be helpful in debugging purposes as to if/when the cable has been powered on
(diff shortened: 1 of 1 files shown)


sonic_y_cable/y_cable_base.py
```diff
@@ -1167,3 +1167,116 @@ def debug_dump_registers(self, option=None):
...
+             None
+
+        Returns:
+            a float:
+                 the float should represent how much time the mux cable is alive/powered on
+        """
+
+        raise NotImplementedError
+
+    def mem_read(self):
```
xinyulin, line 1222: `def mem_read(self, addr, length):`

Could we add two additional input parameters of the mem_read() function
- addr   : the starting address of the MCU's memory space
- length: length to be read, unit: byte

sonic_y_cable/y_cable_base.py
```diff
@@ -1167,3 +1167,116 @@ def debug_dump_registers(self, option=None):
...
+        """
+        This API should return the memory contents of the cable which would be useful in debug for the
+        y-cable
+
+        Args:
+             None
+
+        Returns:
+            a Dictionary:
```
xinyulin, line 1231: The MCU will return a number of bytes data, could we change the return type to bytearray?

prgeor (approved):

PR #304, opened 2022-08-23
Add warning/critical thresholds for PSU power

stephenxs:
#### Description

Two new platform APIs for PSU
- get_psu_power_warning_suppress_threshold
- get_psu_power_critical_threshold

Signed-off-by: Stephen Sun <stephens@nvidia.com>

### DIFF
--- a/sonic_platform_base/psu_base.py
+++ b/sonic_platform_base/psu_base.py
@@ -223,7 +223,7 @@ class PsuBase(device_base.DeviceBase):
         """
         raise NotImplementedError
 
-    def get_psu_power_threshold(self):
+    def get_psu_power_warning_threshold(self):
         """
         Retrieve the warning threshold of the power on this PSU
         The value can be volatile, so the caller should call the API each time it is used.
--- a/sonic_platform_base/psu_base.py
+++ b/sonic_platform_base/psu_base.py
@@ -223,13 +223,13 @@ class PsuBase(device_base.DeviceBase):
         """
         raise NotImplementedError
 
-    def get_psu_power_warning_threshold(self):
+    def get_psu_power_warning_suppress_threshold(self):
         """
-        Retrieve the warning threshold of the power on this PSU
+        Retrieve the warning suppress threshold of the power on this PSU
         The value can be volatile, so the caller should call the API each time it is used.
 
         Returns:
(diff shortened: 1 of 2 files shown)


keboliu (approved):

sonic_platform_base/psu_base.py
```diff
@@ -223,6 +223,26 @@ def get_maximum_supplied_power(self):
         """
         raise NotImplementedError
 
+    def get_psu_power_warning_threshold(self):
```
prgeor, line 226: this naming is confusing as no warning is raised for this threshold. can we name it as "warning-suppress" level?

sonic_platform_base/psu_base.py
```diff
@@ -223,6 +223,26 @@ def get_maximum_supplied_power(self):
...
+        """
+        Retrieve the warning threshold of the power on this PSU
```
prgeor, line 228: Need one API so that user can override the warning suppress level

sonic_platform_base/psu_base.py
stephenxs: Changed to `get_psu_power_warning_suppress_threshold`

sonic_platform_base/psu_base.py
stephenxs: As discussed offline, we do not need a platform API to do it because platform API can not store the information.

sonic_platform_base/psu_base.py
prgeor: @stephenxs pr description still has warning threshold.

stephenxs: @prgeor  you were probably checking an out-of-data commit. Use link https://github.com/sonic-net/sonic-platform-common/pull/304/files to open the latest one

keboliu (approved):

sonic_platform_base/psu_base.py
```diff
@@ -223,6 +223,26 @@ def get_maximum_supplied_power(self):
         """
         raise NotImplementedError
 
+    def get_psu_power_warning_suppress_threshold(self):
```
prgeor, line 226: Please update the description with this function name

prgeor (approved):

PR #306, opened 2022-08-31
Update the return of update_firmware for the image not exist case

sujinmkang:
#### Description

issue : https://github.com/sonic-net/sonic-mgmt/issues/4564
#### Motivation and Context

When the image file path doesn't exit, returns False instead of throwing error.

### DIFF
--- a/sonic_platform_base/component_base.py
+++ b/sonic_platform_base/component_base.py
@@ -109,7 +109,7 @@ class ComponentBase(object):
             image_path: A string, path to firmware image
 
         Returns:
-            False if image not found.
+            A boolean, False only if image_path was is not valid
 
         Raises:
             RuntimeError: update failed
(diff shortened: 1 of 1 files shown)


sonic_platform_base/component_base.py
```diff
@@ -108,6 +108,9 @@ def update_firmware(self, image_path):
         Args:
             image_path: A string, path to firmware image
 
+        Returns:
+            False if image not found.
+
```
prgeor, line 113: can we have return 'False' for any failure. see the definition of install_firmware()?

```
  Returns:
            A boolean, True if install was successful, False if not
```

sonic_platform_base/component_base.py
```diff
@@ -108,6 +108,10 @@ def update_firmware(self, image_path):
         Args:
             image_path: A string, path to firmware image
 
+        Returns:
+            Boolean False if image_path doesn't exist instead of throwing an exception error
+            None when the update is successful
```
prgeor, line 113: how is 'None' returned if update_firmware() results in reboot?

prgeor (approved):

PR #309, opened 2022-09-07
Reduce eeprom calls from xcvr_apis by caching page 00h

vivekrnv:
Signed-off-by: Vivek Reddy Karri <vkarri@nvidia.com>

#### Description

Introduce a simple cache mechanism to cache page 00h, since that is know to be present for all PI types and also most of the individual calls were made to this page. 

This doesn't just affect the CLI since daemons likes xcvrd use SfpBase class and they invoke methods like get_transceiver_info & get_transceiver_bulk_status & get_transceiver_threshold_info and populate the info in STATE_DB. 

Thus the idea is to cache entire page once before invoking these api's and clear after fetching the result. Thus during the next execution, it'll read updated data.

#### Motivation and Context

`sfputil show eeprom` is very slow when asked about a SFP of type CMIS. In comparison, other sfp types are taking about 0.6 sec. 

```
root@r-leopard-01:/home/admin# time sfputil show eeprom -p Ethernet0
Ethernet0: SFP EEPROM detected
................
        Connector: No separable connector
        Encoding: N/A
        Host Lane Count: 8
        Identifier: QSFP-DD Double Density 8X Pluggable Transceiver
        Media Interface Technology: Copper cable unequalized
        Media Lane Assignment Options: N/A
        Media Lane Count: 0
        Nominal Bit Rate(100Mbs): 0
        Specification compliance: passive_copper_media_interface
................

real    0m4.012s
user    0m1.122s
sys     0m0.631s

root@r-leopard-01:/home/admin# python3 -m cProfile -s tottime /usr/local/bin/sfputil show eeprom -p Ethernet0  | grep eeprom
root@r-leopard-01:/home/admin# cat pre_opt.txt
       31    0.002    0.000    4.387    0.142 sfp.py:350(_read_eeprom_specific_bytes)
       29    0.001    0.000    4.059    0.140 xcvr_eeprom.py:15(read)
       31    0.000    0.000    4.387    0.142 sfp.py:374(read_eeprom)
        1    0.000    0.000    4.411    4.411 main.py:611(eeprom)
       29    0.000    0.000    0.000    0.000 xcvr_eeprom.py:29(<dictcomp>)
        1    0.000    0.000    0.000    0.000 eeprom_dts.py:3(<module>)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:10(__init__)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:1(<module>)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:9(XcvrEeprom)
```

Most of the time is spent in _read_eeprom_specific_bytes call and it since it involves calling subprocess.call, calling it 30 times has its toll.

- Unit Tests
- Tested on the live switch .
```
root@r-leopard-01:/home/admin# time sfputil show eeprom -p Ethernet0
...............
real    0m1.063s
user    0m0.333s
sys     0m0.209s

root@r-leopard-01:/home/admin# python3 -m cProfile -s tottime /usr/local/bin/sfputil show eeprom -p Ethernet0 | grep eeprom
        6    0.001    0.000    0.916    0.153 sfp.py:350(_read_eeprom_specific_bytes)
       28    0.000    0.000    0.260    0.009 xcvr_eeprom.py:28(read)
        1    0.000    0.000    0.940    0.940 main.py:611(eeprom)
       28    0.000    0.000    0.255    0.009 xcvr_eeprom.py:16(get_data)
        6    0.000    0.000    0.916    0.153 sfp.py:374(read_eeprom)
        2    0.000    0.000    0.396    0.198 xcvr_eeprom.py:108(enable_cache)
       28    0.000    0.000    0.000    0.000 xcvr_eeprom.py:42(<dictcomp>)
        1    0.000    0.000    0.000    0.000 eeprom_dts.py:3(<module>)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:10(__init__)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:9(XcvrEeprom)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:1(<module>)
        2    0.000    0.000    0.000    0.000 xcvr_eeprom.py:119(disable_cache)
```

```
root@r-leopard-01:/home/admin# python3 -m cProfile -s tottime /usr/local/bin/sfputil show eeprom -p Ethernet0 -d | grep eeprom
        8    0.001    0.000    1.214    0.152 sfp.py:350(_read_eeprom_specific_bytes)
       73    0.000    0.000    0.240    0.003 xcvr_eeprom.py:28(read)
        1    0.000    0.000    1.241    1.241 main.py:611(eeprom)
        4    0.000    0.000    0.705    0.176 xcvr_eeprom.py:108(enable_cache)
       73    0.000    0.000    0.235    0.003 xcvr_eeprom.py:16(get_data)
        8    0.000    0.000    1.214    0.152 sfp.py:374(read_eeprom)
        1    0.000    0.000    0.000    0.000 eeprom_dts.py:3(<module>)
       73    0.000    0.000    0.000    0.000 xcvr_eeprom.py:42(<dictcomp>)
        4    0.000    0.000    0.000    0.000 xcvr_eeprom.py:119(disable_cache)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:9(XcvrEeprom)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:10(__init__)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:1(<module>)
```

Even the other module types can benefit from the caching mechanism
```
root@r-leopard-01:/home/admin# python3 -m cProfile -s tottime /usr/local/bin/sfputil show eeprom -p Ethernet240 -d  | grep eeprom
        5    0.001    0.000    0.844    0.169 sfp.py:350(_read_eeprom_specific_bytes)
       14    0.000    0.000    0.001    0.000 xcvr_eeprom.py:28(read)
        1    0.000    0.000    0.859    0.859 main.py:611(eeprom)
        3    0.000    0.000    0.573    0.191 xcvr_eeprom.py:108(refresh_cache)
        5    0.000    0.000    0.844    0.169 sfp.py:374(read_eeprom)
       14    0.000    0.000    0.000    0.000 xcvr_eeprom.py:16(get_data)
       14    0.000    0.000    0.000    0.000 xcvr_eeprom.py:42(<dictcomp>)
        1    0.000    0.000    0.000    0.000 eeprom_dts.py:3(<module>)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:10(__init__)
        3    0.000    0.000    0.000    0.000 xcvr_eeprom.py:119(clear_cache)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:1(<module>)
        1    0.000    0.000    0.000    0.000 xcvr_eeprom.py:9(XcvrEeprom)
```


vivekrnv:
@keboliu, Can you review?

vivekrnv:
Request for 202205

keboliu (approved):

prgeor:
@vivekrnv are we reading Page0 multiple times inside get_transceiver_info() [https://github.com/sonic-net/sonic-utilities/blob/master/sfputil/main.py#L663(https://github.com/sonic-net/sonic-utilities/blob/master/sfputil/main.py#L663)?

vivekrnv:
> @vivekrnv are we reading Page0 multiple times inside get_transceiver_info() [https://github.com/sonic-net/sonic-utilities/blob/master/sfputil/main.py#L663(https://github.com/sonic-net/sonic-utilities/blob/master/sfputil/main.py#L663)?

Yes, we are. not the entire page every time though.

prgeor:
@vivekrnv the fix assumes that the static fields are always in a particular page. This may not be necessarily true. I think we should decide to cache based upon the field instead of depending upon the page number. From the CMIS spec v5.1, "static" is attributed to a field not to a page. 

image

For eg, see page 15h, not all fields are static:-

image

sonic_platform_base/sonic_xcvr/xcvr_eeprom.py
```diff
@@ -91,3 +104,24 @@ def write_raw(self, offset, size, bytearray_data):
...
+   
+   def refresh_cache(self, **kwargs):
+      """
+      When invoked, it caches the lower and upper pages of 00h i.e offset 0 to 256
+      Any read request made strictly b/w these limits will be read from the cache until cleared
+
+      Returns: 
+         None
+      """
+      # TODO: Use a better mechanism and allow caching for higher pages and more complicated scenarios
```
prgeor, line 116: I think we should consider to have  solution which is field specific rather than offset or page specific. The 'if' checks here will increase if we have static and non-static fields in same page.

vivekrnv:
> @vivekrnv the fix assumes that the static fields are always in a particular page. This may not be necessarily true. I think we should decide to cache based upon the field instead of depending upon the page number. From the CMIS spec v5.1, "static" is attributed to a field not to a page.
> 
> image
> 
> For eg, see page 15h, not all fields are static:-
> 
> image

I don’t understand your concern. Just to confirm, I’m not caching for the entire duration of the program execution. I’m just reading all the data in page 0 upfront and then calling get_transciever_info. I.e. all the y num of calls made will hit the cached page. Once this method returns, I’m clearing the cache. If you think it’a still a problem, can you point towards an example of where it might be problematic

prgeor:
> > @vivekrnv the fix assumes that the static fields are always in a particular page. This may not be necessarily true. I think we should decide to cache based upon the field instead of depending upon the page number. From the CMIS spec v5.1, "static" is attributed to a field not to a page.
> > image
> > For eg, see page 15h, not all fields are static:-
> > image
> 
> I don’t understand your concern. Just to confirm, I’m not caching for the entire duration of the program execution. I’m just reading all the data in page 0 upfront and then calling get_transciever_info. I.e. all the y num of calls made will hit the cached page. Once this method returns, I’m clearing the cache. If you think it’a still a problem, can you point towards an example of where it might be problematic

This proposal will work fine for Page 0 and Page 1. How to handle the case where only few fields in a page are static?

vivekrnv:
> 



> > > @vivekrnv the fix assumes that the static fields are always in a particular page. This may not be necessarily true. I think we should decide to cache based upon the field instead of depending upon the page number. From the CMIS spec v5.1, "static" is attributed to a field not to a page.
> > > image
> > > For eg, see page 15h, not all fields are static:-
> > > image
> > 
> > 
> > I don’t understand your concern. Just to confirm, I’m not caching for the entire duration of the program execution. I’m just reading all the data in page 0 upfront and then calling get_transciever_info. I.e. all the y num of calls made will hit the cached page. Once this method returns, I’m clearing the cache. If you think it’a still a problem, can you point towards an example of where it might be problematic
> 
> This proposal will work fine for Page 0 and Page 1. How to handle the case where only few fields in a page are static?

This fix doesn't bother about fields being static. Since this just reads page upfront once (when the get_* api's are called) and parse all the info (as requested by xcvr_api's) from it and clears the cache in the end. 

So, lets says even if we extend for pages with non-static content, all we are doing is reading upfront and clearing it after the api is called. The updated content will certainly be reflected in the STATE_DB (considering the user is xcvrd)

In any case, this solution is just a beginning, the intent here is to reduce calls to read_eeprom. Extending this to other pages will require further analysis and considerations. And extending the cache to consider fields being static and reading them once per program execution is a good enhancement but that's not in the scope of the PR.

PR #316, opened 2022-10-21
 CmisApi::get_application_advertisement catch AttributeError as well 

stephenxs:
#### Description

Catch both `TypeError` and `AttributeError` in `CmisApi::get_application_advertisement` because an AttributeError will be thrown when updating a dict with `None`.

Manually test
Added unit test


keboliu (approved):

liat-grozovik:
@stephenxs thanks for the fix. do we also need it on 202205?
@prgeor can you please provide your feedback/signoff?

stephenxs:
> @stephenxs thanks for the fix. do we also need it on 202205? @prgeor can you please provide your feedback/signoff?

Yes. Label added.

liat-grozovik (approved):

PR #318, opened 2022-10-28
Fix issue: copper cable should not display DOM information

Junchao-Mellanox:
#### Description
Fix issue: copper cable should not display DOM information

#### Motivation and Context
For sff8636 and sff8436 cables, there is no DOM information. Need not read DOM data if cable is copper.

Manual test


liat-grozovik:
@Junchao-Mellanox  could you please check the coverage report and see what tests can be added to cover touched functions?

keboliu (approved):

prgeor (approved):

PR #345, opened 2023-02-14
[Credo][Ycable] update health_check() to enhance the accuracy 

xinyulin:
#### Description
The fix will check UART retry/abort counts and fail it if those counters are over-limit.
The API, health_check() is used for RMA diagnostics.

#### Motivation and Context
To enhance health_check() accuracy 

Tested on Arista-7050CX3

### DIFF
--- a/sonic_y_cable/credo/y_cable_credo.py
+++ b/sonic_y_cable/credo/y_cable_credo.py
@@ -3674,8 +3674,22 @@ class YCable(YCableBase):
                 return YCable.CABLE_UNHEALTHY
 
             uart_stat = self.get_uart_stat()
-            if api_ver >= 0x18 and uart_stat['Local']['UART2']['RxErrorCnt'] > 100:
-                self.log_error("check cable health fail: uart rx error count overlimit:%d" % (uart_stat['local']['UART2']['RxErrorCnt']))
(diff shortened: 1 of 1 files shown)


sonic_y_cable/credo/y_cable_credo.py
```diff
@@ -3674,9 +3672,29 @@ def health_check(self):
                 return YCable.CABLE_UNHEALTHY
 
             uart_stat = self.get_uart_stat()
-            if api_ver >= 0x18 and uart_stat['Local']['UART2']['RxErrorCnt'] > 100:
-                self.log_error("check cable health fail: uart rx error count overlimit:%d" % (uart_stat['local']['UART2']['RxErrorCnt']))
+            if api_ver >= 0x18:
+                if uart_stat['Local']['UART2']['RxErrorCnt'] > 100:
```
vdahiya12, line 3676: could we check the uart_stat dictionary would have the keys ? and if not what to do ?

xinyulin: My original concept is to use api_ver to avoid to hit KeyError (key is not exist in the dict). but, you are right, it is better to check if the key is exist or not. Please submit one commit to handle the exception

PR #348, opened 2023-02-27
Check SONiC dependencies before installation.

liushilongbuaa:

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,18 @@
 from setuptools import setup
+import pkg_resources
+
+# sonic module dependencies.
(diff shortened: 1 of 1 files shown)


prgeor:
@liushilongbuaa can you elaborate a bit more why installation from PyPi is security concern?

liushilongbuaa:
> @liushilongbuaa can you elaborate a bit more why installation from PyPi is security concern?

Because SONiC packages are not registered in PyPi. When we run setuptools, it will try to download from PyPi.
If someone registered packages using the same name. Unsafe code is executed.

qiluo-msft (approved):

PR #349, opened 2023-02-27
Fix pipeline apt-get purge error.

liushilongbuaa:
#### Description

PR pipeline failed for apt-get purge return code change.
We need to ignore return code when purge packages.
Download necessary packages to accelerate download step.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -34,11 +34,17 @@ steps:
     runVersion: 'latestFromBranch'
     runBranch: 'refs/heads/$(BRANCH_NAME)'
     patterns: |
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -33,11 +33,23 @@ steps:
     artifact: sonic-buildimage.vs
     runVersion: 'latestFromBranch'
     runBranch: 'refs/heads/$(BRANCH_NAME)'
+    patterns: |
+      **/swsssdk-2.0.1-py3-none-any.whl
```
xumia, line 37: Change to common pattern, for example: *.deb, *.whl

xumia (approved):

PR #353, opened 2023-03-14
Add set/get_lpmode APIs for sff8436 and sff8636 optics

longhuan-cisco:
Add set/get_lpmode APIs for sff8436 and sff8636 optics

#### Description
Add set/get_lpmode APIs implementation in sff8436.py and sff8636.py

#### Motivation and Context
According to 8436 and 8636 spec, lpmode can be manipulated via the combination of the Power_over-ride and Power_set software control bits (Address A0h, byte 93 bits 0,1),  as sfp refactor came in, such logic needs to be in sonic-platform-common, thus add set/get_lpmode implementation following the spec.  
This is common routine based on standard spec. Vendor/platform can decide to whether use it or combine it with platform specific actions. 
CMIS/C-CMIS already has the support, thus no need to add.

1. Verified set/get APIs on 100G optics. checked eeprom dump/tx_power/rx_power after setting lpmode. (when optics is in lpmode, oper_status is down, and there's no tx/rx_power, but eeprom is still accessible.) 
3. Verified set/get APIs on 40G optics which is power class 1 (max power 1.5W), checked eeprom dump/tx_power/rx_power after setting lpmode.  In this case, set API will return False. Even if, we toggle the control bits to set it to lpmode, optics will still be in oper_status UP, because as mentioned by sff8436 spec, ```If the Extended Identifier bits (Page 00h, byte 129 bits 6-7) indicate that its power consumption is less than 1.5W then the module shall be fully functional independent of whether it is in low power or high power mode.```

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/sff8436.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/sff8436.py
@@ -308,3 +308,38 @@ class Sff8436Api(XcvrApi):
 
     def get_power_override_support(self):
         return not self.is_copper()
+
+    def set_lpmode(self, lpmode):
+        '''
+        This function sets the module to low power state.
+
+        Args:
+            lpmode (bool): False means "set to high power", True means "set to low power"
+
+        Returns:
+            bool: True if the provision succeeds, False if it fails
+        '''
+        if not self.get_lpmode_support() or not self.get_power_override_support():
+            return False
+
+        if lpmode:
+            return self.set_power_override(True, True)
+        else:
+            return self.set_power_override(True, False)
+
+    def get_lpmode(self):
+        '''
+        Retrieves low power module status
+
+        Returns:
+            bool: True if module in low power else returns False.
+        '''
+        if not self.get_lpmode_support() or not self.get_power_override_support():
+            return False
+
+        power_set = self.get_power_set()
+        power_override = self.get_power_override()
+
+        # Since optics come up by default set to high power, in this case,
+        # power_override not being set, function will return high power mode.
+        return power_set and power_override
--- a/sonic_platform_base/sonic_xcvr/api/public/sff8436.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/sff8436.py
@@ -340,6 +340,6 @@ class Sff8436Api(XcvrApi):
         power_set = self.get_power_set()
         power_override = self.get_power_override()
 
-        # Since optics come up by default set to high power, in this case,
+        # Since typically optics come up by default set to high power, in this case,
         # power_override not being set, function will return high power mode.
         return power_set and power_override
--- a/sonic_platform_base/sonic_xcvr/api/public/sff8436.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/sff8436.py
@@ -322,10 +322,7 @@ class Sff8436Api(XcvrApi):
         if not self.get_lpmode_support() or not self.get_power_override_support():
             return False
 
-        if lpmode:
-            return self.set_power_override(True, True)
-        else:
-            return self.set_power_override(True, False)
+        return self.set_power_override(True, lpmode)
 
     def get_lpmode(self):
         '''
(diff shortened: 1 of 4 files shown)


longhuan-cisco:
@prgeor @mihirpat1 @shyam77git @jaganbal-a could you please review?

sonic_platform_base/sonic_xcvr/api/public/sff8436.py
```diff
@@ -308,3 +308,35 @@ def get_lpmode_support(self):
 
     def get_power_override_support(self):
         return not self.is_copper()
+
+    def set_lpmode(self, lpmode):
+        '''
+        This function sets the module to low power state.
```
mihirpat1, line 314: Can we change "low power state" to LPMode?

longhuan-cisco: changed to LPMode

sonic_platform_base/sonic_xcvr/api/public/sff8436.py
```diff
@@ -308,3 +308,35 @@ def get_lpmode_support(self):
...
+            bool: True if the provision succeeds, False if it fails
+        '''
+        if not self.get_lpmode_support() or not self.get_power_override_support():
+            return False
+
+        return self.set_power_override(True, lpmode)
+
+    def get_lpmode(self):
+        '''
+        Retrieves low power module status
```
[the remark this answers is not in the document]
longhuan-cisco: changed to "mode"

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
```diff
@@ -360,3 +360,35 @@ def get_lpmode_support(self):
 
     def get_power_override_support(self):
         return not self.is_copper()
+
+    def set_lpmode(self, lpmode):
+        '''
+        This function sets LPMode for the module.
+
+        Args:
+            lpmode (bool): False means "set to high power", True means "set to low power"
```
keboliu, line 369: I want to suggest  using "False means LPMode Off, True means LPMode On", the reason is like below:
if we check the output of  CLI "sfputil show lpmode", we can see it is using 'on/off' to indicate the status, it will make it consistent across layers.

> ```
> admin@sonic:# sfputil show lpmode 
> Port         Low-power Mode
> -----------  ----------------
> Ethernet180  Off
> Ethernet216  On
> ```

longhuan-cisco: good suggestion. update it.

mihirpat1 (approved):

jaganbal-a (approved):

prgeor (approved):

sonic_platform_base/sonic_xcvr/api/public/sff8436.py
```diff
@@ -308,3 +308,35 @@ def get_lpmode_support(self):
...
+        '''
+        if not self.get_lpmode_support() or not self.get_power_override_support():
+            return False
+
+        power_set = self.get_power_set()
+        power_override = self.get_power_override()
+
+        # Since typically optics come up by default set to high power, in this case,
+        # power_override not being set, function will return high power mode.
+        return power_set and power_override
```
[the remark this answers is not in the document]
longhuan-cisco: According to the sff-8636/sff-8436 spec, the LPM can be driven by LPMode pin or SW controlled bits.
The standard api get_lpmode/set_lpmode() is supposed to perform SW controlled way of setting and getting the lp mode status.
If a platform uses LPMode pin for driving LP mode, then it is supposed to over-ride the standard definition with their platform definition.

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
```diff
@@ -360,3 +360,35 @@ def get_lpmode_support(self):
...
+
+        Returns:
+            bool: True if the provision succeeds, False if it fails
+        '''
+        if not self.get_lpmode_support() or not self.get_power_override_support():
+            return False
+
+        return self.set_power_override(True, lpmode)
+
+    def get_lpmode(self):
```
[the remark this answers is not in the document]
longhuan-cisco: the truth table is followed in the case of SW controlled way of setting and getting the lp mode status and If a platform uses LPMode pin for driving LP mode, then it is supposed to over-ride the standard definition with their platform definition.

PR #355, opened 2023-03-31
[Backport] [202211] Fix issue: should always check return value of a function if the function may return None

Junchao-Mellanox:
Backport https://github.com/sonic-net/sonic-platform-common/pull/355 to 202211

#### Description

When a function may return None, caller should always check the return value before using it. This PR is to fix such issues in sonic-xcvr.

#### Motivation and Context

When a function may return None, caller should always check the return value before using it. This PR is to fix such issues in sonic-xcvr.

Created unit test cases to avoid such thing happening


liat-grozovik:
@prgeor could you please help to review and merge?

prgeor (approved):

keboliu (approved):

PR #360, opened 2023-04-06
[sonic_xcvr] Prevent VDM dictionary related KeyError when a transceiver module is pulled while a bulk get method is interrogating said module

snider-nokia:
#### Description
This change stops an incomplete VDM dictionary from causing a bulk get method KeyError when a transceiver module is pulled while the bulk method is interrogating the module. 

#### Motivation and Context
When a transceiver module is pulled there is lag prior to the time when the SfpStateUpdateTask actually gets notification of module removal.  During this lag time the module can be in the process of being accessed by another thread/task (for example, DomInfoUpdateTask using a bulk get method like get_transceiver_status()).  In this case the information returned to optoe driver is junk and the VDM dictionary is not created properly (not all keys are populated).  The bulk get method, which currently expects the VDM dictionary to contain all keys, then gets a KeyError exception and the protagonist thread/task blows up.

PMON operates normally and Xcvrd should not fail due to VDM related KeyError when transceiver modules are removed.


snider-nokia:
@prgeor, Can you review?

prgeor (approved):

PR #374, opened 2023-05-18
Convert the tx/rx power unit to the dBm unit in get_…

chiourung:
…transceiver_bulk_status().

#### Description

The tx/rx power unit in the eeprom is mW, and the tx/rx power unit for the API is dBm. It should be converted to the dBm in get_transceiver_bulk_status().


chiourung:
The solution have some problem

PR #376, opened 2023-05-26
Additional prevention of potential fatal exception caused by VDM dictionary missing fields when a transceiver module has just been pulled

snider-nokia:
#### Description
This change stops an incomplete VDM dictionary from causing a bulk get method KeyError when a transceiver module is pulled while the module is being interrogated.   Additional VDM dictionary fields have been addressed with this commit.

#### Motivation and Context
When a transceiver module is pulled there is lag prior to the time when the SfpStateUpdateTask actually gets notification of module removal. During this lag time the module can be in the process of being accessed by another thread/task (for example, DomInfoUpdateTask using a bulk get method like get_transceiver_status()). In this case the information returned to optoe driver is junk and the VDM dictionary is not created properly (not all keys are populated). The bulk get method, which currently expects the VDM dictionary to contain all keys, then gets a KeyError exception and the protagonist thread/task blows up.

PMON operates normally and Xcvrd should not fail due to VDM related KeyError when transceiver modules are removed.

#### Additional Information
The previous commit, https://github.com/sonic-net/sonic-platform-common/pull/360, related to this issue missed some VDM dictionary fields.  This commit addresses those previously missed fields.


prgeor:
@snider-nokia there seems to be some work done by @mihirpat1 which address these kind of issues when Xcvrd is busy reading the module as part of various DOM/VDM polling which is more generic and not selective at different places all around the Xcvrd code. May be we should wait for that PR

snider-nokia:
No problem if so, Prince.  I just know we were hitting these problems sporadically when testing module removal heavily this past week.

prgeor (approved):

prgeor:
@yxieca please cherry pick to 202205

rlhui:
@prgeor - please help create a msft ado for this, thanks.

mihirpat1:
> #### Description
> This change stops an incomplete VDM dictionary from causing a bulk get method KeyError when a transceiver module is pulled while the module is being interrogated. Additional VDM dictionary fields have been addressed with this commit.
> 
> #### Motivation and Context
> When a transceiver module is pulled there is lag prior to the time when the SfpStateUpdateTask actually gets notification of module removal. During this lag time the module can be in the process of being accessed by another thread/task (for example, DomInfoUpdateTask using a bulk get method like get_transceiver_status()). In this case the information returned to optoe driver is junk and the VDM dictionary is not created properly (not all keys are populated). The bulk get method, which currently expects the VDM dictionary to contain all keys, then gets a KeyError exception and the protagonist thread/task blows up.
> 
> #### How Has This Been Tested?
> PMON operates normally and Xcvrd should not fail due to VDM related KeyError when transceiver modules are removed.
> 
> #### Additional Information
> The previous commit, #360, related to this issue missed some VDM dictionary fields. This commit addresses those previously missed fields.

@yi

> @prgeor - please help create a msft ado for this, thanks.

@yxieca @rlhui - The corresponding ADO is 24473731

PR #389, opened 2023-07-26
Comment out tx power validation check and program the passed value 

abdosi:
What I did:

Comment out tx power validation check and program the passed value  to eeprom write function

MSFT ADO: 17893499

Why I did:

For some of our 400G ZR optics calibrated value of tx power is not in min and max range of advertised value which causes xcvrd to crash and then links do not come up . This fix is short term solution to skip the validation check.

How I verify:
After this change not seeing xcvrd crash.

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
@@ -143,10 +143,6 @@ class CCmisApi(CmisApi):
         Return True if the provision succeeds, False if it fails
         '''
         min_prog_tx_output_power, max_prog_tx_output_power = self.get_supported_power_config()
-        #Comment out validation for now.
-        #if tx_power > max_prog_tx_output_power or tx_power < min_prog_tx_output_power:
-        #    raise ValueError('Provisioned TX power out of range. Max: %.1f; Min: %.1f dBm.' 
-        #                     %(max_prog_tx_output_power, min_prog_tx_output_power))
         status = self.xcvr_eeprom.write(consts.TX_CONFIG_POWER, tx_power)
         time.sleep(1)
         return status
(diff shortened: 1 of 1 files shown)


sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
```diff
@@ -143,9 +143,10 @@ def set_tx_power(self, tx_power):
         Return True if the provision succeeds, False if it fails
         '''
         min_prog_tx_output_power, max_prog_tx_output_power = self.get_supported_power_config()
-        if tx_power > max_prog_tx_output_power or tx_power < min_prog_tx_output_power:
-            raise ValueError('Provisioned TX power out of range. Max: %.1f; Min: %.1f dBm.' 
-                             %(max_prog_tx_output_power, min_prog_tx_output_power))
+        #Comment out validation for now.
```
prgeor, line 146: @abdosi can you remove the code itself?

abdosi: @prgeor updated.

gechiang (approved):

gechiang:
MSFT ADO: 24669442

gechiang:
@yxieca , @StormLiangMS , can you help cherry-pick this bug fix to the needed branches accordingly?
Thanks!

gechiang:
@yxieca, please let us know if this can be cherry-picked cleanly into 202205.  If not, we will have to raise a 202205 specific PR for it...
If you do not feel this bug fix should be part of 202205, let us know as well so we can pick into the MSFT repo if it will not be picked into 202205
Thanks!

gechiang:
@yxieca , can you also approve this for 202211 so we can mark the ADO as complete. Thanks!

PR #392, opened 2023-08-09
Fix NumRegField byte(s) read

prgeor:
#### Description
NumRegField read worked fine when the fields were single bit. Now that we have RegBitsFields with multiple bits we need to adjust the mask accordingly.

#### Motivation and Context
NumRegFields having RegBitsFields did not read the byte correctly. This is now fixed with this change.

Unit test passed


prgeor:
@andywongarista can you review this fix

sonic_platform_base/sonic_xcvr/fields/xcvr_field.py
```diff
@@ -86,6 +87,7 @@ def __init__(self, name, bitpos, offset=None, **kwargs):
         super(RegBitField, self).__init__(name, offset, **kwargs)
         assert bitpos < 64
         self.bitpos = bitpos
+        self.bitmask = 1 << self.bitpos
```
AnoopKamath, line 90: Are we using bitmask in RegBitField ?

prgeor: @andywongarista not directly, but when RegBitField is grouped inside a NumberRegField or CodeRegField

AnoopKamath: @prgeor ok. Thanks

AnoopKamath (approved):

yxieca (approved):

PR #409, opened 2023-10-24
Fix CCmisApi identifier overrides ycable identifier in syslog.

xincunli-sonic:
**Summary**
This PR fixes the CCmisApi overrides ycable as SYSLOG identifier when CCmisApi became the last caller of sonic-py-common/logger.py's syslog module.

**Type of change**
 - [x] Bug fix
 - [ ] Testbed and Framework(new/improvement)
 - [ ]  Test case(new/improvement)
 
**Approach**
Use logging module instead of common logger to avoid overriding.

**What is the motivation for this PR?**
SysLog shows `NOTICE pmon#CCmisApi: ` instead of `NOTICE pmon#ycable: `, which cause debug confusion and loganalysis wrong identified the component.

**How did you do it?**
Removed the helper_logger from CCmisApi instead using logging directly.

**How did you verify/test it?**
Put the change into lab device, then inspect the syslog after restart ycabled daemon process.

**Any platform specific information?**
This is common issue when ycabled started.

Signed-off-by: Xincun Li [xincun.li@microsoft.com](mailto:xincun.li@microsoft.com)

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
@@ -6,17 +6,19 @@
 from sonic_py_common import logger
 from ...fields import consts
 from .cmis import CmisApi
+import logging
 import time
+
+
 BYTELENGTH = 8
 VDM_FREEZE = 128
 VDM_UNFREEZE = 0
-SYSLOG_IDENTIFIER = "CCmisApi"
 
-helper_logger = logger.Logger(SYSLOG_IDENTIFIER)
 
 class CCmisApi(CmisApi):
     def __init__(self, xcvr_eeprom):
         super(CCmisApi, self).__init__(xcvr_eeprom)
+        logging.basicConfig(level=logging.DEBUG)
 
     def get_freq_grid(self):
         '''
@@ -814,7 +816,7 @@ class CCmisApi(CmisApi):
             trans_status['rxsigpowerhighwarning_flag'] = self.vdm_dict['Rx Signal Power [dBm]'][1][7]
             trans_status['rxsigpowerlowwarning_flag'] = self.vdm_dict['Rx Signal Power [dBm]'][1][8]
         except KeyError:
-            helper_logger.log_debug('fields not present in VDM')
+            logging.debug('fields not present in VDM')
         return trans_status
 
     def get_transceiver_pm(self):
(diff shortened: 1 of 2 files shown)


sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
```diff
@@ -6,17 +6,19 @@
 from sonic_py_common import logger
 from ...fields import consts
 from .cmis import CmisApi
+import logging
```
vdahiya12, line 9: recommend not having logging here, as it logs on the console, in this case its not going to be caught anywhere inside pmon. 
For this change, you can ignore the exception as pass, Let's try going with the route of passing a logger instance to the constructor, or else
try fixing the sonic_py_common lib function to take the first invocation itself as the final identifier. 
Both ways should work

vdahiya12 (changes requested):

sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
xincunli-sonic: Per chatting with @vdahiya12 offline, since many other places are using logging, let us keep it in use till we decide the better way to handle such log identifier overriding.

vdahiya12 (approved):

vdahiya12 (approved):

sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
```diff
@@ -814,7 +815,7 @@ def get_transceiver_status(self):
             trans_status['rxsigpowerhighwarning_flag'] = self.vdm_dict['Rx Signal Power [dBm]'][1][7]
             trans_status['rxsigpowerlowwarning_flag'] = self.vdm_dict['Rx Signal Power [dBm]'][1][8]
         except KeyError:
-            helper_logger.log_debug('fields not present in VDM')
+            logging.debug('fields not present in VDM')
```
prgeor, line 818: @xincunli-sonic is this log coming to syslog file by default? Could you check?

xincunli-sonic: Thanks @prgeor, confirmed this logging would be consider as discard, so, I pushed a new change with importing syslog. Also, this is a rare case, it should not impact other modules that are using syslog by race condition. Please help review it again.

xincunli-sonic:
This should be addressed by this PR: https://github.com/sonic-net/sonic-buildimage/pull/17171.

PR #426, opened 2024-01-08
File system based sensors implementation for voltage and current sensors

bmridul:
A file system based sensors implementation for Sensormond. Platforms can provide the sensor definition
in a file and Sensormond can use that to monitor the sensors.

HLD Reference: https://github.com/sonic-net/SONiC/pull/1394

#### Description

Chassis base class looks for the presence of the file for sensors definition and uses that to built a list of sensor 
objects if available.

#### Motivation and Context

See HLD mentioned above. This provides a simple implementation for platform support for Sensormond if the platforms have only Sysfs based sensors.

Unit tested and tested on target.

root@sonic:/usr/local/etc# cat sensors.yaml
voltage_sensors:
  - name: VP3P3_CPU_RTC
    sensor: '/sys/bus/i2c/devices/23-0064/hwmon/hwmon49/in9_input'
    high_thresholds: [ 3560, 3600, 3630 ]
    low_thresholds: [ 2970, 3000, 3040 ]
  - name: VP3P3_CPU_SATA
    sensor: '/sys/bus/i2c/devices/23-0064/hwmon/hwmon49/in7_input'
    high_thresholds: [ 3560, 3600, 3630 ]
    low_thresholds: [ 2970, 3003, 3040 ]

cisco@sonic:~$ show platform volt
        Sensor    Voltage    High TH    Low TH    Crit High TH    Crit Low TH    Warning          Timestamp
--------------  ---------  ---------  --------  --------------  -------------  ---------  -----------------
 VP3P3_CPU_RTC    3300 mV       3600      3000            3630           3040      False  20231125 20:01:05
VP3P3_CPU_SATA    3286 mV       3600      3003            3630           3040      False  20231125 20:01:05


prgeor:
@Junchao-Mellanox @keboliu could you review

sonic_platform_base/chassis_base.py
```diff
@@ -30,6 +32,8 @@ class ChassisBase(device_base.DeviceBase):
     REBOOT_CAUSE_HARDWARE_RESET_FROM_ASIC = "Reset from ASIC"
     REBOOT_CAUSE_NON_HARDWARE = "Non-Hardware"
 
+    SENSORS_YAML_FILE = "/usr/local/etc/sensors.yaml"
```
keboliu, line 35: Why not put the file under the platform folder? I suppose the sensor list will be platform-specific?

sonic_platform_base/sensor_fs.py
```diff
@@ -0,0 +1,159 @@
...
+        'low_thresholds': ['N/A', 'N/A', 'N/A'],
+        'position': -1,
+    }
+
+    def __init__(self, sensor_type='sensor', **kw):
+
+        super(SensorFs, self).__init__()
+
+        if 'sensor' not in kw:
+            raise Exception('Failed to initialize sensor')
```
keboliu, line 29: I suppose some daemon will call this class, will the daemon catch this Exception? What is the expected behavior for the daemon when hits this exception?

sonic_platform_base/sensor_fs.py
```diff
@@ -0,0 +1,159 @@
...
+
+    def get_high_threshold(self):
+        """Returns the sensor high threshold value"""
+        return self.high_thresholds[1]
+
+    def get_low_threshold(self):
+        """Returns the sensor low threshold value"""
+        return self.low_thresholds[1]
+
+    def set_high_threshold(self, value):
```
keboliu, line 60: I didn't go through all the related PRs, maybe I missed them, but I want to understand here, will the new threshold value be saved? if not if the daemon restarts how it will get the previous set value?

sonic_platform_base/sensor_fs.py
bmridul: Daemon is Sensormond. It will continue and not process this sensor. Look at the voltage sensor factory function below which is instantiating the sensors.

sonic_platform_base/sensor_fs.py
bmridul: This is simply providing an implementation for the base class. The threshold values in default case are picked from the sensors.yaml file. So on daemon restart, the file is read again and they are populated. This is currently not used but is provided for any future use.

sonic_platform_base/chassis_base.py
bmridul: The sensor list is platform specific, however the location of the file may not be. However I agree it might be better. I will look into this and get back.

bmridul: I have added the file under platform folder.

bmridul:
@keboliu , Pls recheck

sonic_platform_base/sensor_fs.py
```diff
@@ -0,0 +1,159 @@
...
+import logging
+
+
+class SensorFs(SensorBase):
+    """Implementation of file system based sensor class"""
+
+    __defaults = {
+        'name': '',
+        'sensor': '',
+        'high_thresholds': ['N/A', 'N/A', 'N/A'],
```
prgeor, line 19: @bmridul There are 3 high/low threshold values. Can you add comment in code what each value correspond to?

sonic_platform_base/chassis_base.py
```diff
@@ -54,9 +57,25 @@ def __init__(self):
...
 
+        # Initialize voltage and current sensors lists from data file if available
+        try:
+            (platform_path, _) = device_info.get_paths_to_platform_and_hwsku_dirs()
+            self.sensors_yaml_file = platform_path + "/sensors.yaml"
+
+            with open(self.sensors_yaml_file, 'r') as f:
+                sensors_data = yaml.safe_load(f)
+                if 'voltage_sensors' in sensors_data:
+                    self._voltage_sensor_list = sensor_fs.VoltageSensorFs.factory(sensors_data['voltage_sensors'])
```
prgeor, line 72: @bmridul where is the change where sensormond is actually using these list of sensors?

sonic_platform_base/sensor_fs.py
```diff
@@ -0,0 +1,159 @@
...
+        if self.minimum_sensor is None or tmp < self.minimum_sensor:
+            self.minimum_sensor = tmp
+        return self.minimum_sensor
+
+    def get_maximum_recorded(self):
+        """Retrieves the maximum recorded sensor measurement"""
+        tmp = self.get_value()
+        if tmp is None:
+            return None
+        if  self.maximum_sensor is None or tmp > self.maximum_sensor:
```
prgeor, line 102: @bmridul can we have these in https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/sensor_base.py as the default implementation?

sonic_platform_base/sensor_fs.py
```diff
@@ -0,0 +1,159 @@
...
+class VoltageSensorFs(SensorFs, VoltageSensorBase):
+    """File system based voltage sensor class"""
+
+    DEVICE_TYPE = "voltage_sensor"
+
+    def __init__(self, **kw):
+        super(VoltageSensorFs, self).__init__(self.DEVICE_TYPE, **kw)
+
+    @staticmethod
+    def factory(sensors_data):
```
prgeor, line 120: @bmridul Factory class need to be repeated for every Sensor. Can you make it agnostic to the sensor Class so that we don't need to add new factory for future newer sensors?

sonic_platform_base/sensor_fs.py
bmridul: ack

sonic_platform_base/chassis_base.py
bmridul: Sensormond retrieves the sensor list from the chassis class 

https://github.com/sonic-net/sonic-platform-daemons/blob/d8977f3608e5c9263d9b9d9b33087890637ac436/sonic-sensormond/scripts/sensormond#L196

Sensor list in the chassis class is populated by default from the sensors.yaml file if present, OR provided by the platform in the chassis implementation.

sonic_platform_base/sensor_fs.py
bmridul: Yes, the difference is sensor type. I will check how to do this.

sonic_platform_base/sensor_fs.py
bmridul: Should be possible. sensor_base will appear a bit different than other base classes. I will check.

sonic_platform_base/chassis_base.py
```diff
@@ -54,9 +57,25 @@ def __init__(self):
         # List of ThermalBase-derived objects representing all thermals
         # available on the chassis
         self._thermal_list = []
+
         self._voltage_sensor_list = []
         self._current_sensor_list = []
 
+        # Initialize voltage and current sensors lists from data file if available
```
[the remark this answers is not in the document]
bmridul: ok..can be done.. however that will need change in daemon repo. Will check and send another PR.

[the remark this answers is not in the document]
prgeor: @bmridul here is one way

image

sonic_platform_base/sensor_fs.py
prgeor: @bmridul here is one way

image

sonic_platform_base/chassis_base.py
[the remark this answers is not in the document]
prgeor: @bmridul daemon repo change is fine

sonic_platform_base/sensor_fs.py
bmridul: Fixed.

sonic_platform_base/chassis_base.py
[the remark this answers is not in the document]
bmridul: @prgeor I looked into this. Sensormond retrieves the sensor list from chassis and module classes which are implemented by platform. If we take this change in Sensormond code, it will have to overwrite the list provided by the platform and modify the chassis implementation. This, though technically doable in python, is not clean object oriented design. I think its better here as this is directly provided in the base chassis class.

Place where the sensor list would can get overwritten into chassis class..

https://github.com/sonic-net/sonic-platform-daemons/blob/d8977f3608e5c9263d9b9d9b33087890637ac436/sonic-sensormond/scripts/sensormond#L457

Code where it is referenced.

https://github.com/sonic-net/sonic-platform-daemons/blob/d8977f3608e5c9263d9b9d9b33087890637ac436/sonic-sensormond/scripts/sensormond#L196

Let me know what u think.

sonic_platform_base/sensor_fs.py
bmridul: I looked into this, this would imply creating the private members of this implementation be put in the abstract base class. I think these are better left here.

sonic_platform_base/chassis_base.py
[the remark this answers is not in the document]
prgeor: @bmridul lets discuss. sensormond is the only one standing odd for its implementation details in chassis_base.py which is completely abstract.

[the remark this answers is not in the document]
bmridul: I have put this login in Sensormond. Sensormond looks at this in addition to the chassis class, so the chassis API does not need to change. Will send sensormond PR shortly.

sonic_platform_base/sensor_fs.py
prgeor: @bmridul this part I am still not clear...why most of these APIs are not in base class given that it has nothing to do with FS implement or not

prgeor (approved):

PR #427, opened 2024-01-10
Fix memory map parsing issue for SFF8636

stephenxs:
#### Description

Fix memory map parsing issue for SFF8636: `extended` can not be parsed from `10/40G Ethernet Compliance Code` due to being masked.
The field `10/40G Ethernet Compliance Code` is a bit field but is parsed as an enumeration with the assumption that only 1 bit can be set, which fails to parse a value in case more than 1 bit is set. However, the bit `extended` can co-exist with bit `40GBASE-CR4`, which means the parsing logic can fail in such cases.
In PR https://github.com/sonic-net/sonic-platform-common/pull/412 the `extended` bit was masked out but it failed to parse `extended` bit as a result.
In this PR, the combination `extended` + `40GBASE-CR4` is handled by adding the value to the enumeration.

### DIFF
--- a/sonic_platform_base/sonic_xcvr/codes/public/sff8636.py
+++ b/sonic_platform_base/sonic_xcvr/codes/public/sff8636.py
@@ -60,6 +60,7 @@ class Sff8636Codes(Sff8024):
         32: "10GBASE-LR",
         64: "10GBASE-LRM",
         128: "Extended",
+        136: "40GBASE-CR4,Extended"
     }
 
     SONET_COMPLIANCE = {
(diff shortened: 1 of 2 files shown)


keboliu (approved):

prgeor (approved):

StormLiangMS:
Hi @stephenxs  do you test this PR with 202305? 

@prgeor  would you support this cherry pick to 202305?

stephenxs:
> Hi @stephenxs do you test this PR with 202305?
> 
> @prgeor would you support this cherry pick to 202305?

Yes, I did.

PR #448, opened 2024-03-26
Fix xcvrd crash when CMIS cable is flat memory

chiourung:
#### Description
Fix https://github.com/sonic-net/sonic-buildimage/issues/18468
xcvrd crashes due to PR https://github.com/sonic-net/sonic-platform-common/pull/397.

Insert CMIS cable with flat memory

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -14,7 +14,7 @@ from ...codes.public.sff8024 import Sff8024
 from ...fields import consts
 from ..xcvr_api import XcvrApi
 from .cmisCDB import CmisCdbApi
-from .cmisVDM import CmisVdmApi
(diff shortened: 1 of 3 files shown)


prgeor:
@chiourung appreciate the quick fix. I think the real issue here is the Xcvrd is calling `post_port_dom_info_to_db()` function which should not be called for flat memory. If you do that, I think we won't see the crash.  @mihirpat1 thoughts?

mihirpat1:
> @chiourung appreciate the quick fix. I think the real issue here is the Xcvrd is calling `post_port_dom_info_to_db()` function which should not be called for flat memory. If you do that, I think we won't see the crash. @mihirpat1 thoughts?

@prgeor Agreed, `post_port_dom_info_to_db` is not needed for flat memory modules since it has static information.

prgeor:
@chiourung closing since this is now addressed via PR https://github.com/sonic-net/sonic-platform-daemons/pull/458

PR #454, opened 2024-04-15
Added new APIs and enhanced the required APIs in "module_base.py and

rameshraghupathy:
Added new APIs and enhanced the required APIs in "module_base.py and chassis_base.py" to support SmartSwitch

#### Description
The ChassisBase class and ModuleBase class need to support new SmartSwitch platforms. 

#### Motivation and Context
Some new API and some API enhancements are required to support SmartSwitch.  This has been captured in the PMON HLD document.  These changes are implementation based on those requirements.

The corresponding platform APIs were implemented.  The new CLI changes and chassisd changes invoke these APIs. UT was performed and demo was done.

### DIFF
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -313,10 +313,13 @@ class ChassisBase(device_base.DeviceBase):
         index: An integer, the index of the module to retrieve
 
         Returns:
-            A string giving the NPU-DPU port association:
-            Ex: For index: 1 will return the dpu0 port association which is
-            "Ethernet192: Ethernet0" where the string left of ":" (Ethernet192)
-            is the NPU port and the string right of ":" (Ethernet0) is the DPU port
+            A dictionary giving the NPU-DPU port association:
+            Ex: When queried for DPU0 it will return
+            {
+                "interface": {"Ethernet224": "Ethernet0"}
+            }
+            where "Ethernet192: is the NPU port and the string
+            right of ":" (Ethernet0) is the DPU port
         """
         raise NotImplementedError
 
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -242,7 +242,7 @@ class ChassisBase(device_base.DeviceBase):
     def get_module(self, index):
         """
         Retrieves module represented by (0-based) index <index>
-        On a SmartSwitch index:0 will fetch switch, index:1 will fetch
+        On a SmartSwitch index:0 is not used, index:1 will fetch
         DPU0 and so on
 
         Args:
@@ -250,7 +250,7 @@ class ChassisBase(device_base.DeviceBase):
             retrieve
 
         Returns:
-            An object dervied from ModuleBase representing the specified
+            An object derived from ModuleBase representing the specified
             module
         """
         module = None
@@ -270,7 +270,7 @@ class ChassisBase(device_base.DeviceBase):
         Args:
             module_name: A string, prefixed by SUPERVISOR, LINE-CARD or FABRIC-CARD
             Ex. SUPERVISOR0, LINE-CARD1, FABRIC-CARD5
-            SmartSwitch Example: SWITCH, DPU1, DPU2 ... DPUX
+            SmartSwitch Example: DPU0, DPU1, DPU2 ... DPUX
 
         Returns:
             An integer, the index of the ModuleBase object in the module_list
@@ -313,7 +313,7 @@ class ChassisBase(device_base.DeviceBase):
         Returns:
             A string giving the NPU-DPU port association:
             Ex: For index: 1 will return the dup0 port association which is
-            "Ethernet-BP0: Ethernet0" where the string left of ":" (Ethernet-BP0)
+            "Ethernet192: Ethernet0" where the string left of ":" (Ethernet192)
             is the NPU port and the string right of ":" (Ethernet0) is the DPU port
         """
         raise NotImplementedError
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -299,7 +299,7 @@ class ChassisBase(device_base.DeviceBase):
         Returns:
             Returns:True for SmartSwitch and False for other platforms
         """
-        raise NotImplementedError
+        return False 
 
     def get_module_dpu_data_port(self, index):
         """
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -221,21 +221,23 @@ class ChassisBase(device_base.DeviceBase):
     def get_num_modules(self):
         """
         Retrieves the number of modules available on this chassis
-        On a SmarSwitch chassis this includes the number of DPUs.
+        On a SmarSwitch chassis this will be the number of DPUs.
 
         Returns:
-            An integer, the number of modules available on this chassis
+            An integer, the number of modules available on this chassis.
+            On a SmartSwitch this will be the number of DPUs
         """
         return len(self._module_list)
 
     def get_all_modules(self):
         """
         Retrieves all modules available on this chassis. On a SmartSwitch
-        chassis this includes the DPUs.
+        chassis this will return all the DPUs.
 
         Returns:
             A list of objects derived from ModuleBase representing all
-            modules available on this chassis
+            modules available on this chassis. On a SmartSwitch this 
+            will be a list of DPU objects.
         """
         return self._module_list
 
@@ -312,7 +314,7 @@ class ChassisBase(device_base.DeviceBase):
 
         Returns:
             A string giving the NPU-DPU port association:
-            Ex: For index: 1 will return the dup0 port association which is
+            Ex: For index: 1 will return the dpu0 port association which is
             "Ethernet192: Ethernet0" where the string left of ":" (Ethernet192)
             is the NPU port and the string right of ":" (Ethernet0) is the DPU port
         """
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -289,8 +289,8 @@ class ChassisBase(device_base.DeviceBase):
         Returns None for non-smartswitch chassis.
 
         Returns:
-            An integer, indicating the DPU ID Ex: name:DPU0 return value 1,
-            name:DPU1 return value 2, name:DPUX return value X+1
+            An integer, indicating the DPU ID Ex: name:DPU0 return value 0,
+            name:DPU1 return value 1, name:DPUX return value X
         """
         raise NotImplementedError
 
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -244,8 +244,7 @@ class ChassisBase(device_base.DeviceBase):
     def get_module(self, index):
         """
         Retrieves module represented by (0-based) index <index>
-        On a SmartSwitch index:0 is not used, index:1 will fetch
-        DPU0 and so on
+        On a SmartSwitch index:0 will fetch DPU0 and so on
 
         Args:
             index: An integer, the index (0-based) of the module to
@@ -306,7 +305,7 @@ class ChassisBase(device_base.DeviceBase):
     def get_module_dpu_data_port(self, index):
         """
         Retrieves the DPU data port NPU-DPU association represented for
-        the DPU index. Platforms that need to overwrite the platform.json
+        the DPU index. Platforms that need to overwrite the hwsku.json
         file will use this API. This is valid only on the Switch and not on DPUs
 
         Args:
@@ -318,7 +317,7 @@ class ChassisBase(device_base.DeviceBase):
             {
                 "interface": {"Ethernet224": "Ethernet0"}
             }
-            where "Ethernet192: is the NPU port and the string
+            where "Ethernet224: is the NPU port and the string
             right of ":" (Ethernet0) is the DPU port
         """
         raise NotImplementedError
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -151,9 +151,7 @@ class ChassisBase(device_base.DeviceBase):
             A bool value, should return False by default or for fixed-platforms.
             Should return True for supervisor-cards, line-cards etc running as part
             of modular-chassis.
-            For SmartSwitch platforms this should return True even if they are
-            fixed-platforms, as they are treated like a modular chassis as the
-            DPU cards are treated like line-cards of a modular-chassis.
+            For SmartSwitch this should return False.
         """
         return False
 
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -298,27 +298,7 @@ class ChassisBase(device_base.DeviceBase):
         Returns:
             Returns:True for SmartSwitch and False for other platforms
         """
-        return False 
-
-    def get_module_dpu_data_port(self, index):
-        """
-        Retrieves the DPU data port NPU-DPU association represented for
-        the DPU index. Platforms that need to overwrite the hwsku.json
-        file will use this API. This is valid only on the Switch and not on DPUs
-
-        Args:
-        index: An integer, the index of the module to retrieve
-
-        Returns:
-            A dictionary giving the NPU-DPU port association:
-            Ex: When queried for DPU0 it will return
-            {
-                "interface": {"Ethernet224": "Ethernet0"}
-            }
-            where "Ethernet224: is the NPU port and the string
-            right of ":" (Ethernet0) is the DPU port
-        """
-        raise NotImplementedError
+        return False
 
     ##############################################
     # Fan methods
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -151,6 +151,9 @@ class ChassisBase(device_base.DeviceBase):
             A bool value, should return False by default or for fixed-platforms.
             Should return True for supervisor-cards, line-cards etc running as part
             of modular-chassis.
+            For SmartSwitch platforms this should return True even if they are
+            fixed-platforms, as they are treated like a modular chassis as the
+            DPU cards are treated like line-cards of a modular-chassis.
         """
         return False
 
@@ -218,6 +221,7 @@ class ChassisBase(device_base.DeviceBase):
     def get_num_modules(self):
         """
         Retrieves the number of modules available on this chassis
+        On a SmarSwitch chassis this includes the number of DPUs.
 
         Returns:
             An integer, the number of modules available on this chassis
@@ -226,7 +230,8 @@ class ChassisBase(device_base.DeviceBase):
 
     def get_all_modules(self):
         """
-        Retrieves all modules available on this chassis
+        Retrieves all modules available on this chassis. On a SmartSwitch
+        chassis this includes the DPUs.
 
         Returns:
             A list of objects derived from ModuleBase representing all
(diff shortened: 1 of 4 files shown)


sonic_platform_base/module_base.py
```diff
@@ -196,6 +206,62 @@ def get_maximum_consumed_power(self):
         """
         raise NotImplementedError
 
+    ##############################################
+    # SmartSwitch methods
+    ##############################################
+
+    def get_reboot_cause(self):
```
gpunathilell, line 213: Based on the function docstring this should be `get_dpu_id` instead of `get_reboot_cause` 
```suggestion
    def get_dpu_id(self):
```

sonic_platform_base/module_base.py
```diff
@@ -21,6 +21,8 @@ class ModuleBase(device_base.DeviceBase):
     MODULE_TYPE_SUPERVISOR = "SUPERVISOR"
     MODULE_TYPE_LINE    = "LINE-CARD"
     MODULE_TYPE_FABRIC  = "FABRIC-CARD"
+    MODULE_TYPE_DPU  = "DPU"
+    MODULE_TYPE_SWITCH  = "SWITCH"
```
oleksandrivantsiv, line 25: Do we need to have module type "SWITCH"? What is its purpose?

oleksandrivantsiv, line 25: Do we need to have module type "SWITCH"? What is its purpose?

sonic_platform_base/module_base.py
```diff
@@ -196,6 +206,62 @@ def get_maximum_consumed_power(self):
...
+        Retrieves the dpu state object having the detailed dpu state progression.
+        Fetched from ChassisStateDB.
+
+        Returns:
+            An object instance of the DPU_STATE (see DB schema)
+            Returns None on switch module
+        """
+        raise NotImplementedError
+
+    def get_health_info(self):
```
oleksandrivantsiv, line 253: The mechanism of fetching the health information from the DPU should be vendor-independent. Do we need to add it to the Platform API?

sonic_platform_base/chassis_base.py
```diff
@@ -263,12 +270,54 @@ def get_module_index(self, module_name):
...
+        raise NotImplementedError
+
+    def is_smartswitch(self):
+        """
+        Retrieves whether the sonic instance is part of smartswitch
+
+        Returns:
+            Returns:True for SmartSwitch and False for other platforms
+        """
+        raise NotImplementedError
```
gpunathilell, line 302: Since this API call is supposed to return `False` for all the other platforms, isn't it better to return `False` by default instead of raising a `NotImplementedError`?

sonic_platform_base/module_base.py
```diff
@@ -196,6 +206,62 @@ def get_maximum_consumed_power(self):
...
+        """
+        Retrieves the DPU ID. Returns None for non-smartswitch chassis.
+
+        Returns:
+            An integer, indicating the DPU ID. DPU0 returns 1, DPUX returns X+1
+            Returns '0' on switch module
+        """
+        raise NotImplementedError
+
+    def get_reboot_cause(self):
```
gpunathilell, line 223: The strings for the causes mentioned below need to be defined in `ModuleBase` similar to the ones in  `ChassisBase` (https://github.com/sonic-net/sonic-platform-common/blob/5389771ecc014e3e274588fe63ed99036c32b00b/sonic_platform_base/chassis_base.py#L20)

sonic_platform_base/module_base.py
rameshraghupathy: This is a Microsoft requirement.  My understanding is that, "SWITCH" is the functional  equivalent of "SUPERVISOR" of a modular chassis.

sonic_platform_base/module_base.py
rameshraghupathy: Yes, Thanks! Fixed it.

sonic_platform_base/module_base.py
rameshraghupathy: Agree, done!

sonic_platform_base/module_base.py
rameshraghupathy: Yes, though the fetching of the DPU health-info is vendor-independent, as per the pmon hld that we all agreed upon the health-info needs to be updated into redis chassisStateDB and its contents is exactly the same as the Switch health-info. So, in a way the health-info structure is fixed besides the vendor specific fields. There are some emerging use-cases where this will come in handy. So let us keep it.

sonic_platform_base/module_base.py
prgeor: @rameshraghupathy @oleksandrivantsiv one purpose is to have the output to display different modules in a chassis/smart switch

image

sonic_platform_base/chassis_base.py
```diff
@@ -263,12 +270,54 @@ def get_module_index(self, module_name):
...
+        Retrieves the DPU data port NPU-DPU association represented for
+        the DPU index. Platforms that need to overwrite the platform.json
+        file will use this API. This is valid only on the Switch and not on DPUs
+
+        Args:
+        index: An integer, the index of the module to retrieve
+
+        Returns:
+            A string giving the NPU-DPU port association:
+            Ex: For index: 1 will return the dup0 port association which is
```
vvolam, line 315: nit: spelling error "dpu0"

sonic_platform_base/module_base.py
rameshraghupathy: It was decided that SWITCH module is not needed and the only module in smartswitch is DPU.

sonic_platform_base/module_base.py
rameshraghupathy: All of us have agreed to remove the "get_health_info" API and it is removed in the HLD

sonic_platform_base/chassis_base.py
```diff
@@ -226,7 +230,8 @@ def get_num_modules(self):
 
     def get_all_modules(self):
         """
-        Retrieves all modules available on this chassis
+        Retrieves all modules available on this chassis. On a SmartSwitch
+        chassis this includes the DPUs.
 
         Returns:
             A list of objects derived from ModuleBase representing all
```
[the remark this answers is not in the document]
rameshraghupathy: Done

sonic_platform_base/chassis_base.py
rameshraghupathy: Done

sonic_platform_base/module_base.py
```diff
@@ -196,6 +219,54 @@ def get_maximum_consumed_power(self):
...
+            A tuple (string, string) where the first element is a string
+            containing the cause of the previous reboot. This string must
+            be one of the predefined strings in this class. If the first
+            string is "REBOOT_CAUSE_HARDWARE_OTHER", the second string can be
+            used to pass a description of the reboot cause.
+
+        """
+        raise NotImplementedError
+
+    def get_state_info(self):
```
oleksandrivantsiv, line 249: What is the difference between `get_state_info` and `get_oper_status`? Can you please add an example of the expected return value?

From the HLD, I understood that `get_oper_status` should return a simple online or offline state. And `get_state_info` should return the state progression. But from the description of the `get_oper_status`, it is different in this PR.

rameshraghupathy: Sure will add an example of the return value.

sonic_platform_base/module_base.py
```diff
@@ -152,6 +170,11 @@ def get_oper_status(self):
             A string, the operational status of the module from one of the
             predefined status values: MODULE_STATUS_EMPTY, MODULE_STATUS_OFFLINE,
             MODULE_STATUS_FAULT, MODULE_STATUS_PRESENT or MODULE_STATUS_ONLINE
+            The SmartSwitch platforms will have these additional status
```
oleksandrivantsiv, line 173: This should remain the same

sonic_platform_base/module_base.py
```diff
@@ -196,6 +214,67 @@ def get_maximum_consumed_power(self):
...
+            'dpu_data_plane_time': '20240626 21:10:07',
+            'dpu_midplane_link_reason': 'INTERNAL-MGMT : admin state - UP, oper_state - UP, status - OK, HOST-MGMT : admin state - UP, oper_state - UP, status - OK',
+            'dpu_midplane_link_state': 'UP',
+            'dpu_midplane_link_time': '20240626 21:13:25',
+            'id': '0'
+            }
+        """
+        raise NotImplementedError
+
+    def get_health_info(self):
```
oleksandrivantsiv, line 268: I guess now, this should be removed

sonic_platform_base/module_base.py
```diff
@@ -196,6 +214,67 @@ def get_maximum_consumed_power(self):
...
+    def get_state_info(self):
+        """
+        Retrieves the dpu state object having the detailed dpu state progression.
+        Fetched from ChassisStateDB.
+
+        Returns:
+            An object instance of the DPU_STATE (see DB schema)
+            Returns None on switch module
+
+        Sample Output: {
```
oleksandrivantsiv, line 253: A small suggestion here is not changes the object to the following:
```
{
    'dpu_control_plane': {'state': 'UP', 'time': '20240626 21:13:25', 'reason': '...'},
    'dpu_data_plane': {'state': 'UP', 'time': '20240626 21:13:25', 'reason': '...'},
    'dpu_midplane_link': {'state': 'UP', 'time': '20240626 21:13:25', 'reason': '...'}
}
``` 
This simplifies the implementation and makes the object compact.

oleksandrivantsiv:
LGTM. I have two small requests, but besides that, all is good.

sonic_platform_base/module_base.py
rameshraghupathy: yes

sonic_platform_base/module_base.py
```diff
@@ -196,6 +214,63 @@ def get_maximum_consumed_power(self):
...
+    ##############################################
+    # SmartSwitch methods
+    ##############################################
+
+    def get_dpu_id(self):
+        """
+        Retrieves the DPU ID. Returns None for non-smartswitch chassis.
+
+        Returns:
+            An integer, indicating the DPU ID. DPU0 returns 0, DPUX returns X
```
prgeor, line 226: @rameshraghupathy please specify DPU ID >= 0

sonic_platform_base/chassis_base.py
```diff
@@ -151,6 +151,9 @@ def is_modular_chassis(self):
             A bool value, should return False by default or for fixed-platforms.
             Should return True for supervisor-cards, line-cards etc running as part
             of modular-chassis.
+            For SmartSwitch platforms this should return True even if they are
```
vvolam, line 154: NIT: This statement is little confusing at first read. Should it be "For SmartSwitch platforms, this should return True even if they are fixed platforms, as they are treated like a modular chassis. The DPU cards in these systems are treated as line cards of a modular chassis."

sonic_platform_base/chassis_base.py
```diff
@@ -218,32 +221,38 @@ def get_component(self, index):
...
-            modules available on this chassis
+            modules available on this chassis. On a SmartSwitch this 
+            will be a list of DPU objects.
         """
         return self._module_list
 
     def get_module(self, index):
         """
         Retrieves module represented by (0-based) index <index>
+        On a SmartSwitch index:0 is not used, index:1 will fetch
```
vvolam, line 247: Are we starting from index 0 or 1? As the statement for get_dpu_id() is contradicting?

sonic_platform_base/chassis_base.py
```diff
@@ -263,12 +272,57 @@ def get_module_index(self, module_name):
...
+        Args:
+        index: An integer, the index of the module to retrieve
+
+        Returns:
+            A dictionary giving the NPU-DPU port association:
+            Ex: When queried for DPU0 it will return
+            {
+                "interface": {"Ethernet224": "Ethernet0"}
+            }
+            where "Ethernet192: is the NPU port and the string
```
vvolam, line 321: Is it "Ethernet 224"?

sonic_platform_base/chassis_base.py
```diff
@@ -263,12 +272,57 @@ def get_module_index(self, module_name):
...
             module_name: A string, prefixed by SUPERVISOR, LINE-CARD or FABRIC-CARD
             Ex. SUPERVISOR0, LINE-CARD1, FABRIC-CARD5
+            SmartSwitch Example: DPU0, DPU1, DPU2 ... DPUX
 
         Returns:
             An integer, the index of the ModuleBase object in the module_list
         """
         raise NotImplementedError
 
+    ##############################################
```
vvolam, line 282: Do we have an API to power-on DPU, if it is down?

sonic_platform_base/chassis_base.py
```diff
@@ -263,12 +272,57 @@ def get_module_index(self, module_name):
...
+    def is_smartswitch(self):
+        """
+        Retrieves whether the sonic instance is part of smartswitch
+
+        Returns:
+            Returns:True for SmartSwitch and False for other platforms
+        """
+        return False 
+
+    def get_module_dpu_data_port(self, index):
```
prgeor, line 306: @rameshraghupathy We have a `role` type to indicate DPU data port. You can define this in port_confi.ini or hwsku.json 
Example https://github.com/sonic-net/sonic-buildimage/blob/master/device/mellanox/x86_64-nvidia_sn4280-r0/ACS-SN4280/hwsku.json#L125

Can we make use of hwsku.json or port_config.ini? Lets keep one approach. I am not sure if this hwsku.json/port_confi.ini is already mentioned in HLD

sonic_platform_base/module_base.py
```diff
@@ -541,6 +616,8 @@ def get_midplane_ip(self):
         line-card and return the midplane IP-address of the line-card.
         When called from the line-card, the module will represent the
         Supervisor and return its midplane IP-address.
+        When called from the DPU, returns the midplane IP-address of the dpu-card.
```
prgeor, line 619: @rameshraghupathy This is a module_base Class so this will not be called from DPU.

sonic_platform_base/module_base.py
```diff
@@ -541,6 +616,8 @@ def get_midplane_ip(self):
...
+        When called from the Switch returns the midplane IP-address of Switch.
```
prgeor, line 620: @rameshraghupathy This will be called from the NPU host and for each DPU it should return the corresponding midplane ip address

sonic_platform_base/chassis_base.py
rameshraghupathy: Good catch! Fixed it.

sonic_platform_base/chassis_base.py
rameshraghupathy: yup

sonic_platform_base/chassis_base.py
rameshraghupathy: Yes, under module_base  "def set_admin_state(self, up):"

sonic_platform_base/module_base.py
```diff
@@ -196,6 +214,63 @@ def get_maximum_consumed_power(self):
...
+            A tuple (string, string) where the first element is a string
+            containing the cause of the previous reboot. This string must
+            be one of the predefined strings in this class. If the first
+            string is "REBOOT_CAUSE_HARDWARE_OTHER", the second string can be
+            used to pass a description of the reboot cause.
+
+        """
+        raise NotImplementedError
+
+    def get_state_info(self):
```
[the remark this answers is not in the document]
rameshraghupathy: This was agreed upon and is driven by the HLD. We have to change it everywhere, if we decide to. Can we make a decision?

sonic_platform_base/module_base.py
```diff
@@ -196,6 +214,63 @@ def get_maximum_consumed_power(self):
...
+            },
+            'dpu_data_plane': {
+                'state': 'UP',
+                'time': '20240626 21:13:25',
+                'reason': 'DPU container named polaris is running, pciemgrd running : OK'
+            },
+            'dpu_midplane_link': {
+                'state': 'UP',
+                'time': '20240626 21:13:25',
+                'reason': 'INTERNAL-MGMT : admin state - UP, oper_state - UP, status - OK'
```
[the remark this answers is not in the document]
rameshraghupathy: This is the time of each of the sate change event. For example the 'time' in 'dpu_data_plane' represents when the 'state' update happened. The control-plane and data-plane state change time will be updated by the DPUs.  The 'midplane_link' state would be updated by the NPU

sonic_platform_base/module_base.py
rameshraghupathy: Done

sonic_platform_base/chassis_base.py
rameshraghupathy: I'm using hwsku.json

sonic_platform_base/module_base.py
rameshraghupathy: Fixed it.

sonic_platform_base/module_base.py
rameshraghupathy: Yes, Fixed it.  Now reads
`    def get_midplane_ip(self):
        """
        Retrieves the midplane IP-address of the module in a modular chassis
        When called from the Supervisor, the module could represent the
        line-card and return the midplane IP-address of the line-card.
        When called from the line-card, the module will represent the
        Supervisor and return its midplane IP-address.
        
        When called from the SmartSwitch returns the midplane IP-address of
        the DPU module.

        Returns:
            A string, the IP-address of the module reachable over the midplane`

sonic_platform_base/chassis_base.py
rameshraghupathy: Fixed it:  "For SmartSwitch this should return False."

sonic_platform_base/chassis_base.py
```diff
@@ -263,12 +269,57 @@ def get_module_index(self, module_name):
...
+        """
+        Retrieves the DPU data port NPU-DPU association represented for
+        the DPU index. Platforms that need to overwrite the hwsku.json
```
vvolam, line 306: Are we going to use platform.json or hwsku.json?

vvolam (approved):
LGTM, Please close on Prince comments before merging. Thanks!

sonic_platform_base/chassis_base.py
prgeor: @rameshraghupathy can we then remove this API as we follow suggested HLD approach?

sonic_platform_base/module_base.py
```diff
@@ -17,10 +17,25 @@ class ModuleBase(device_base.DeviceBase):
     # Device type definition. Note, this is a constant.
     DEVICE_TYPE = "module"
 
+    # Possible reboot causes
```
prgeor, line 20: @rameshraghupathy is there any need to duplicate here? Can we use from Chasssis base? The platform file where these reboot causes will be used can anyways import them from chassis base and use. I don't see these used in this file. If so can this be removed?

sonic_platform_base/chassis_base.py
rameshraghupathy: Removed

sonic_platform_base/chassis_base.py
rameshraghupathy: We are going to use platform.json

sonic_platform_base/module_base.py
rameshraghupathy: I didn't make any change here. Your question is valid. We can revisit this with an AI as this may break the platform code.

prgeor: @rameshraghupathy I still see the REBOOT causes defined in this file

rameshraghupathy: Yes, this can be cleaned.  Removed it.

prgeor (approved):

PR #465, opened 2024-05-07
Add base class implementation for local users' passwords reset

azmy98:
#### Description
I added base class implementation for local-users-passwords-reset feautre.

#### Motivation and Context
It defines the abstract class the vendor needs to implement for local-users-passwords-reset feature

By inheriting from this base class

### DIFF
--- a/sonic_platform_base/local_users_passwords_reset_base.py
+++ b/sonic_platform_base/local_users_passwords_reset_base.py
@@ -1,0 +1,59 @@
+'''
+    local_users_passwords_reset_base.py
+
+    Abstract base class for implementing platform-specific
+    local users' passwords reset base functionality for SONiC
+'''
+try:
+    import json
+    import subprocess
+
+    from sonic_py_common.logger import Logger
+except ImportError as e:
+    raise ImportError (str(e) + "- required module not found")
+
+# Global logger class instance
+logger = Logger()
+
+
+DEFAULT_USERS_FILEPATH = '/etc/sonic/default_users.json'
+
+
+class LocalUsersConfigurationResetBase(object):
+    """
+    Abstract base class for resetting local users' passwords on the switch
+    """
+    def should_trigger(self):
+        '''
+        define the condition to trigger
+        '''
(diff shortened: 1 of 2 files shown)


tests/local_users_passwords_reset_base_test.py
```diff
@@ -0,0 +1,31 @@
...
+from unittest import mock
+from sonic_platform_base.local_users_passwords_reset_base import LocalUsersConfigurationResetBase
+
+
+class TestLocalUsersConfigurationResetBase:
+    '''
+    Collection of LocalUsersConfigurationResetBase test methods
+    '''
+    @staticmethod
+    def test_fan_base():
```
dgsudharsan, line 14: Copy paste error. Can you rename the API (test_fan_base)?

dgsudharsan (changes requested):

dgsudharsan (approved):

sonic_platform_base/local_users_passwords_reset_base.py
```diff
@@ -0,0 +1,59 @@
...
+        # the condition to trigger start() method, the default implementation will be by checking if a long reboot press was detected.
+        raise NotImplementedError
+
+    @staticmethod
+    def reset_password(user, hashed_password, expire=False):
+        '''
+        This method is used to reset the user's password and expire it (optional) using Linux shell commands.
+        '''
+        # Use 'chpasswd' shell command to change password
+        subprocess.call([f"echo '{user}:{hashed_password}' | sudo chpasswd -e"], shell=True)
```
qiluo-msft, line 39: Do not concat command line. It is vulnerable to command argument injection. For example, hashed_password may include blank char.

qiluo-msft: @maipbui to help review this part.

maipbui: Suggest not to use shell=True, instead use shell=False with an array of strings

azmy98: It did not work, the quotations are not working correctly when I split it into list and the pipe does not work

PR #472, opened 2024-05-16
[Component]Fix test_get_firmware_update_notification fail

PJHsieh:
#### Description
Fix test_get_firmware_update_notification fail

#### Motivation and Context

Erro Log:

> pytest_assert(isinstance(notif, STRING_TYPE), "Component {}: Firmware update notification appears to be incorrect from image {}".format(i, image))
E Failed: Component 0: Firmware update notification appears to be incorrect from image current

Refer to API get_firmware_update_notification header, shall return 'None' instead of None.

"""
Retrieves a notification on what should be done in order to complete the component firmware update

Args:
    image_path: A string, path to firmware image

Returns:
    A string containing the component firmware update notification if required.
    By default 'None' value will be used, which indicates that no actions are required
"""

Run platform_tests/api/test_component.py::TestComponentApi::test_get_firmware_update_notification with no NO api implementation.


prgeor:
@PJHsieh please check the code coverge

PR #474, opened 2024-06-06
Add XcvrApi for configuring application 0

noaOrMlnx:
#### Description

When NOS configures the CMIS modules, we need to be aware to a case where some of the lanes are disabled.
According to CMIS spec, in that case, an application 0 is needed to make sure all disabled lanes are zeroed.
This function sets "0" in all lanes before configuring the module, so de-activated lanes will remain with 0. 

#### Motivation and Context

This was done according to CMIS specification document.

This addition to CMIS thread was tested with a CMIS active cable, with only 4 active lanes (and 4 de-active).
We made sure the link is up, right places are zeroed, and have traffic.

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -2483,7 +2483,7 @@ class CmisApi(XcvrApi):
         This function applies App0 configuration - in case some lanes are de-activated
         '''
         for lane in range(self.NUM_CHANNELS):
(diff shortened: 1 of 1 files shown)


prgeor:
@noaOrMlnx please check this API (`decommsion_all_datapaths`) which is invoked by CMIS manager thread at the beginning of configuring any speed to put all lanes with 0 application to start with https://github.com/sonic-net/sonic-platform-common/blob/cd6a5a42d2d9e4973afe1bc5d8b1b5ff5f02d54c/sonic_platform_base/sonic_xcvr/api/public/cmis.py#L2126

noaOrMlnx:
@prgeor thanks, I checked it and indeed this works and solved the issue. I will close 2 relevant PRs.

PR #479, opened 2024-06-28
fix(sfp): fix rx power conversion

ishidawataru:
#### Description

fixes how to calculate the rx power when the external calibration is used according to the SFF-8472 Rev 12.4 section 9.3.

#### Motivation and Context

fixes https://github.com/sonic-net/sonic-platform-common/issues/449

No. Asking the original reporter of this bug to test this as I don't have access to a real device that uses this feature


sonic_platform_base/sonic_sfp/sff8472.py
```diff
@@ -828,49 +821,23 @@ def calc_rx_power(self, eeprom_data, offset, size):
...
+                result = 0
+                for i in range(0, 5):
+                    field_name = f'RX_PWR_{i}'
+                    rx_pwr = self.dom_ext_calibration_constants[field_name]
+                    off = rx_pwr['offset']
+                    size = rx_pwr['size']
+                    coeff = self.float_from_bytes(eeprom_data[off:off+size])
+                    result += coeff * rx_pwr_ad ** i
+
+                result = result * 0.0001 # uW to mW
```
JaapKeuter, line 840: Here the comment should make it clear that the `result` value is in 0.1 uW units, not uW (the formula in the documentation at section 9.3 is _also_ misleading in that regards). Otherwise the conversion factor may be interpreted off by an order of magnitude by an unsuspecting reader.

JaapKeuter:
This code has been tested, see [issue 449](https://github.com/sonic-net/sonic-platform-common/issues/449#issuecomment-2227283551), and found to be correct. The only remark is about a comment in the code.

With a changed comment this code should be merged, IMHO.

JaapKeuter:
@ishidawataru : Could you pay attention to this change, i.e. finalizing the draft and putting it forward for merging please?

PR #481, opened 2024-07-03
Generate media key in platform sepcific way and fix cable issues

jemifdo:
#### Description
- Add API to support the generation of the preemphasis key in a platform
  vendor dictated way instead of generic one
- Add break to return the cable length immediately and avoid reading the
  junk values
- Ignore ASCII errors as some cables might have some junk values causing
  the daemon to exit

Verified in Celestica platform DS2000 and DS3000 with Amphenol and Molex DAC


prgeor:
@jemifdo are you still looking for this fix. Can you elaborate the bug in detail please?

prgeor:
@jemifdo closing this

PR #491, opened 2024-08-19
Make is_flat_memory more robust

Junchao-Mellanox:
#### Description

Currently, `is_flat_memory` could return 3 different values:

- True when the module is flat memory
- False when the module is not flat memory
- None when the EEPROM read returns None or empty bytearray

When it returns None, we should not assume that the module is NOT a flat memory. Instead, it would be safer to treat it as a flat memory.

 

#### Motivation and Context
This PR is to make `is_flat_memory` only returns two state, True or False. And if EEPROM reading returns None or empty bytearray, we should treat the module as a flat memory.

Manual test

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -368,7 +368,7 @@ class CmisApi(XcvrApi):
         return float("{:.3f}".format(voltage))
 
     def is_flat_memory(self):
-        return self.xcvr_eeprom.read(consts.FLAT_MEM_FIELD)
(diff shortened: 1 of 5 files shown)


mihirpat1:
@Junchao-Mellanox Can you please modify the below description as well?
https://github.com/sonic-net/sonic-platform-common/blob/7f07fa68eff8954cbd36400d39fd6f3d81b9a850/sonic_platform_base/sonic_xcvr/api/xcvr_api.py#L494-L496

keboliu (dismissed):

mihirpat1 (approved):

keboliu (approved):

PR #500, opened 2024-09-20
Add is_copper definition to cmis api

bobbymcgonigle:
#### Description

For CMIS it's pretty easy to tell if a module is copper because the media interface will always be passive_copper_media_interface. This PR adds a check for that.

#### Motivation and Context

Similar to other SFF xcvr api's it would be helpful to know if a CMIS module is copper or optical. We can use this information for tunings and in other areas where appropriate.

Manually tested a lot of copper and optical modules ensuring they return the correct type.

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -372,7 +372,7 @@ class CmisApi(XcvrApi):
         Returns True if the module is copper, False otherwise
         '''
         media_intf = self.get_module_media_type()
-        return media_intf == "passive_copper_media_interface" or media_intf == "active_cable_media_interface"
+        return media_intf == "passive_copper_media_interface"
 
     def is_flat_memory(self):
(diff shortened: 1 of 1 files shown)


sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -367,6 +367,13 @@ def get_voltage(self):
             return None
         return float("{:.3f}".format(voltage))
 
+    def is_copper(self):
+        '''
+        Returns True if the module is copper, False otherwise
+        '''
+        media_intf = self.get_module_media_type()
+        return media_intf == "passive_copper_media_interface" or media_intf == "active_cable_media_interface"
```
prgeor, line 375: @active cable is not passive media. Active cables like AEC have datapath state machine. @mihirpat1 can you check

mihirpat1: @prgeor Confirmed that active cables like AEC have datapath state machine. Hence, this check needs to be modified.

prgeor:
@bobbymcgonigle please reopen if required. closing for now to remove from tracking

PR #504, opened 2024-10-09
Add negative case for pci_detach() and pci_reattach()

vvolam:
Address the comments added in https://github.com/sonic-net/sonic-platform-common/pull/501

#### Description

- Add negative for pci_detach() and pci_reattach()
- Remove module_name from get_bus_info(), pci_reattach() and pci_detach()
- Add MODULE_TYPE_SMARTSWITCH to reboot API. This reboot_type is used for rebooting DPUs when entire smartswitch undergoes for a reboot.
- Also fix unittests accordingly.

### DIFF
--- a/sonic_platform_base/module_base.py
+++ b/sonic_platform_base/module_base.py
@@ -266,7 +266,7 @@ class ModuleBase(device_base.DeviceBase):
         """
         raise NotImplementedError
 
-    def get_bus_info(self):
+    def get_pci_bus_info(self):
         """
         Retrieves the bus information.
 
@@ -284,7 +284,7 @@ class ModuleBase(device_base.DeviceBase):
         """
         raise NotImplementedError
 
-    def pci_reattach(self):
+    def pci_rescan(self):
         """
         Rescans and reconnects the PCI device.
 
--- a/sonic_platform_base/module_base.py
+++ b/sonic_platform_base/module_base.py
@@ -52,6 +52,8 @@ class ModuleBase(device_base.DeviceBase):
     MODULE_REBOOT_FPGA_COMPLEX = "FPGA"
     # Module reboot type to reboot DPU
     MODULE_REBOOT_DPU = "DPU"
+    # Module reboot type to reboot SMART SWITCH
+    MODULE_REBOOT_SMARTSWITCH = "SMARTSWITCH"
 
     def __init__(self):
         # List of ComponentBase-derived objects representing all components
@@ -168,9 +170,11 @@ class ModuleBase(device_base.DeviceBase):
         Args:
             reboot_type: A string, the type of reboot requested from one of the
             predefined reboot types: MODULE_REBOOT_DEFAULT, MODULE_REBOOT_CPU_COMPLEX,
-            MODULE_REBOOT_FPGA_COMPLEX or MODULE_REBOOT_DPU
+            MODULE_REBOOT_FPGA_COMPLEX, MODULE_REBOOT_DPU or MODULE_REBOOT_SMARTSWITCH
 
             MODULE_REBOOT_DPU is only applicable for smartswitch chassis.
+
+            MODULE_REBOOT_SMARTSWITCH is only applicable for smartswitch chassis.
 
         Returns:
             bool: True if the request has been issued successfully, False if not
@@ -277,6 +281,15 @@ class ModuleBase(device_base.DeviceBase):
 
         Returns: True once the PCI is successfully detached.
         Returns False, if PCI detachment fails or specified device is not found.
+        """
+        raise NotImplementedError
+
+    def pci_reattach(self):
+        """
+        Rescans and reconnects the PCI device.
+
+        Returns: True once the PCI is successfully reconnected.
+        Returns False, if PCI rescan fails or specified device is not found.
         """
         raise NotImplementedError
 
--- a/sonic_platform_base/module_base.py
+++ b/sonic_platform_base/module_base.py
@@ -284,7 +284,7 @@ class ModuleBase(device_base.DeviceBase):
         """
         raise NotImplementedError
 
-    def pci_rescan(self):
+    def pci_reattach(self):
         """
         Rescans and reconnects the PCI device.
 
--- a/sonic_platform_base/module_base.py
+++ b/sonic_platform_base/module_base.py
@@ -262,28 +262,22 @@ class ModuleBase(device_base.DeviceBase):
         """
         raise NotImplementedError
 
-    def get_bus_info(self, module_name):
+    def get_bus_info(self):
         """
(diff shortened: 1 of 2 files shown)


rameshraghupathy (approved):

sonic_platform_base/module_base.py
```diff
@@ -264,7 +264,7 @@ def get_state_info(self):
 
     def get_bus_info(self, module_name):
```
gpunathilell, line 265: Do we need the module_name as a parameter for these functions? The method would be part of the `Module` class which is implemented, which means that `get_bus_info` `get_pci_reattach` and `get_pci_detach` would be executed for specific modules

vvolam: good point. Fixed it. Thank you!

oleksandrivantsiv (approved):

prgeor:
@vvolam please add more details in the PR description.

sonic_platform_base/module_base.py
```diff
@@ -262,28 +262,22 @@ def get_state_info(self):
         """
         raise NotImplementedError
 
-    def get_bus_info(self, module_name):
+    def get_bus_info(self):
         """
-        Retrieves the bus information for the specified by "module_name" on a SmartSwitch.
+        Retrieves the bus information. Returns None for non-smartswitch chassis.
```
prgeor, line 267: @vvolam do we need to mention smartswitch or not here? This is abstract class and let the implementation choose to decide whether its supported or not. The calller should validate the return value

sonic_platform_base/module_base.py
```diff
@@ -262,28 +262,22 @@ def get_state_info(self):
...
         Returns:
             Returns the PCI bus information in BDF format like "[DDDD:]BB:SS:F"
         """
         raise NotImplementedError
 
-    def pci_detach(self, module_name):
+    def pci_detach(self):
         """
         Detaches the DPU PCI device specified by "module_name" on a SmartSwitch.
+        Returns False for non-smartswitch chassis.
```
prgeor, line 278: @vvolam same comment as above

sonic_platform_base/module_base.py
vvolam: @prgeor Fixed it. Thank you!

sonic_platform_base/module_base.py
vvolam: Done!

sonic_platform_base/module_base.py
```diff
@@ -52,6 +52,8 @@ class ModuleBase(device_base.DeviceBase):
     MODULE_REBOOT_FPGA_COMPLEX = "FPGA"
     # Module reboot type to reboot DPU
     MODULE_REBOOT_DPU = "DPU"
+    # Module reboot type to reboot SMART SWITCH
```
prgeor, line 55: @vvolam the module class is for pluggable or non-pluggable modules. We already have DPU reboot. Why we need MODULE_REBOOT_SMARTSWITCH?

sonic_platform_base/module_base.py
```diff
@@ -262,28 +266,30 @@ def get_state_info(self):
         """
         raise NotImplementedError
 
-    def get_bus_info(self, module_name):
+    def get_bus_info(self):
```
prgeor, line 269: @vvolam  suggest to use get_pci_bus_info() since there can be many bus

sonic_platform_base/module_base.py
```diff
@@ -262,28 +266,30 @@ def get_state_info(self):
...
-        Detaches the DPU PCI device specified by "module_name" on a SmartSwitch.
+        Detaches the PCI device.
 
         Returns: True once the PCI is successfully detached.
+        Returns False, if PCI detachment fails or specified device is not found.
         """
         raise NotImplementedError
 
-    def pci_reattach(self, module_name):
+    def pci_reattach(self):
```
prgeor, line 287: @vvolam nit...rename to pci_rescan ?

sonic_platform_base/module_base.py
```diff
@@ -262,28 +266,30 @@ def get_state_info(self):
...
-        Retrieves the bus information for the specified by "module_name" on a SmartSwitch.
+        Retrieves the bus information.
 
         Returns:
             Returns the PCI bus information in BDF format like "[DDDD:]BB:SS:F"
         """
         raise NotImplementedError
 
-    def pci_detach(self, module_name):
+    def pci_detach(self):
```
prgeor, line 278: @vvolam nit rename it as pci_remove?

vvolam: @prgeor pci_remove is Linux kernel standard function name. To avoid confusion and conflicts better to avoid same names.

sonic_platform_base/module_base.py
vvolam: @prgeor,  Though I modified, wanted to rename back as pci_reattach is Linux kernel standard function name. To avoid confusion and conflicts better to avoid same names.

sonic_platform_base/module_base.py
vvolam: @prgeor Done. Thanks!

sonic_platform_base/module_base.py
vvolam: @prgeor As updated in the https://github.com/sonic-net/SONiC/pull/1847, MODULE_REBOOT_SMARTSWITCH is the reboot type when rebooting the DPU and when entire switch is going for a reboot.

sonic_platform_base/module_base.py
```diff
@@ -168,10 +170,12 @@ def reboot(self, reboot_type):
         Args:
             reboot_type: A string, the type of reboot requested from one of the
             predefined reboot types: MODULE_REBOOT_DEFAULT, MODULE_REBOOT_CPU_COMPLEX,
-            MODULE_REBOOT_FPGA_COMPLEX or MODULE_REBOOT_DPU
+            MODULE_REBOOT_FPGA_COMPLEX, MODULE_REBOOT_DPU or MODULE_REBOOT_SMARTSWITCH
```
rameshraghupathy, line 173: Why does the module need to know if other modules are rebooting as well? Do you expect anything different that needs to be done in the case of a chassis reboot?

rameshraghupathy (commented):
Looks ok to me

sonic_platform_base/module_base.py
vvolam: According to our last discussion, in the context of the DPU, it needs to determine whether it is undergoing an individual reboot or if the entire switch is rebooting. MODULE_REBOOT_DPU is the reboot type when individual DPU is undergoing reboot and MODULE_REBOOT_SMARTSWITCH is when DPU undergoing reboot as part of entire switch reboot.

prgeor (approved):

PR #507, opened 2024-10-22
[SmartSwitch] Add a new API for the DPU chassis to query dataplane and midplane states

oleksandrivantsiv:
#### Description
Add a definition for a new DPU chassis API required for querying DPU dataplane and midplane states.

#### Motivation and Context

A new API is required to enable querying of the DPU dataplane and midplane states. These states will be monitored by the chassisd service running on the DPU and pushed to the CHASSIS_STATE_DB upon any changes. This will allow the NPU to subscribe to DPU state changes.  

The API tests will be added in scope of chassisd changes.


sonic_platform_base/chassis_base.py
```diff
@@ -280,26 +280,64 @@ def get_module_index(self, module_name):
     # SmartSwitch methods
     ##############################################
 
-    def get_dpu_id(self, name):
+    def get_dpu_id(self, **kwargs):
```
prgeor, line 283: @oleksandrivantsiv since this is a base class can we make use `get_module_index()` ?

sonic_platform_base/chassis_base.py
```diff
@@ -280,26 +280,64 @@ def get_module_index(self, module_name):
...
+        """
+        Checks whether the current SONiC instance is running on a DPU.
 
         Returns:
-            Returns:True for SmartSwitch and False for other platforms
+            bool: True if the instance is running on a DPU, False otherwise.
         """
         return False
 
+    def get_dpu_dataplane_state(self):
```
prgeor, line 318: @oleksandrivantsiv do we need to have dpu specified in function name or keep it generic as `get_dataplane_state()`  since this is an abstract base class

sonic_platform_base/chassis_base.py
```diff
@@ -280,26 +280,64 @@ def get_module_index(self, module_name):
...
+        Retrieves the status of the DPU dataplane.
+
+        This method is applicable only on Smart Switch DPUs.
+
+        Returns:
+            bool: True if the dataplane is UP, False if it is down.
+        """
+        raise NotImplementedError
+
+    def get_dpu_controlplane_state(self):
```
prgeor, line 329: @oleksandrivantsiv do we need to have dpu specified in function name or keep it generic as `get_controlplane_state()`  since this is an abstract base class

sonic_platform_base/chassis_base.py
oleksandrivantsiv: `get_module_index` has a different meaning. It returns `An integer, the index of the ModuleBase object in the module_list`. `get_dpu_id` returns the physical ID of the DPU (its position)

dgsudharsan (approved):

sonic_platform_base/chassis_base.py
prgeor: @oleksandrivantsiv can you rebase, looks like this API already exist https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/chassis_base.py#L283

oleksandrivantsiv: Yes, the `get_dpu_id` already exists. I changed the parameter from `name` to `**kwargs`. This API will have a different meaning for switch and DPU. Please check the description

prgeor (approved):

prgeor:
@rameshraghupathy can you review

sonic_platform_base/chassis_base.py
```diff
@@ -280,26 +280,60 @@ def get_module_index(self, module_name):
...
+    def is_dpu(self):
+        """
+        Checks whether the current SONiC instance is running on a DPU.
+
+        Returns:
+            bool: True if the instance is running on a DPU, False otherwise.
+        """
+        return False
+
+    def get_dataplane_state(self):
```
rameshraghupathy, line 318: get_state_info(self): API already covers this.  Please refer to the HLD.  This API appears to be redundant.

sonic_platform_base/chassis_base.py
```diff
@@ -280,26 +280,60 @@ def get_module_index(self, module_name):
...
+        """
+        Retrieves the status of the dataplane.
+
+        Returns:
+            bool: True if the dataplane is UP, False if it is down.
+        """
+        raise NotImplementedError
+
+    def get_controlplane_state(self):
```
rameshraghupathy, line 327: get_state_info(self): API already covers this.  Please refer to the HLD.  This API appears to be redundant.

sonic_platform_base/chassis_base.py
oleksandrivantsiv: @rameshraghupathy we discussed this last week. This is needed for the DPU chassisd to populate the data plane and control plane states from the DPU to the CHASSIS_STATE_DB. This API is defined on the chassis level. What you are referring to is the module-level API that will run on the NPU side and has a completely different meaning.

rameshraghupathy: Got it

rameshraghupathy (approved):

PR #518, opened 2024-11-29
[CMIS]The "get_error_description" function should return 'OK' instead of None when there are no errors.

keboliu:
#### Description

The "get_error_description" function should return 'OK' instead of 'None' when there are no errors.

#### Motivation and Context
When there are no errors on the cable, get_error_description currently returns 'None'. In this case, we should return 'OK' so that the CLI (show interface transceiver error-description) provides a proper output. 

Call this API directly on different cables with different error statuses.


Junchao-Mellanox (approved):

keboliu:
@prgeor  would you please help to merge?

liat-grozovik (approved):

liat-grozovik:
@keboliu on which branches this change should go?

PR #521, opened 2024-12-13
High power class enabling for SFF-8636 modules

longhuan-cisco:
#### Description

Add platform common API support for enabling high power class if module's power class is greater or equal to 5 (Refer to SFF-8636 spec `6.2.6 Control Functions (Page 00h, Bytes 86-99)` for byte 93 definition)

#### Motivation and Context

Without enabling high power class for those modules, the module won't operate properly with power output and link won't come up.

Verified the module operate with proper power output.

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/sff8636.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/sff8636.py
@@ -440,11 +440,9 @@ class Sff8636Api(XcvrApi):
         '''
         ret = True
 
-        if power_class < 5:
-            return ret
-        elif power_class >= 8:
+        if power_class >= 8:
             ret = self.xcvr_eeprom.write(consts.HIGH_POWER_CLASS_ENABLE_CLASS_8, enable)
-        else:  # Power class 5, 6, 7
+        elif power_class >= 5:
             ret = self.xcvr_eeprom.write(consts.HIGH_POWER_CLASS_ENABLE_CLASS_5_TO_7, enable)
 
         return ret
--- a/sonic_platform_base/sonic_xcvr/api/public/sff8636.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/sff8636.py
@@ -12,6 +12,7 @@ from ...codes.public.sff8636 import Sff8636Codes
 
 class Sff8636Api(XcvrApi):
     NUM_CHANNELS = 4
+    POWER_CLASS_PREFIX = "Power Class "
 
     def __init__(self, xcvr_eeprom):
         super(Sff8636Api, self).__init__(xcvr_eeprom)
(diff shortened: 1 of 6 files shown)


mihirpat1:
@longhuan-cisco Please help in fixing the built failure

longhuan-cisco:
> @longhuan-cisco Please help in fixing the built failure

@mihirpat1 It was due to issue in CICD infra. Issue was gone after rerun.

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
```diff
@@ -12,6 +12,7 @@
 
 class Sff8636Api(XcvrApi):
     NUM_CHANNELS = 4
+    POWER_CLASS_PREFIX = "Power Class "
```
prgeor, line 15: @longhuan-cisco is the space at the end necessary?

sonic_platform_base/sonic_xcvr/api/xcvr_api.py
```diff
@@ -637,3 +637,14 @@ def get_error_description(self):
         """
         raise NotImplementedError
 
+    def set_high_power_class(self, enable):
```
mihirpat1, line 640: @longhuan-cisco Any reason for not defining this API in https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/sonic_xcvr/sfp_optoe_base.py too?

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
```diff
@@ -403,3 +404,46 @@ def get_lpmode(self):
...
+        ret = True
+
+        if power_class < 5:
+            return ret
+        elif power_class >= 8:
+            ret = self.xcvr_eeprom.write(consts.HIGH_POWER_CLASS_ENABLE_CLASS_8, enable)
+        else:  # Power class 5, 6, 7
+            ret = self.xcvr_eeprom.write(consts.HIGH_POWER_CLASS_ENABLE_CLASS_5_TO_7, enable)
+
+        return ret
```
prgeor, line 449: @longhuan-cisco can we keep the set_high_power_class API as dumb. Let the SFF manager do the logic handling as to what power class to program?

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
```diff
@@ -403,3 +404,46 @@ def get_lpmode(self):
...
+            return power_class
+        if not power_class_str.startswith(self.POWER_CLASS_PREFIX):
+            return power_class
+
+        prefix_len = len(self.POWER_CLASS_PREFIX)
+        power_class = int(power_class_str[prefix_len:prefix_len + 1])
+
+        return power_class
+
+    def set_high_power_class(self, enable):
```
prgeor, line 428: @longhuan-cisco let the argument take pclass, enable two arguments. see other comment below

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
```diff
@@ -403,3 +404,46 @@ def get_lpmode(self):
...
+        power_class = self.get_power_class()
+        ret = True
+
+        if power_class < 5:
+            return ret
+        elif power_class >= 8:
+            ret = self.xcvr_eeprom.write(consts.HIGH_POWER_CLASS_ENABLE_CLASS_8, enable)
+        else:  # Power class 5, 6, 7
+            ret = self.xcvr_eeprom.write(consts.HIGH_POWER_CLASS_ENABLE_CLASS_5_TO_7, enable)
+
```
prgeor, line 448: @longhuan-cisco you also need to make sure that the SW is in control for driving the power class. The way to do that is by doing a write of `1` to Byte 93 bit 0 i.e power override. Let do that in sff manager.

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
prgeor: @longhuan-cisco the API `set_high_power_class()` is considered to be dumb, provides the eeprom write to enable the SFF-mgr requested power class. Its upto the SFF manager to read the advertisment to figure out what power class is supported by the module and only pass that as an argument to the API.

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
prgeor: @longhuan-cisco the point you are missing here is if you dont set the power override bit to 1 then LPmode must be de-asserted which is an extra thing to ensure. In SW control, where B93 bit 0 = 1, LPmode is don't care. Anyways, we should discuss this in your SFF manager PR

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
```diff
@@ -403,3 +405,46 @@ def get_lpmode(self):
...
+            bool: True if the provision succeeds, False if it fails
+        '''
+        ret = True
+
+        if power_class < 5:
+            return ret
+        elif power_class >= 8:
+            ret = self.xcvr_eeprom.write(consts.HIGH_POWER_CLASS_ENABLE_CLASS_8, enable)
+        else:  # Power class 5, 6, 7
+            ret = self.xcvr_eeprom.write(consts.HIGH_POWER_CLASS_ENABLE_CLASS_5_TO_7, enable)
```
prgeor, line 448: @longhuan-cisco can we simply this as follows
if power_class >= 8:

else if power_class >= 5

prgeor (dismissed):

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
longhuan-cisco: updated it.

sonic_platform_base/sonic_xcvr/api/public/sff8636.py
longhuan-cisco: Moved this discussion to sff_mgr PR at https://github.com/sonic-net/sonic-platform-daemons/pull/574#discussion_r1989878742

prgeor (approved):

PR #522, opened 2024-12-15
Change Virtium SSD which doesn't support SmartCMD, to use only smartctl

noaOrMlnx:
#### Description

Add support for Virtium SSD which can't use SmartCMD - to use only smartctl (using generic ssd info parser).

#### Motivation and Context

There are Virtium SSDs which can't receive SSD info from 'SmartCMD', and must use 'smartctl' command. 
For those SSDs, after calling the generic ssd parser, the function is finished and the specific vendor parser won't be called. 

I tested it manually with a special Virtium SSD which doesn't support SmartCMD

output after the change:
root@sonic:/home/admin# show platform ssdhealth 
Disk Type    : NVME
Device Model : Virtium VTPM24CEXI080-BM110006
Health       : 100.0%
Temperature  : 52.0C

output before the change:
root@sonic:/home/admin# show platform ssdhealth 
[Error] Cannot read SMART information on device /dev/nvme0n1
Disk Type    : NVME
Device Model : Virtium VTPM24CEXI080-BM110006
Health       : N/A
Temperature  : 51.0C

### DIFF
--- a/sonic_platform_base/sonic_storage/ssd.py
+++ b/sonic_platform_base/sonic_storage/ssd.py
@@ -84,7 +84,6 @@ class SsdUtil(StorageCommon):
             "M.2"               : { "utility" : INNODISK, "parser" : self.parse_innodisk_info },
             "StorFly"           : { "utility" : VIRTIUM,  "parser" : self.parse_virtium_info },
             "Virtium"           : { "utility" : VIRTIUM,  "parser" : self.parse_virtium_info },
-            "SpecialVirtium"    : { "utility" : SMARTCTL,  "parser" : self.parse_generic_ssd_info },
             "Swissbit"          : { "utility" : SMARTCTL, "parser" : self.parse_swissbit_info },
             "Micron"            : { "utility" : SMARTCTL, "parser" : self.parse_micron_info },
             "Intel"             : { "utility" : SMARTCTL, "parser" : self.parse_intel_info },
@@ -104,10 +103,12 @@ class SsdUtil(StorageCommon):
 
         # Known vendor part
         if self.model:
+            # For some Virtium SSDs, parse_generic_ssd_info should be called.
+            # Since it was called above, no need to parse a specific vendor SSD info.
             if self.model in ['Virtium VTPM24CEXI080-BM110006']:
-                vendor = 'SpecialVirtium'
-            else:
-                vendor = self._parse_vendor()
+                return
+
+            vendor = self._parse_vendor()
             if vendor:
                 try:
                     self.fetch_vendor_ssd_info(diskdev, vendor)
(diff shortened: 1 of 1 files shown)


noaOrMlnx:
@wangxin A kind reminder, can you please handle the failure which is caused by moving from Bullseye to Bookworm? 
Thanks

keboliu (dismissed):

noaOrMlnx:
@prgeor can you please take a look?

sonic_platform_base/sonic_storage/ssd.py
```diff
@@ -103,7 +104,10 @@ def fetch_parse_info(self, diskdev):
 
         # Known vendor part
         if self.model:
-            vendor = self._parse_vendor()
+            if self.model in ['Virtium VTPM24CEXI080-BM110006']:
+                vendor = 'SpecialVirtium'
```
assrinivasan, line 108: Hi @noaOrMlnx - suggestion: since we are already using smartctl in the fetch generic SSD info function, we can simply return if the vendor is this special virtium model. That way, the already fetched and parsed info will not be overwritten.

assrinivasan (changes requested):

sonic_platform_base/sonic_storage/ssd.py
noaOrMlnx: Hi @assrinivasan, I am not sure what do you mean - the vendor is not special virtium, but virtium. 
Special Virtium "vendor" is made for the Virtium SSD which doesn't support the use of SmartCMD. 
So if we are using an SSD which doesn't support SmartCMD, it will be part of the "SpecialVirtium" list and will use smartctl instead.

assrinivasan: Hi @noaOrMlnx - if you look at line 102, we are first calling generic parser for all types of storage devices. The generic parser already uses smartctl.

So my point here is that there is no need for a separate 'SpecialVirtium' that uses smartctl again. If you determine that the model of this SSD is the one you mention in line 107, instead of setting vendor = SpecialVirtium you can simply return from this function and avoid re-parsing data using smartctl.

noaOrMlnx: @assrinivasan all comments were addressed, and PR description has changed accordingly. 
Thanks

keboliu (approved):

assrinivasan (approved):

assrinivasan:
@noaOrMlnx could you also fix output before/after change to have uniform output format? One has `disk type` whereas other does not.

assrinivasan (approved):

noaOrMlnx:
> @noaOrMlnx could you also fix output before/after change in the PR description to have uniform output format? One has `disk type` whereas other does not.

@assrinivasan Done

noaOrMlnx:
@bingwang-ms can you please help to check that?

noaOrMlnx:
@bingwang-ms @prgeor kind reminder. can you please help to merge?

liat-grozovik (approved):

PR #527, opened 2025-01-03
Create is_transceiver_vdm_supported API for CMIS transceivers

mihirpat1:
#### Description

is_transceiver_vdm_supported API needs to be created so that xcvrd can directly skip freezing/unfreezing of VDM stats (done to read VDM and PM data from the transceiver) for transceivers which do not support VDM feature.

#### Motivation and Context

XCVRD will call `is_transceiver_vdm_supported` API before performing freeze of VDM stats.

Following types of modules were tested with this changeset
1. CMIS module supporting VDM - ensured is_transceiver_vdm_supported returns True
2. SFF-8472 module - ensured is_transceiver_vdm_supported returns False

MSFT ADO - 30657693

### DIFF
--- a/sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
+++ b/sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
@@ -55,6 +55,10 @@ class SfpOptoeBase(SfpBase):
         api = self.get_xcvr_api()
         return api.is_coherent_module() if api is not None else None
 
+    def is_transceiver_vdm_supported(self):
+        api = self.get_xcvr_api()
+        return api.is_transceiver_vdm_supported() if api is not None else None
+
     def get_transceiver_vdm_real_value(self):
         """
         Retrieves VDM real (sample) values for this xcvr (applicable for CMIS and C-CMIS)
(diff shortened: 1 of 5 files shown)


sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
```diff
@@ -55,6 +55,10 @@ def is_coherent_module(self):
         api = self.get_xcvr_api()
         return api.is_coherent_module() if api is not None else None
 
+    def is_transceiver_vdm_supported(self):
+        api = self.get_xcvr_api()
+        return api.is_transceiver_vdm_supported() if api is not None else None
```
prgeor, line 60: return False on falure? @mihirpat1

mihirpat1: @prgeor I am currently returning None because the caller (`xcvrd` in this case) treats None as False and will not perform VDM-related operations. This approach ensures that the current behavior remains unchanged. Additionally, if in the future the caller needs to differentiate between False (explicitly not supported) and None (indeterminate or error state), the current implementation will allow for this distinction without requiring changes to the existing implementation.

prgeor (approved):

prgeor:
@mihirpat1 we need in 202412 branch

mihirpat1:
@kperumalbfn Can you please help to cherry-pick this to 202411 branch?
@r12f Can you please help to cherry-pick this to 202412 branch?

MSFT ADO - 30657693

PR #546, opened 2025-02-21
Set the eeprom cache permissions to 755

PJHsieh:
#### Description

The issue arises because, after a system reboot, the permissions of /var/cache/sonic/decode-syseeprom/ or its file /var/cache/sonic/decode-syseeprom/syseeprom_cache may be incorrectly set. This misconfiguration can cause the Platform API to require root privileges for access.

This problem has been observed in various platforms, such as the AS7326_56X, where running show platform syseeprom results in errors due to permission issues.

#### Motivation and Context

To address this, it's essential to ensure that the cache directory and its files have the correct permissions set upon system startup. Implementing a fix that adjusts these permissions appropriately can prevent the Platform API from requiring root privileges, thereby resolving the issue.

Reboot and use admin account to run show platform syseeprom.


PJHsieh:
@prgeor , please review this PR, thanks.

PR #547, opened 2025-02-25
Adding is_copper api to cmis

bobbymcgonigle:
#### Description
For CMISwe can tell if a module is copper because the media interface will always be passive_copper_media_interface. This PR adds a check for that.

#### Motivation and Context
Similar to other SFF xcvr api's it would be helpful to know if a CMIS module is copper or optical. We can use this information for tunings and in other areas where appropriate.

Wrote a unit test and tested manually on device using multiple optics and DACs

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -524,6 +524,13 @@ class CmisApi(XcvrApi):
             return None
         return float("{:.3f}".format(voltage))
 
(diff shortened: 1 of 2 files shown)


prgeor:
@bobbymcgonigle will review once out of draft

bobbymcgonigle:
> @bobbymcgonigle will review once out of draft

Thanks Prince, this PR is ready for review

prgeor (approved):

arista-hpandya:
Can we port this to msft-202503? Thanks!

PR #548, opened 2025-02-28
Restore optoe page to default 0 for active cables

prgeor:
#### Description
Restore the optoe's page to page ZERO before reading the Lower page > 127

#### Motivation and Context
When the module is busy with CDB command, the optoe kernel driver at the end of user page read request, tries to restore the page to Page ZERO and if this write fails because the module could be busy with CDB command (since these modules donot support background CDB mode).  Because of this when Xcvrd tries to read offset (> 127) in Lower page, since optoe does not restore the page to 0 for any Lower page access, the page read to these offset will result in wrong page to be read (because of previous page select write failure) and Xcvrd parsing fails causing crash.

Now in optoe EEPROM read API we ensure to restore the Page to ZERO page if the Page select byte is non-zero

TBD:
1. Test on Arista switch
2. Test on Cisco switch

### DIFF
--- a/sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
+++ b/sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
@@ -7,7 +7,9 @@
 
 from ..sfp_base import SfpBase
 
-SFP_OPTOE_PAGE_OFFSET = 127
+SFP_OPTOE_PAGE_SELECT_OFFSET = 127
+SFP_OPTOE_UPPER_PAGE0_OFFSET = 128
+SFP_OPTOE_PAGE_SIZE = 128
 
 class SfpOptoeBase(SfpBase):
     def __init__(self):
@@ -264,10 +266,10 @@ class SfpOptoeBase(SfpBase):
             pass
 
     def get_optoe_current_page(self):
-        return self.read_eeprom(SFP_OPTOE_PAGE_OFFSET, 1)
+        return self.read_eeprom(SFP_OPTOE_PAGE_SELECT_OFFSET, 1)[0]
 
     def set_page0(self):
-        self.write_eeprom(SFP_OPTOE_PAGE_OFFSET, 1, bytearray([0x00]))
+        self.write_eeprom(SFP_OPTOE_PAGE_SELECT_OFFSET, 1, bytearray([0x00]))
 
     def set_optoe_write_timeout(self, write_timeout):
         sys_path = self.get_eeprom_path()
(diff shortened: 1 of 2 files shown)


sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
```diff
@@ -273,6 +283,12 @@ def set_optoe_write_timeout(self, write_timeout):
     def read_eeprom(self, offset, num_bytes):
         try:
             with open(self.get_eeprom_path(), mode='rb', buffering=0) as f:
+                if offset >= SFP_OPTOE_UPPER_PAGE0_OFFSET  and \
+                    offset < (SFP_OPTOE_UPPER_PAGE0_OFFSET+SFP_OPTOE_PAGE_SIZE) and \
+                        self.get_optoe_current_page() != 0:
+                # Restoring the page to 0 helps in cases where the optoe driver failed to restore
+                # the page when say the module was busy with CDB command processing
```
mihirpat1, line 290:
```suggestion
                    # Restoring the page to 0 helps in cases where the optoe driver failed to restore
                    # the page when say the module was busy with CDB command processing
```

mihirpat1:
@prgeor Wondering if the current changeset requires special handling for SFF-8472 based transceivers since these transceivers have page select byte at byte 127, A2h instead of A0h.

prgeor:
> @prgeor Wondering if the current changeset requires special handling for SFF-8472 based transceivers since these transceivers have page select byte at byte 127, A2h instead of A0h.

@mihirpat1 Good question. I think we should be good in that case because 0xA0 address space offset > 127 is a reserved space. We don't have a use case today

image


In other words, all Access for offset > 127 will be in 0xA2 address space in SFP case.

mihirpat1 (approved):

vdahiya12 (approved):

yxieca:
Why do you only need to reset the page code in 128-255 offset range? do you also need to reset the page code when offset is greater than 255?

yxieca (approved):

prgeor:
> Why do you only need to reset the page code in 128-255 offset range? do you also need to reset the page code when offset is greater than 255?

@yxieca The optoe kernel driver before r/w to the page >=1 already sets the page selection register.

prgeor:
@kperumalbfn could you help cherry pick to 202411
@bingwang-ms could you help cherry pick to 202405

mihirpat1:
@kperumalbfn @bingwang-ms 
MSFT ADO - 31626749

PR #557, opened 2025-04-01
[Smartswitch] Add module specific pcie attach/detach functions for smartswitch platforms

gpunathilell:
#### Description

As there could be platforms which have multiple PCIE devices per dpu, the module_base implementation is handling the PCIE removal and attachment along with adding the entry details in the PCIE table, so that the pcie daemon which is running will ignore the errors generated from the DPUs,

#### Motivation and Context

This was done because there are multiple PCIE devices in some platforms and only one in others, we need to have a platform independent method for removal and attach algorithms along with state db entries

### DIFF
--- a/sonic_platform_base/module_base.py
+++ b/sonic_platform_base/module_base.py
@@ -7,7 +7,14 @@
 
 import sys
 from . import device_base
+import swsscommon
+import json
+import threading
 
+# PCI state database constants
+PCIE_DETACH_INFO_TABLE = "PCIE_DETACH_INFO"
+PCIE_OPERATION_DETACHING = "detaching"
+PCIE_OPERATION_ATTACHING = "attaching"
 
 class ModuleBase(device_base.DeviceBase):
     """
@@ -16,6 +23,8 @@ class ModuleBase(device_base.DeviceBase):
     """
     # Device type definition. Note, this is a constant.
     DEVICE_TYPE = "module"
+
+    _pci_operation_lock = threading.Lock()
 
     # Possible card types for modular chassis
     MODULE_TYPE_SUPERVISOR = "SUPERVISOR"
@@ -73,6 +82,8 @@ class ModuleBase(device_base.DeviceBase):
         self._thermal_list = []
         self._voltage_sensor_list = []
         self._current_sensor_list = []
+        self.state_db_connector = None
+        self.pci_bus_info = None
 
         # List of SfpBase-derived objects representing all sfps
         # available on the module
@@ -81,6 +92,22 @@ class ModuleBase(device_base.DeviceBase):
         # List of ASIC-derived objects representing all ASICs
         # visibile in PCI domain on the module
         self._asic_list = []
+    
+    def __del__(self):
+        """
+        Destructor - ensures any PCI state DB entries are cleaned up when the module is deleted
+        """
+        try:
+            bus_info_list = self.get_pci_bus_info()
+            for bus in bus_info_list:
+                self.pci_entry_state_db(bus, PCIE_OPERATION_ATTACHING)
+        except NotImplementedError:
+            pci_bus = self.get_pci_bus_from_platform_json()
+            if pci_bus:
+                self.pci_entry_state_db(pci_bus, PCIE_OPERATION_ATTACHING)
+        except Exception:
+            pass
+
 
     def get_base_mac(self):
         """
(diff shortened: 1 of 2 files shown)


gpunathilell:
@vvolam @rameshraghupathy Please review

rameshraghupathy (commented):
@vvolam @gpunathilell   Adding @bmridul to review

wenchungw (approved):

sonic_platform_base/module_base.py
```diff
@@ -271,10 +294,126 @@ def get_pci_bus_info(self):
...
+                self.state_db_connector = swsscommon.swsscommon.DBConnector("STATE_DB", 0)
+            if operation == PCIE_OPERATION_ATTACHING:
+                self.state_db_connector.delete(PCIE_DETACH_INFO_TABLE_KEY)
+                return
+            self.state_db_connector.hset(PCIE_DETACH_INFO_TABLE_KEY, "bus_info", pcie_string)
+            self.state_db_connector.hset(PCIE_DETACH_INFO_TABLE_KEY, "dpu_state", operation)
+        except Exception as e:
+            sys.stderr.write("Failed to write pcie bus infoto state database: {}\n".format(str(e)))
+
+    def pci_reattach_from_platform_json(self):
```
rameshraghupathy, line 382: This is platform specific in the sense that "/sys/bus/pci/rescan" may or may not be suitable for all platforms.  Can you move this to your module.py ?

gpunathilell: @rameshraghupathy Correct me if I am wrong, but there is no other way of performing the rescan (unless the DPUs are under the same parent bus, but doing it in the current method would work even if the DPUs are not under the same bus, so the current implementation is the most generic)
https://github.com/torvalds/linux/blob/7a13c14ee59d4f6c5f4277a86516cbc73a1383a8/Documentation/ABI/testing/sysfs-bus-pci#L74
If that doesn't work then we should be implementing a platform specific pcie_reattach() function.

gpunathilell: As discussed in the HLD meeting, we will look into moving the rescan for the whole PCIE tree to chassis instead of module

gpunathilell: @rameshraghupathy I had a discussion with Vasundhara, we can move the implementation for the rescan to chassis, but we would have to call both chassis rescan and module rescan separately each time (since the pci bus removal is module specific). So the plan is to remove the independent implementation currently present `pci_reattach_from_platform_json` and `pci_removal_from_platform_json`. So pci_detach() and pci_rescan() have to be implemented by vendor.

sonic_platform_base/module_base.py
```diff
@@ -271,10 +294,126 @@ def get_pci_bus_info(self):
...
+            PCIE_DETACH_INFO_TABLE_KEY = PCIE_DETACH_INFO_TABLE+"|"+pcie_string
+            if not self.state_db_connector:
+                self.state_db_connector = swsscommon.swsscommon.DBConnector("STATE_DB", 0)
+            if operation == PCIE_OPERATION_ATTACHING:
+                self.state_db_connector.delete(PCIE_DETACH_INFO_TABLE_KEY)
+                return
+            self.state_db_connector.hset(PCIE_DETACH_INFO_TABLE_KEY, "bus_info", pcie_string)
+            self.state_db_connector.hset(PCIE_DETACH_INFO_TABLE_KEY, "dpu_state", operation)
+        except Exception as e:
+            sys.stderr.write("Failed to write pcie bus infoto state database: {}\n".format(str(e)))
```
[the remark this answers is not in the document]
gpunathilell: Fixed

sonic_platform_base/module_base.py
gpunathilell: @vvolam @rameshraghupathy I have removed the relevant pcie removal code, Please re review

sonic_platform_base/module_base.py
```diff
@@ -16,6 +26,7 @@ class ModuleBase(device_base.DeviceBase):
     """
     # Device type definition. Note, this is a constant.
     DEVICE_TYPE = "module"
+    PCI_OPERATION_LOCK_FILE_PATH = "/var/lock/{}_pci.lock"
```
vvolam, line 29: Query: Is this file separate for each module?

sonic_platform_base/module_base.py
```diff
@@ -271,10 +295,90 @@ def get_pci_bus_info(self):
...
+        Handles PCI device rescan by updating state database and reattaching device.
+        If pci_reattach is not implemented, falls back to platform.json based rescan.
+
+        Returns:
+            bool: True if operation was successful, False otherwise
+        """
+        try:
+            bus_info_list = self.get_pci_bus_info()
+            with self._pci_operation_lock():
+                for bus in bus_info_list:
```
vvolam, line 376: should we do the attach first and then remove the entry just to be safe?

vvolam (commented):
LGTM, other than these minor query/comment

sonic_platform_base/module_base.py
gpunathilell: the lock should be valid across all modules, so we are using a generic lock file

vvolam: As discussed offline, could you check if this lock is still required?

sonic_platform_base/module_base.py
```diff
@@ -271,10 +295,90 @@ def get_pci_bus_info(self):
...
+        Retrieves the PCI bus information from platform.json file.
+
+        Returns:
+            str: PCI bus information if found in platform.json
+            None: If module is not found in platform.json or any error occurs
+        """
+        if self.pci_bus_info:
+            return self.pci_bus_info
+        try:
+            with open("/usr/share/sonic/platform/platform.json", 'r') as f:
```
rameshraghupathy, line 313: shouldn't this be "/usr/share/sonic/device/{platform}/platform.json" ?

sonic_platform_base/module_base.py
```diff
@@ -271,10 +295,90 @@ def get_pci_bus_info(self):
         Retrieves the bus information.
 
         Returns:
-            Returns the PCI bus information in BDF format like "[DDDD:]BB:SS:F"
+            Returns the PCI bus information in list of BDF format like "[DDDD:]BB:SS:F"
         """
         raise NotImplementedError
 
+    def get_pci_bus_from_platform_json(self):
```
rameshraghupathy, line 302: This function is not used anywhere in this file.  This looks like an utility function.  Shouldn't this be under utilities?

sonic_platform_base/module_base.py
```diff
@@ -271,10 +295,90 @@ def get_pci_bus_info(self):
...
+
+        Returns:
+            bool: True if operation was successful, False otherwise
+        """
+        try:
+            bus_info_list = self.get_pci_bus_info()
+            with self._pci_operation_lock():
+                for bus in bus_info_list:
+                    self.pci_entry_state_db(bus, PCIE_OPERATION_ATTACHING)
+                return self.pci_reattach()
```
rameshraghupathy, line 378: Is the intent to attach one module at a time?

sonic_platform_base/module_base.py
gpunathilell: Clarification, File lock is required, and the lock is applicable per module, just to prevent reattach of the module while we are removing the module

sonic_platform_base/module_base.py
gpunathilell: Function removed

sonic_platform_base/module_base.py
gpunathilell: Unused function, this is removed

sonic_platform_base/module_base.py
gpunathilell: In case of our platform, we have multiple buses, so the state db entry is added in series, but the reattach function is called only once

sonic_platform_base/module_base.py
```diff
@@ -271,10 +295,70 @@ def get_pci_bus_info(self):
...
+
+        Raises:
+            RuntimeError: If state database connection fails
+        """
+        try:
+            # Do not use import if swsscommon is not needed
+            import swsscommon
+            PCIE_DETACH_INFO_TABLE_KEY = PCIE_DETACH_INFO_TABLE+"|"+pcie_string
+            if not self.state_db_connector:
+                self.state_db_connector = swsscommon.swsscommon.DBConnector("STATE_DB", 0)
```
vvolam, line 335: Check if state_db connection is successful?

vvolam (approved):

sonic_platform_base/module_base.py
```diff
@@ -271,10 +295,70 @@ def get_pci_bus_info(self):
...
+            bool: True if operation was successful, False otherwise
+        """
+        try:
+            bus_info_list = self.get_pci_bus_info()
+            with self._pci_operation_lock():
+                for bus in bus_info_list:
+                    self.pci_entry_state_db(bus, PCIE_OPERATION_DETACHING)
+                return self.pci_detach()
+        except Exception:
+            return False
```
rameshraghupathy, line 317: Can you add an error message in case of failure?

sonic_platform_base/module_base.py
```diff
@@ -687,3 +771,79 @@ def get_all_asics(self):
...
+            if not os.path.exists(source_file):
+                return True
+
+            shutil.copy2(source_file, target_file)
+
+            # Restart sensord
+            os.system("service sensord restart")
+
+            return True
+        except Exception as e:
```
rameshraghupathy, line 798: Can you add an error message in case of failure?

sonic_platform_base/module_base.py
```diff
@@ -687,3 +771,79 @@ def get_all_asics(self):
...
+                return True
+
+            # Remove the file
+            os.remove(target_file)
+
+            # Restart sensord
+            os.system("service sensord restart")
+
+            return True
+        except Exception as e:
```
rameshraghupathy, line 824: Can you add an error message in case of failure?

sonic_platform_base/module_base.py
```diff
@@ -271,10 +295,70 @@ def get_pci_bus_info(self):
...
+        """
+        try:
+            bus_info_list = self.get_pci_bus_info()
+            with self._pci_operation_lock():
+                return_value = self.pci_reattach()
+                for bus in bus_info_list:
+                    self.pci_entry_state_db(bus, PCIE_OPERATION_ATTACHING)
+                return return_value
+        except Exception:
+            return False
```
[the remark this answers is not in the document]
gpunathilell: Done

sonic_platform_base/module_base.py
```diff
@@ -271,10 +295,72 @@ def get_pci_bus_info(self):
...
         Returns:
-            Returns the PCI bus information in BDF format like "[DDDD:]BB:SS:F"
+            Returns the PCI bus information in list of BDF format like "[DDDD:]BB:SS:F"
         """
         raise NotImplementedError
 
+    def handle_pci_removal(self):
+        """
+        Handles PCI device removal by updating state database and detaching device.
+        If pci_detach is not implemented, falls back to platform.json based removal.
```
vvolam, line 305: This comment needs to be fixed as we are not having any fallback mechanism now.

gpunathilell: Done

judyjoseph (approved):

PR #562, opened 2025-05-08
[cmis] add read only cache decorator

jianyuewu:
Improve CMIS startup performance by adding a read-only cache decorator.

#### Description
This patch introduces read only cache decorator, which caches the results of static EEPROM-read methods in CmisApi to avoid redundant I2C transactions. By memoizing these methods, we greatly reduce the number of EEPROM reads during startup. Unit tests have been added to verify correct caching behavior and cache invalidation.
Some problematic cable scenarios:
If user could have issued reset since some fields in EEPROM were not accessible initially (the cable has some problem in this case, first plugin data read is incorrect, after re-plugin, then data will be correct.) In this case, with cache enabled, there is a WA: we can reset the sfp by xcvrd restart, then data will be correct.

#### Motivation and Context
On systems managing up to 512 CMIS ports, the cumulative cost of repeated EEPROM reads at startup causes noticeable delays. Caching static getter methods eliminates unnecessary I2C calls and speeds up initialization.

Test with config reload, speed up around 2 seconds.
Before:
```bash
$ time sudo config reload -y
Acquired lock on /etc/sonic/reload.lock
Disabling container and routeCheck monitoring ...
Stopping SONiC target ...
Running command: /usr/local/bin/sonic-cfggen -j /etc/sonic/init_cfg.json -j /etc/sonic/config_db.json --write-to-db
Running command: /usr/local/bin/db_migrator.py -o migrate
Running command: /usr/local/bin/sonic-cfggen -d -y /etc/sonic/sonic_version.yml -t /usr/share/sonic/templates/sonic-environment.j2,/etc/sonic/sonic-environment
Restarting SONiC target ...
Enabling container and routeCheck monitoring ...
Reloading Monit configuration ...
Reinitializing monit daemon
Released lock on /etc/sonic/reload.lock

real    0m22.324s
user    0m1.388s
sys     0m0.361s
```
After:
```bash
$ time sudo config reload -y -f                                                                                                                                                                                                       
Acquired lock on /etc/sonic/reload.lock                                                                                                                                                                                                                   
Disabling container and routeCheck monitoring ...                                                                                                                                                                                                         
Stopping SONiC target ...                                                                                                                                                                                                                                 
Running command: /usr/local/bin/sonic-cfggen -j /etc/sonic/init_cfg.json -j /etc/sonic/config_db.json --write-to-db                                                                                                                                       
Running command: /usr/local/bin/db_migrator.py -o migrate                                                                                                                                                                                                 
Running command: /usr/local/bin/sonic-cfggen -d -y /etc/sonic/sonic_version.yml -t /usr/share/sonic/templates/sonic-environment.j2,/etc/sonic/sonic-environment                                                                                           
Restarting SONiC target ...                                                                                                                                                                                                                               
Enabling container and routeCheck monitoring ...                                                                                                                                                                                                          
Reloading Monit configuration ...                                                                                            
Reinitializing monit daemon                                                                                                  
Released lock on /etc/sonic/reload.lock                                                                                      
                                                                                                                             
real    0m20.811s                                                                                                            
user    0m1.367s                                                                                                             
sys     0m0.361s
```


Junchao-Mellanox (approved):

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -119,6 +145,21 @@ def __init__(self, xcvr_eeprom):
         self.vdm = CmisVdmApi(xcvr_eeprom) if not self.is_flat_memory() else None
         self.cdb = CmisCdbApi(xcvr_eeprom) if not self.is_flat_memory() else None
 
+    def clear_cache(self, method_name=None):
```
prgeor, line 148: @jianyuewu seems not used. let the cache clear automatically when Xcvrd restart.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -109,6 +109,32 @@ class VdmSubtypeIndex(Enum):
...
+def read_only_cache_dict_and_list(func):
+    cache_name = f'_{func.__name__}_cache'
+    def wrapper(self):
+        if not hasattr(self, cache_name):
+            cache_value = func(self)
+            if cache_value is not None and isinstance(cache_value, (dict, list)) and len(cache_value) > 0:
+                setattr(self, cache_name, cache_value)
+            return cache_value
+        return getattr(self, cache_name)
+    return wrapper
```
prgeor, line 136: @jianyuewu can we have only one function implementation for cached value be it return value from collections or otherwise? i.e just need `read_only_cached_api_return()` ?

sonic_platform_base/sonic_xcvr/api/public/cmis.py
prgeor, line 1: @jianyuewu can you move this code under so that it can be reused for Sfp, QSFP+, QSFP28 if needed in future

https://github.com/sonic-net/sonic-platform-common/tree/master/sonic_platform_base/sonic_xcvr

./utils/cache.py

sonic_platform_base/sonic_xcvr/api/public/cmis.py
jianyuewu: Combined cache function will have more complicated logic, I can try it.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
jianyuewu: Sure, I'll move it.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
jianyuewu: UT need this clear_cache😊 Otherwise, next test of same function with different output will fail. Though in offical code, cache is in CmisApi, when xcvrd process restart or recreate CmisApi object, old instance and cache will be deleted.

prgeor: @jianyuewu you can still move this to test code

jianyuewu: Agree, I can put it in test code, will move it.

keboliu (dismissed):

mihirpat1 (commented):
@jianyuewu Where are we clearing the cache?

jianyuewu:
> @jianyuewu Where are we clearing the cache?

No need explicitly clear the cache, as it will clear automatically when Xcvrd restart.

liat-grozovik:
@mihirpat1 are you OK to approve and merge this PR?
@jianyuewu i added also 202411 as i beleive this is needed to any system supports cmis host management. if you dont think please remove the label

mihirpat1:
> > @jianyuewu Where are we clearing the cache?
> 
> No need explicitly clear the cache, as it will clear automatically when Xcvrd restart.

@jianyuewu In case of SFP removal, shouldn't we delete the cache to ensure that the transceiver specific information is deleted?

jianyuewu:
> > > @jianyuewu Where are we clearing the cache?
> > 
> > 
> > No need explicitly clear the cache, as it will clear automatically when Xcvrd restart.
> 
> @jianyuewu In case of SFP removal, shouldn't we delete the cache to ensure that the transceiver specific information is deleted?

OK, agree, this case, we should delete cache firstly.

prgeor (commented):
@mihirpat1 please check if the api gets reinitialized on OIR for other platforms, else on new OIR we will be using the old cache

jianyuewu:
@prgeor @mihirpat1 BTW, which place is suitable to clear_cache()? Now I added to reset_module() API, not sure if suitable or not. Thank you~

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1290,6 +1310,8 @@ def reset_module(self, reset = False):
         Return True if no action.
         '''
         if reset:
+            # Clear cached values when module is reset (module removal event)
```
prgeor, line 1313: @jianyuewu there is no guarantee platforms has to call this reset_module() API on OIR

jianyuewu: @prgeor OK, is there some suitable place to call clear_cache()? Thank you😊 I checked the code, didn't find suitable place yet, previously thought maybe put in get_xcvr_api() when it is not present, but might cost some time every call in get_xcvr_api(). Is that acceptable?

Junchao-Mellanox: Calling `get_present()` in `get_xcvr_api()` cannot guarantee the correctness. It could be the case that `get_xcvr_api()` is called after a "plug out" ---- "plug in" sequence. In that case, `get_present` returns True but cable was replaced. Some options from my side:

1. Make the cache configurable. The cache is disabled by default. We enable it on NVIDIA platform. Other vendors may choose enable/disable it by themselves.
2. Calling `clear_cache` in `chassis.get_change_event` when module is removed. Currently, only xcvrd calls `get_change_event`, so it means that other process may not detect the module plug out event and cannot clear the cache.

jianyuewu: Code updated😊

jianyuewu:
Enable cache review in https://github.com/jianyuewu/sonic-buildimage/pull/7

jianyuewu:
Code updated, added remove_xcvr_api and move clear_cache() API to UT.😊

keboliu:
@prgeor could you please review the updated code?

prgeor (commented):
@jianyuewu overall LGTM.

sonic_platform_base/sfp_base.py
```diff
@@ -479,3 +479,9 @@ def get_xcvr_api(self):
         if self._xcvr_api is None:
             self.refresh_xcvr_api()
         return self._xcvr_api
+
+    def remove_xcvr_api(self):
```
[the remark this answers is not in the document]
jianyuewu: Junchao had tested it, it is safe, code is:
```python
import random
import threading
import time
 
class SFP:
    def __init__(self, api):
        self.api = api
 
    def get_api(self):
        return self.api
 
    def reset(self):
        self.api = None
 
 
class Api:
    def __init__(self, value):
        self.value = value
 
 
g_api = Api(1000000)
sfp = SFP(g_api)
 
 
def thread_func():
    global sfp
    time.sleep(1)
    i = 0
    while True:
        api = sfp.get_api()
        if api is None:
            continue
        if i % 2 == 0:
            value = api.value
            sfp.reset() # set api object to None
            sfp.api = Api(value - 1) # re-create api object
            if sfp.api.value <= 0:
                break
        else:
            print(f'{threading.get_ident()}: {api.value}: {id(api)}')
        i += 1
 
 
def thread_func1():
    global sfp
    time.sleep(1)
    i = 0
    while True:
        api = sfp.get_api()
        if api is None:
            continue
        if i % 2 == 0:
            value = api.value
            sfp.reset() # set api object to None
            time.sleep(10)
            sfp.api = Api(value - 1) # re-create api object
            if sfp.api.value <= 0:
                break
        else:
            print(f'{threading.get_ident()}: {api.value}: {id(api)}')
        i += 1
 
def thread_func2():
    global sfp
    api = sfp.get_api()
    time.sleep(10)
    sfp.reset()
 
def thread_func3():
    global sfp
    api = sfp.get_api()
    time.sleep(10)
    for i in range(60):
        print(api.value)
 
 
t1 = threading.Thread(target=thread_func2)
t2 = threading.Thread(target=thread_func3)
 
t1.start()
t2.start()
t1.join()
t2.join()
```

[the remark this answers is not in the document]
prgeor: @Junchao-Mellanox @jianyuewu please try this program. In you are not introducing race condition. Here is an example

```
 cat race.py
import random
import threading
import time

class SFP:
    def __init__(self, api):
        self.api = api

    def get_api(self):
        return self.api

    def reset(self):
        self.api = None


class Api:
    def __init__(self, value):
        self.value = value


g_api = Api(1000000)
sfp = SFP(g_api)


def thread_func1():
    global sfp
    i = 0
    while True:
        api = sfp.get_api()
        if api is None:
            break

        time.sleep(5)
        sfp.reset() # set api object to None

def thread_func2():
    global sfp
    i = 1

    while True:
        api = sfp.get_api()
        if api is None:
            break

        time.sleep(1)
        sfp.api.value = sfp.api.value + 1

t1 = threading.Thread(target=thread_func1)
t2 = threading.Thread(target=thread_func2)

t1.start()
t2.start()
t1.join()
t2.join()
```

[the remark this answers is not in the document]
prgeor: Its OK the thread to stop but not crash like below:-

```
prgeor@prgeor-dev-vm:~/scripts$ python3 race.py
Exception in thread Thread-2:
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "race.py", line 46, in thread_func2
    sfp.api.value = sfp.api.value + 1
AttributeError: 'NoneType' object has no attribute 'value'
prgeor@prgeor-dev-vm:~/scripts$
```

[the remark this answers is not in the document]
jianyuewu: > @Junchao-Mellanox @jianyuewu please try this program. In your program you are not introducing race condition. Here is an example
> 
> ```
>  cat race.py
> import random
> import threading
> import time
> 
> class SFP:
>     def __init__(self, api):
>         self.api = api
> 
>     def get_api(self):
>         return self.api
> 
>     def reset(self):
>         self.api = None
> 
> 
> class Api:
>     def __init__(self, value):
>         self.value = value
> 
> 
> g_api = Api(1000000)
> sfp = SFP(g_api)
> 
> 
> def thread_func1():
>     global sfp
>     i = 0
>     while True:
>         api = sfp.get_api()
>         if api is None:
>             break
> 
>         time.sleep(5)
>         sfp.reset() # set api object to None
> 
> def thread_func2():
>     global sfp
>     i = 1
> 
>     while True:
>         api = sfp.get_api()
>         if api is None:
>             break
> 
>         time.sleep(1)
>         sfp.api.value = sfp.api.value + 1
> 
> t1 = threading.Thread(target=thread_func1)
> t2 = threading.Thread(target=thread_func2)
> 
> t1.start()
> t2.start()
> t1.join()
> t2.join()
> ```

Hi @prgeor Looks sfp.api.value should be api.value, after test, there is no issue. Because we are using api in local, instead of global in official code.
```diff
diff race_test.py race.py
46c46
<         api.value = api.value + 1
---
>         sfp.api.value = sfp.api.value + 1
54c54
```
I also tried if use global sfp.api.value, then issue will be:

```bash
# python3 race_test.py
Exception in thread Thread-2 (thread_func2):
  File "/usr/lib/python3.11/threading.py", line 1038, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.11/threading.py", line 975, in run
    self._target(*self._args, **self._kwargs)
  File "//race_test.py", line 46, in thread_func2
    sfp.api.value = sfp.api.value + 1
                    ^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'value'
```

jianyuewu:
Some problematic cable scenarios:
If user could have issued reset since some fields in EEPROM were not accessible initially (the cable has some problem in this case, after re-plugin, then data will be correct.) In this case, with cache enabled, there is a WA: we can reset the sfp by **xcvrd restart**, then data will be correct.

keboliu (dismissed):

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1,4 +1,3 @@
-
```
mihirpat1, line 1: @jianyuewu Please revert this change

jianyuewu: Thanks, this change reverted.

keboliu (approved):

jianyuewu:
> hi @jianyuewu , do you mind to help create manual cherry-pick PR to 202412?

Yes, the PR is https://github.com/Azure/sonic-platform-common.msft/pull/86

PR #570, opened 2025-05-20
Correct bit range for ACTIVE_APSEL_CODE

infn-ke:
AppSelCode in cmis uses bits 4-7, range api in python is not inclusive.

#### Description
Fix range for ACTIVE_APSEL_HOSTLANE.

#### Motivation and Context
Reporting op active apsel is wrong for applications larger than 8.

Reading active apsel for when apsel is larger than 8.


prgeor:
@r12f @bingwang-ms can you take this bug fix to tagged branches?

PR #594, opened 2025-09-09
[Smartswitch] Restart sensord with lock

gpunathilell:
#### Motivation and Context

On startup of pmon in dark mode on switch, there are four processes spawned to power off DPUs, if all are executed in parallel there are four sensord processes running (since service restart commands are not queued). Since we perform handle_sensor_removal on dark mode start, the solution is to have a file lock to make sure only one module has control to restart sensord each time, queueing the sensord restart requests so that we have a single process running always, irrespective of number of restarts

Check the number of sensord processes starting before fix:
```
docker exec -it pmon ps -x|grep sensord
    170 ?        Ss     0:00 /usr/sbin/sensord -f daemon
    171 ?        Ss     0:00 /usr/sbin/sensord -f daemon
    172 ?        Ss     0:00 /usr/sbin/sensord -f daemon
    173 ?        Ss     0:00 /usr/sbin/sensord -f daemon
```
after Addition of fix, only one process is found
```
root@sonic:/home/admin# docker exec -it pmon ps -x|grep sensord
    189 ?        Ss     0:00 /usr/sbin/sensord -f daemon
```

### DIFF
--- a/sonic_platform_base/module_base.py
+++ b/sonic_platform_base/module_base.py
@@ -27,6 +27,7 @@ class ModuleBase(device_base.DeviceBase):
     # Device type definition. Note, this is a constant.
     DEVICE_TYPE = "module"
     PCI_OPERATION_LOCK_FILE_PATH = "/var/lock/{}_pci.lock"
+    SENSORD_OPERATION_LOCK_FILE_PATH = "/var/lock/sensord.lock"
 
     # Possible card types for modular chassis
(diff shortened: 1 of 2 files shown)


vvolam (approved):
LGTM, Thank you

rameshraghupathy (approved):

vvolam:
@judyjoseph could you help with merge this PR?

judyjoseph (approved):

PR #595, opened 2025-09-09
Fix health check for SSD vendors: add a parser for ATP, and add a generic health ID for other brands

markx-arista:
#### Description

Fix health check for SSD vendors: add a parser for ATP, and add a generic health ID for other brands.
Each vendor stores health information in different SMART attributes.
ATP stores it in attribute ID 248, we add a parser for it.
We also have SSDs use Attribute ID 231 and it is commonly used, so add it in the generic parser.
Skip obtaining vendor SSD info for ATP and Virtium NVMe SSD because they are handle by parse_generic_ssd_info and parse_vendor_ssd_info will overwrite data with N/A.
Add unit test cases for ATP SATA/NVMe SSD.

#### Motivation and Context

`show platform ssdhealth` shows N/A health for some qualified SSDs.

#### Back port request
- [x] 202412
- [x] 202505

We have tested the code change on DUTs with different SSDs including all the qualified SSDs that show N/A in health and also on the ones that worked fine before.

### DIFF
--- a/sonic_platform_base/sonic_storage/ssd.py
+++ b/sonic_platform_base/sonic_storage/ssd.py
@@ -103,12 +103,12 @@ class SsdUtil(StorageCommon):
 
         # Known vendor part
         if self.model:
-            # For some Virtium SSDs, parse_generic_ssd_info should be called.
-            # Since it was called above, no need to parse a specific vendor SSD info.
-            if self.model in ['Virtium VTPM24CEXI080-BM110006']:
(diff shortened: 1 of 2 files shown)


markx-arista:
> Hi @markx-arista I noticed this failure only on 7050cx3, why we need to cherrypick to 202412?

Hi, there were failures on our 202412, but we just found out those SSDs are not qualified. So PR to 202412 is not necessary, I will close it.

PR #599, opened 2025-09-23
Add API to check if module is Linear Pluggable Optics (LPO)

longhuan-cisco:
#### Description

Add `is_lpo` API to check if the module is Linear Pluggable Optics (LPO)

#### Motivation and Context

This API will be consumed by https://github.com/sonic-net/sonic-platform-daemons/pull/687

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -111,6 +111,20 @@ class CmisApi(XcvrApi):
     LowPwrRequestSW = 4
     LowPwrAllowRequestHW = 6
 
+    LPO_HOST_ELECTRICAL_INTERFACE_IDS = [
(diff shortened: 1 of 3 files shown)


sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -111,6 +111,20 @@ class CmisApi(XcvrApi):
     LowPwrRequestSW = 4
     LowPwrAllowRequestHW = 6
 
+    LPO_HOST_ELECTRICAL_INTERFACE_IDS = [
```
prgeor, line 114: @longhuan-cisco which SFF spec version are you using?

mihirpat1: @longhuan-cisco @prgeor I corrected the LPO related IDs since this changes was causing test failures for https://github.com/sonic-net/sonic-platform-common/pull/596

PR #604, opened 2025-10-06
[Smartswitch] Host container differentiation for sensor commands

gpunathilell:
#### Description

Differentiate commands being executed in host and container seperately for some smartswitch specific functions

#### Motivation and Context

This is done because these functions are called from inside the pmon docker during chassis command execution and from the host directly during reboot command execution


gpunathilell:
@vvolam and @rameshraghupathy please review

vvolam (approved):

PR #637, opened 2026-03-20
Implement leak detection common components

fraserg-arista:
#### Description

This implements the leak detection parts of the pmon design as described in https://github.com/sonic-net/SONiC/pull/2215.

### DIFF
--- a/sonic_platform_base/liquid_cooling_base.py
+++ b/sonic_platform_base/liquid_cooling_base.py
@@ -5,7 +5,10 @@
     to interact with a liquid cooling module in SONiC
 """
 
+from abc import ABC, abstractmethod
 from enum import Enum
+from typing import Dict, List
+
 from . import device_base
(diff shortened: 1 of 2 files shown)


judyjoseph (approved):
LGTM

yijingyan2:
Hi @fraserg-arista, could you please check the build failure https://dev.azure.com/mssonic/build/_build/results?buildId=1101023&view=results in submodule advanced PR https://github.com/sonic-net/sonic-buildimage/pull/26986. seems like the build failure is related with your change. Can you please take a look? Thanks

fraserg-arista:
There's a bit of a circular dependency here - I updated the constructor on a class, which means the existing tests that use it will fail. But the PR that fixes the tests (https://github.com/sonic-net/sonic-platform-daemons/pull/776) needs the updated constructor from this PR.

PR #640, opened 2026-03-24
Add complete bank support for CMIS transceivers

bobby-nexthop:
#### Description

This PR extends the bank parameter support introduced in PR #632 to fully enable multi-bank CMIS transceiver support throughout the stack.

Changes include:

1. **Memory Map Bank Awareness**:
   - Updated CmisFlatMemMap, CmisMemMap, and CCmisMemMap to accept bank parameter
   - Implemented bank-aware getaddr() method in CmisFlatMemMap that correctly calculates linear offsets using the formula from sonic-linux-kernel PR
   - Added bank property to CmisFlatMemMap for read-only access
   - Added constants: CMIS_EEPROM_PAGE_SIZE, CMIS_NUM_NON_BANKED_PAGES, CMIS_NUM_BANKED_PAGES, CMIS_ARCH_PAGES
   - Uses unified formula: linear_offset = (bank * CMIS_ARCH_PAGES + page) * page_size + offset
   - Each bank is treated as a full 256-page (32KB) architectural block for proper alignment

2. **API Factory Updates**:
   - Added bank parameter to create_xcvr_api() and _create_cmis_api()
   - Updated CMIS API instantiation to pass bank parameter to memory maps
   - Modified id_mapping to pass bank parameter for all CMIS module types (0x18, 0x19, 0x1b, 0x1e)

3. **SfpBase Integration**:
   - Updated refresh_xcvr_api() to pass bank parameter when creating xcvr_api

4. **Field Definitions**:
   - Added BANKS_SUPPORTED_FIELD constant to consts.py
   - Added BanksSupported field to MODULE_CHAR_ADVT in CmisMemMap

This enables proper support for CMIS modules with multiple memory banks, allowing correct EEPROM access across all banks as defined in the CMIS specification. The offset calculation formula matches the optoe kernel driver implementation in sonic-linux-kernel PR #473.

#### Motivation and Context

We are implementing banking support as allowed by the CMIS spec. Our HLD is available at https://github.com/sonic-net/SONiC/pull/2183

I don't have banked hardware yet to test on, have confirmed that these changes are backwards compatible and don't break existing systems


Natgerbi:
@bobby-nexthop, is this PR still relevant? I see that you also have PR https://github.com/sonic-net/SONiC/pull/2183, which seems to be more maintained and includes the paging part that does not exist here.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
```diff
@@ -19,11 +19,18 @@
 from ...fields.consts import *
 from ...fields.public.cmis import CableLenField
 
+# Constants matching optoe driver
+CMIS_EEPROM_PAGE_SIZE = 128
+CMIS_NUM_NON_BANKED_PAGES = 16   # pages 00h-0Fh
```
prgeor, line 24: @bobby-nexthop can we add a unit test to cover a scenario to ensure that NONE of the EEPROM memory map field mistakenly does not set bank >=1 for Unbanked Pages.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
```diff
@@ -138,12 +145,41 @@ def __init__(self, codes):
...
+        For lower memory (page 0, offset < 128):
+            linear_offset = offset
+
+        For paged memory:
+            offset_in_paged_area = (page * page_size + offset) - 128
+            bytes_per_bank = CMIS_ARCH_PAGES * page_size  (256 * 128 = 32KB)
+            linear_offset = 128 + (bank * bytes_per_bank) + offset_in_paged_area
+
+        Simplified:
+            linear_offset = (bank * CMIS_ARCH_PAGES + page) * page_size + offset
```
prgeor, line 169: @bobby-nexthop will be good to put an ASSERT here to make sure none of the memmap fields mistakenly access NON-banked pages. And in the unit test we can exercise this. This way we ensure this requirement at single place.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
```diff
@@ -138,12 +145,41 @@ def __init__(self, codes):
...
+        Note: Each bank is treated as a full 256-page (32KB) architectural block,
+        even though only pages 10h-FFh (240 pages) are actually banked. This ensures
+        proper alignment and matches the kernel driver behavior.
+        """
+        if page == 0 and offset < 128:
+            # Lower memory - not affected by banking
+            return offset
+
+        # For all paged memory (including bank 0), use the unified formula
+        # that treats each bank as a 256-page (32KB) block
```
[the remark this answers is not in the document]
bgallagher-nexthop: This suggestion is invalid, since the addressing approach of treating all upper memory as if it were banked is intentional.

abelamit:
> @bobby-nexthop, is this PR still relevant? I see that you also have PR [sonic-net/SONiC#2183](https://github.com/sonic-net/SONiC/pull/2183), which seems to be more maintained and includes the paging part that does not exist here.

hi @bobby-nexthop - can you please comment if you plan to rebase this commit on top of the other commit proposed? It seems like they are not correlated.

prgeor:
> @bobby-nexthop, is this PR still relevant? I see that you also have PR [sonic-net/SONiC#2183](https://github.com/sonic-net/SONiC/pull/2183), which seems to be more maintained and includes the paging part that does not exist here.

@Natgerbi can you be more specific? feel free to quote the ref section or snapshot here

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
```diff
@@ -138,12 +145,41 @@ def __init__(self, codes):
...
+    def bank(self):
+        """Returns the bank number (read-only)."""
+        return self._bank
+
     def getaddr(self, page, offset, page_size=128):
-        return page * page_size + offset
+        """
+        Calculate linear offset for optoe driver using instance's bank.
+
+        For lower memory (page 0, offset < 128):
```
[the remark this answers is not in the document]
aditya-nexthop: test added

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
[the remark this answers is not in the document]
bgallagher-nexthop: I've actually addressed this by clamping the bank parameter to 0 for non-banked pages. This will result in us writing 0 to the bank select register when accessing non-banked pages.

sonic_platform_base/sonic_xcvr/xcvr_api_factory.py
```diff
@@ -86,10 +86,10 @@ def _create_cmis_api(self):
              ('EOPTOLINK' in vendor_name and vendor_pn in EOP_800G_VENDOR_PN_LIST):
             api = self._create_api(CmisCodes, CmisMemMap, CmisFr800gApi)
         else:
-            xcvr_eeprom = XcvrEeprom(self.reader, self.writer, CmisMemMap(CmisCodes))
+            xcvr_eeprom = XcvrEeprom(self.reader, self.writer, CmisMemMap(CmisCodes, bank=bank))
```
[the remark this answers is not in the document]
bgallagher-nexthop: Fixed.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
bgallagher-nexthop: - I have changed getaddr() to clamp bank to 0 when accessing unbanked pages.
- I have added unit-tests that ensure we calculate the same offset regardless of what bank is set to.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
bgallagher-nexthop: I think the new unit-tests provide coverage to prevent incorrect offset calculation. Let me know if I am missing anything.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
```diff
@@ -19,11 +19,18 @@
 from ...fields.consts import *
 from ...fields.public.cmis import CableLenField
 
+# Constants matching optoe driver
+CMIS_EEPROM_PAGE_SIZE = 128
+CMIS_NUM_NON_BANKED_PAGES = 16   # pages 00h-0Fh
+CMIS_NUM_BANKED_PAGES = 240      # pages 10h-FFh
+CMIS_ARCH_PAGES = 256            # architectural pages per bank (matches OPTOE_ARCH_PAGES)
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
```diff
@@ -138,12 +144,53 @@ def __init__(self, codes):
...
+        even though only pages 10h-FFh (240 pages) are actually banked. This ensures
+        proper alignment and matches the kernel driver behavior.
+        """
+        if page == 0 and offset < 128:
+            # Lower memory - not affected by banking or paging.
+            return offset
+
+        # If we are accessing a non-banked page, there is no reason to set the bank
+        # to a non-zero value. 
+        bank = 0 if page < CMIS_NUM_NON_BANKED_PAGES else self.bank
```
prgeor, line 181: @Natgerbi @abelamit please note this TODO for your platform.

sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
```diff
@@ -303,6 +312,36 @@ def set_optoe_write_max(self, write_max):
         except (OSError, IOError):
             pass
 
+    def set_optoe_max_bank_size(self, max_bank_size):
```
micas-net, line 315: why bank size is set in optoe, we should move it into chassis. the scope of sfp is within one bank, not the whole eeprom

sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
```diff
@@ -303,6 +312,36 @@ def set_optoe_write_max(self, write_max):
...
+        """Write max_bank_size to the optoe driver's sysfs entry; required before banked EEPROM offsets are accessible"""
+        try:
+            sys_path = self.get_eeprom_path().replace("eeprom", "max_bank_size")
+            with open(sys_path, mode='w') as f:
+                f.write(str(max_bank_size))
+        except (OSError, IOError):
+            pass
+
+    def _read_optoe_max_bank_size(self):
```
micas-net, line 324: should also move to chassis if necessary

sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
aditya-nexthop: Yes, however, the API and the ability to access the EEPROM and detect module removal/insertion is at a per-bank SFP object level. We do ensure read-before-write semantics to ensure we don't write `max_bank_size` if it is already correct to prevent repeated writes. Do you have any recommendations on where we could put it in the chassis and how we could set it and also handle setting it upon transceiver removal/insertion?

bgallagher-nexthop: While each SFP object is associated with a single bank, I chose this approach for the following reasons:
- the write needs to be triggered again in the event of a new transceiver being plugged in with a different number of banks than the old one. Housing this logic in `refresh_xcvr_api` gives us that behaviour for free. If it was in the Chassis code, this would not be the case.
- writing to max_bank_size is an implementation detail specific to the optoe driver, so it is nice to keep this logic in optoe-specific code.

The change has been implemented with read-before-write idempotency, so the following sequence of events would occur if a Chassis created 4 Sfp objects for a 4-banked transceiver:
- Sfp object with bank=0 would do nothing (this is so unbanked transceivers continue to operate as they previously did before banking support was introduced)
- Sfp object with bank=1 will perform the write to max_bank_size.
- Sfp object with bank=2 will see the value has already been written and do nothing.
- Sfp object with bank=3 will see the value has already been written and do nothing.

We will ultimately only write to the max_bank_size sysfs path once, while still keeping this optoe specific quirk internal to SfpOptoeBase.

sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
bgallagher-nexthop: Same comment as above.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py
bgallagher-nexthop: Screenshot 2026-05-15 at 1 33 57 PM

For reference, this is the relevant part of the CMIS spec this TODO is referring to (2 CDB instances are possible to support, accessible via bank selection).

Natgerbi:
> see that you also have PR [sonic-net/SONiC#2183](https://github.com/sonic-net/SONiC/pull/2183), which seems to be more maintained and includes the paging part that does not exist

What I meant is that section 7.5 of the HLD, “Memory Map Abstraction Changes (cmis.py),” already aligns with the new abstraction layer based on memory-map pages. However, I think the actual implementation of the page abstraction can be added later in a separate PR.

micas-net (approved):
Seems good to me!

sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
```diff
@@ -303,6 +312,36 @@ def set_optoe_write_max(self, write_max):
...
+        sys_path = self.get_eeprom_path().replace("eeprom", "max_bank_size")
+        with open(sys_path) as f:
+            if int(f.read().strip()) == max_bank_size:
+                return
+        with open(sys_path, mode='w') as f:
+            f.write(str(max_bank_size))
+
+    def _read_optoe_max_bank_size(self):
+        """Determine optoe max_bank_size from the module's CMIS BanksSupported advertisement, or None if non-CMIS, flat-memory, or unreadable."""
+        id_byte = self.read_eeprom(0, 1)
```
prgeor, line 326: @bgallagher-nexthop the optoe API here assumes its accessing CMIS memory map even before API is refreshed. 
TODO: Refresh the API and then set the max bank if the API supports it.

prgeor (approved):

PR #641, opened 2026-03-24
Add support for get_power_override_support in sfp API

arista-hpandya:
#### Description

This PR introduces the get_power_override_support method to the SFP API within sonic-platform-common. Needed to address: https://github.com/sonic-net/sonic-buildimage/issues/26372

sfp_base.py: Added the abstract method definition to the SfpBase class to define the interface.

sfp_optoe_base.py: Provided an implementation in SfpOptoeBase that delegates the call to the underlying transceiver API (xcvr_api).

This allows platform-specific implementations and upper-layer applications to query whether a specific SFP module supports power override capabilities before attempting to get or set override values.

#### Motivation and Context

Currently, while get_power_override and set_power_override exist, there isn't a standardized way to check for feature support across different hardware modules. This change improves the robustness of the SFP interface by allowing callers to verify support programmatically, preventing unnecessary errors or invalid calls on modules that do not support this feature.

This will help us move away from hardcoded functions like these that need constant maintenance: https://github.com/sonic-net/sonic-mgmt/blob/a1d1fd8717c4525055a3a4a9e3250232ee3b0d86/tests/platform_tests/api/test_sfp.py#L374-L378

Verified by running the test_sfp sonic-mgmt tests to use this API

NA

### DIFF
--- a/sonic_platform_base/sfp_base.py
+++ b/sonic_platform_base/sfp_base.py
@@ -290,6 +290,16 @@ class SfpBase(device_base.DeviceBase, EepromReadWriteMixin):
         """
         raise NotImplementedError
 
+    def get_power_override_support(self):
+        """
+        Retrieves whether power override is supported by this SFP
+
+        Returns:
+            bool: True if power override is supported, False if not supported
+            None: if not implemented or xcvr_api is not available
+        """
+        return None
+
     def get_temperature(self):
         """
         Retrieves the temperature of this SFP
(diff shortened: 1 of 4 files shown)


prgeor (dismissed):

prgeor:
@arista-hpandya code coverage missing in the pr checked

deepak-singhal0408:
@arista-hpandya please help address code-coverage. thanks,

arista-hpandya:
Added test coverage, can I get approval for this? @prgeor @deepak-singhal0408

prgeor (approved):

PR #646, opened 2026-03-25
[agent][ssd] Fix NameError: nand_endurance undefined in parse_micron_info()

rustiqly:
#### What I did
Fix `NameError` in `parse_micron_info()` where `nand_endurance` is referenced but never defined.

#### How I did it
- Parse `nand_endurance` from SMART attributes using `_parse_re()` with the same `NAND_Endurance` regex pattern used in `parse_virtium_info()`
- Guard the health calculation with `nand_endurance != NOT_AVAILABLE` check
- Remove unused `erase_fail_count` variable (was fetched but never used in the calculation)

#### How to verify it
Run on a Micron SSD where `Percent_Lifetime_Used` and `Percent_Lifetime_Remain` are both unavailable — should no longer crash with `NameError`.

#### Which release branch to backport
master

#### Description for the changelog
Fix NameError crash in Micron SSD health parsing due to undefined nand_endurance variable.

Fixes: #644

### DIFF
--- a/sonic_platform_base/sonic_storage/ssd.py
+++ b/sonic_platform_base/sonic_storage/ssd.py
@@ -338,9 +338,9 @@ class SsdUtil(StorageCommon):
 
             if health_raw == NOT_AVAILABLE:
                 average_erase_count = self.parse_id_number(MICRON_AVG_ERASE_COUNT_ID, self.vendor_ssd_info)
-                erase_fail_count = self.parse_id_number(MICRON_ERASE_FAIL_COUNT_ID, self.vendor_ssd_info)
+                nand_endurance = self._parse_re(r'NAND_Endurance\s*\d*\s*(\d+?)\s+', self.vendor_ssd_info)
 
-                if average_erase_count != NOT_AVAILABLE and erase_fail_count != NOT_AVAILABLE:
+                if average_erase_count != NOT_AVAILABLE and nand_endurance != NOT_AVAILABLE:
                     try:
                         self.health = 100 - (float(average_erase_count) * 100 / float(nand_endurance))
                     except (ValueError, ZeroDivisionError) as ex:
--- a/sonic_platform_base/sonic_storage/ssd.py
+++ b/sonic_platform_base/sonic_storage/ssd.py
@@ -275,8 +275,8 @@ class SsdUtil(StorageCommon):
                 health_raw = self.parse_id_number(VIRTIUM_HEALTH_ID, self.vendor_ssd_info)
                 self.health = float(health_raw.split()[2]) if health_raw != NOT_AVAILABLE else NOT_AVAILABLE
             else :
-                nand_endurance = self._parse_re(r'(?m)NAND_Endurance.*?(\d+)\s*$', self.vendor_ssd_info)
-                avg_erase_count = self._parse_re(r'(?m)Average_Erase_Count.*?(\d+)\s*$', self.vendor_ssd_info)
+                nand_endurance = self._parse_re(r'NAND_Endurance\s*\d*\s*(\d+?)\s+', self.vendor_ssd_info)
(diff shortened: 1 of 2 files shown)


yxieca:
The nand_endurance fix is correct — it was indeed undefined and would crash. However, I am not sure about removing the erase_fail_count check from the guard condition.

The original code checked both average_erase_count and erase_fail_count before computing health. While erase_fail_count is not used in the formula, it may have been an intentional sanity gate — i.e., do not report health percentage if we cannot read erase failure data from the drive.

Suggestion: keep erase_fail_count in the guard and add nand_endurance:

```python
if average_erase_count != NOT_AVAILABLE and erase_fail_count != NOT_AVAILABLE and nand_endurance != NOT_AVAILABLE:
```

This way we fix the NameError without changing the original gating logic. Also, erase_fail_count should be kept as a parsed variable even if unused in the formula — removing it changes the behavior silently.

---
*This comment was posted by an AI agent on behalf of @yxieca.*

yxieca:
@rustiqly  please consider adding unit test.

rustiqly:
@yxieca acknowledged, I will work on adding the unit test coverage.

rustiqly:
Added Micron SSD unit coverage in `69fa5f7` for the NAND endurance fallback path, covering the case where `Percent_Lifetime_Used` is unavailable and health is computed from average erase count + NAND endurance.\n\nLocal test note: `pytest -q -o addopts='' tests/test_ssd.py -k micron_ssd` is blocked on this host by missing `sonic_py_common` dependency during test collection.

rustiqly:
Pushed follow-up `c589d68` for https://github.com/sonic-net/sonic-platform-common/pull/646 to fix the Azure unit-test failure.

Root cause from CI:

```text
test_micron_ssd_health_from_nand_endurance
AssertionError: assert 'N/A' == 75.0
```

The new fallback path was passing the full parsed SMART attribute tail into `float()` instead of extracting the raw value column. The fix now uses `average_erase_count.split()[-1]` and `nand_endurance.split()[-1]` before computing health.

Validation on this host:

```text
python3 -m py_compile sonic_platform_base/sonic_storage/ssd.py
```

The targeted pytest collection is still blocked locally by missing `sonic_py_common`, but this addresses the exact assertion reported by Azure.

rustiqly:
Pushed follow-up `c4bfc51` for https://github.com/sonic-net/sonic-platform-common/pull/646 to fix the Azure unit failure.

Root cause: `NAND_Endurance` in the SMART attribute table appears as a normal SMART row (`999 NAND_Endurance ... 3028`). The previous regex only handled a bare label/value form and did not skip the SMART row fields, so the fallback left health as `N/A`.

Change: parse the final numeric field from the SMART row for `NAND_Endurance` (and the analogous Virtium `Average_Erase_Count` path).

Validation:
- `python3 -m py_compile sonic_platform_base/sonic_storage/ssd.py`
- Standalone regex verification against SMART rows for `NAND_Endurance=3028` and `Average_Erase_Count=757`

Local pytest is blocked on this host because `sonic_py_common` is not installed in the system Python environment; Azure has the repo test environment and should now exercise `tests/test_ssd.py::TestSsd::test_micron_ssd_health_from_nand_endurance`.

/azp run

rustiqly:
Pushed follow-up `2ff5458` for https://github.com/sonic-net/sonic-platform-common/pull/646 to fix the new Azure Python unit failure.

Root cause: the previous NAND_Endurance regex change accidentally altered the existing Virtium parser to read the last column (`0`) instead of the endurance raw value (`20000`), causing the health fallback to return `N/A`.

Fix: restored the Virtium-specific raw-value parsing and kept the Micron NAND_Endurance fallback as a separate SMART-table parse.

Validation: confirmed the regex extraction for both Virtium (`20000`) and Micron (`3028`) sample rows locally. Full pytest could not run on this host because the local environment is missing `sonic_py_common`; Azure will run the complete test suite.

PR #655, opened 2026-04-15
Add domain-aware PCIe handling and improve pcie_common robustness

t2sharma:
Description:
This PR enhances pcie_common.py to support domain-aware PCIe enumeration and improves overall robustness and safety of the utility.

Problem:
The current upstream implementation:

Assumes PCI domain is implicitly 0000
Uses non-domain-aware lspci output
Uses unsafe YAML parsing (yaml.load)
Does not handle duplicate PCIe entries
Lacks visibility into PCIe AER error statistics

This leads to incorrect PCIe detection on platforms where:

PCI domains are non-zero
Multiple devices share similar BDF across domains

Changes:
1. Domain-aware PCIe enumeration
Switched to:
lspci -D
lspci -D -n
Parses full BDF: domain:bus:device.function

2. Added domain support in validation
get_pcie_check() now validates using:
domain + bus + device + function
Backward compatible (defaults to "0000")

3. Safer YAML parsing:
yaml.safe_load()

4. Duplicate PCIe filtering:
Avoids duplicate entries using (domain, bus, dev, fn) key

5. Vendor ID capture:
Adds vendor field for better device identification

6. Improved subprocess handling
Checks return codes
Captures stderr

7. AER stats support (new feature)
Added:
get_pcie_aer_stats()
Reads from:
/sys/bus/pci/devices/.../aer_dev_*

8. Config file improvements
Supports revision-based configs:
pcie_<rev>.yaml
Preserves YAML key order (sort_keys=False)
Backward Compatibility
Existing configs without domain continue to work (default = 0000)
No breaking change to existing workflows
Existing CLI behavior unchanged
Validation
Verified PCIe detection on Arctos platform
Compared output with:
lspci -D

Confirmed:
Correct domain parsing
No duplicate entries
Accurate sysfs validation

Impact:
Enables correct PCIe validation on multi-domain platforms
Improves debugging visibility (AER stats)
Makes code safer and more maintainable

### DIFF
--- a/sonic_platform_base/sonic_pcie/pcie_common.py
+++ b/sonic_platform_base/sonic_pcie/pcie_common.py
@@ -35,49 +35,70 @@ class PcieUtil(PcieBase):
 
     # load current PCIe device
     def get_pcie_device(self):
-        pciDict = {}
         pciList = []
-        p1 = "^(\w+):(\w+)\.(\w)\s(.*)\s*\(*.*\)*"
-        p2 = "^.*:.*:.*:(\w+)\s*\(*.*\)*"
-        command1 = ["sudo", "lspci"]
-        command2 = ["sudo", "lspci", "-n"]
-        # run command 1
-        proc1 = subprocess.Popen(command1, universal_newlines=True, stdout=subprocess.PIPE)
-        output1 = proc1.stdout.readlines()
-        (out, err) = proc1.communicate()
-        # run command 2
-        proc2 = subprocess.Popen(command2, universal_newlines=True, stdout=subprocess.PIPE)
-        output2 = proc2.stdout.readlines()
-        (out, err) = proc2.communicate()
+        seen = set()
 
-        if proc1.returncode > 0:
-            for line1 in output1:
-                print(line1.strip())
-            return
-        elif proc2.returncode > 0:
-            for line2 in output2:
-                print(line2.strip())
-            return
-        else:
-            for (line1, line2) in zip(output1, output2):
-                pciDict.clear()
-                match1 = re.search(p1, line1.strip())
-                match2 = re.search(p2, line2.strip())
-                if match1 and match2:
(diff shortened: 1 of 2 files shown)


bmridul:
Pls add a sample UT log.

sonic_platform_base/sonic_pcie/pcie_common.py
```diff
@@ -35,49 +35,70 @@ def load_config_file(self):
...
+        lines1 = output1.splitlines()
+        lines2 = output2.splitlines()
+
+        for line1, line2 in zip(lines1, lines2):
+            match1 = re.search(p1, line1.strip())
+            match2 = re.search(p2, line2.strip())
+            if not (match1 and match2):
+                print("CAN NOT MATCH PCIe DEVICE")
+                print("lspci   :", line1.strip())
+                print("lspci -n:", line2.strip())
```
bgallagher-nexthop: We are missing the new -D flag here right?
```suggestion
                print("lspci -D  :", line1.strip())
                print("lspci -D -n:", line2.strip())
```

t2sharma:
@bgallagher-nexthop  I checked the  CI log (Azure Pipelines / Azure.sonic-platform-common) in detail. The failure is in tests/pcie_common_test.py, where the mock still expects ['sudo', 'lspci'] and ['sudo', 'lspci', '-n']. The updated implementation now correctly uses domain-aware commands ['sudo', 'lspci', '-D'] and ['sudo', 'lspci', '-D', '-n'], so the mock falls through and output never gets assigned, causing the UnboundLocalError. We need to update the unit test mock to match the new commands, and also update the expected PCIe entries to include the new domain field returned by the implementation. Please help with updating of the script for Pipeline to Pass.

t2sharma:
manual ut -f
[pcie_module_UT.txt](https://github.com/user-attachments/files/27138255/pcie_module_UT.txt)

PR #662, opened 2026-05-08
Add LED and SFP methods to support new port LED policy

justin-oliver:
As described in https://github.com/sonic-net/SONiC/pull/2223, we aim to implement a new policy for controlling front panel port LEDs. This PR introduces the platform APIs required to support the LED policy.

#### Description

Add `led_base.py` with basic abstract methods for describing and controlling LEDs.

Update `sfp_base.py` with an LED list that contains LedBase derived objects representing all LEDs tied to a particular port. A platform must implement `get_all_leds` if it wishes for LED daemon (ledd) to use the new port LED policy. Otherwise it will stick to the current V1 policy.

#### Motivation and Context

The current port LED handling by ledd has several issues with ambiguous behavior, particularly around port breakout scenarios where we need deterministically to map X subports to Y port LEDs. The LED state determination is also inconsistent across different SKUs and platform vendors and doesn't account for admin status of the interface. By handling LED control directly in ledd, we can have a unified policy for all platforms.

See https://github.com/sonic-net/SONiC/pull/2223 for full details.

Added basic unit tests to cover the new APIs.

### DIFF
--- a/sonic_led/led_control_base.py
+++ b/sonic_led/led_control_base.py
@@ -10,6 +10,11 @@ try:
 except ImportError as e:
     raise ImportError (str(e) + " - required module not found")
 
+#
+# Platform API V1 (Deprecated)
+# This class is maintained for backward compatibility. New platforms should
+# implement Platform API V2 (sonic_platform_base.led_base.LedBase).
+#
 class LedControlBase(object):
     __metaclass__ = abc.ABCMeta
 
--- a/sonic_platform_base/led_base.py
+++ b/sonic_platform_base/led_base.py
@@ -1,0 +1,93 @@
+"""
+    led_base.py
+
+    Abstract base class for implementing a platform-specific class with which
+    to interact with a LED in SONiC
+"""
+
+try:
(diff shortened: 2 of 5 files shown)


prgeor (changes requested):
@justin-oliver sfp_base or any transceiver related files it NOT the right place to add LED related code. Please remove LED code from sfp_base.py

justin-oliver:
> @justin-oliver sfp_base or any transceiver related files it NOT the right place to add LED related code. Please remove LED code from sfp_base.py

This change is intended to track the LEDs tied to a specific SFP port as per https://github.com/sonic-net/SONiC/pull/2223

Would you prefer we define port LEDs at a higher level like `ChassisBase` and map them to their SFP port there? Something like
```python
        # List of SfpBase-derived objects representing all sfps
        # available on the chassis
        self._sfp_list = []
+       self._sfp_led_list = {}  # sfp: [sfp_leds]
``` 

We can arrange a meeting to discuss further if needed.

PR #665, opened 2026-05-13
Add Chassis.get_cooling_type() API for liquid/air cooling detection

shreyansh-nexthop:
#### Description

Adds a generic `Chassis.get_cooling_type()` API to ChassisBase so callers can ask "is this chassis liquid- or air-cooled?"
* New constants on ChassisBase: `COOLING_TYPE_LIQUID`, `COOLING_TYPE_AIR`, `COOLING_TYPE_UNKNOWN`.
* New virtual method get_cooling_type() — default returns `COOLING_TYPE_UNKNOWN`. Vendor Chassis subclasses override this to return the real value based on the platform signal.

#### Motivation and Context

Required by the redfish feature in `sonic-buildimage` - [PR#27349-sonic-buildimage](https://github.com/sonic-net/sonic-buildimage/pull/27349). The redfish docker should auto enable only on liquid-cooled BMC chassis.

Did manual verification by loading the image on the BMC Card.


shreyansh-nexthop:
This PR is no longer needed. This PR intended to introduce a new function `get_cooling_type` to know whether a device is liquid or air cooled, but with function `is_liquid_cooled` added via #674, it's no longer needed.

PR #667, opened 2026-05-18
Refactor CMIS memory map into pages

aditya-nexthop:
Cherry pick commits from https://github.com/nexthop-ai/sonic-platform-common/pull/1 to public fork since it wasn't possible to move the original PR to `sonic-net`

This PR implements part 1 of 2 of the HLD described [here](https://github.com/sonic-net/SONiC/pull/2207).
It introduces the CmisPage class and refactors CMIS Mem Map into pages

#### Description

Refactor the CMIS memory map from a single monolithic class into per-page
classes that each own their own field definitions.

- Introduces a new `sonic_platform_base/sonic_xcvr/mem_maps/public/cmis_pages/`
  package with one module per CMIS page:
  - `pg_00_administrative_lower.py` / `pg_00_administrative_upper.py`
  - `pg_01_advertising.py`
  - `pg_02_thresholds.py`
  - `pg_10_lane_datapath_config.py`
  - `pg_11_lane_datapath_status.py`
  - `pg_12_tunable_laser_ctrl_status.py`
  - `pg_13_module_perf_diag_ctrl.py`
  - `pg_2f_vdm_advertising_ctrl.py`
  - `pg_9f_cdb_message.py`
- Adds a `CmisPage` base class (`cmis_pages/base.py`) that:
  - Stores its own `page` and `bank` numbers.
  - Implements `getaddr(offset, page_size)` so each page computes linear
    EEPROM offsets locally instead of relying on `CmisFlatMemMap.getaddr`.
  - Provides `register_fields(memmap)`, which composes the page's fields
    onto the parent memory map and merges shared `RegGroupField`s (e.g.
    `ADMIN_INFO_FIELD`, `ADVERTISING_FIELD`) that span multiple pages,
    re-sorting members by offset to preserve the `RegGroupField` invariant.
- Adds `cmis_pages/cmis_page_consts.py` with CMIS layout constants
  (`CMIS_EEPROM_PAGE_SIZE`, `CMIS_ARCH_PAGES`,
  `CMIS_NUM_NON_BANKED_PAGES`, and page-number constants).
- Reworks `cmis.py`:
  - `CmisFlatMemMap` now exposes an `add_pages(*pages)` helper that appends
    to `self.pages` and calls `register_fields` on each page.
  - `CmisFlatMemMap` composes only the lower and upper halves of page 0x00.
  - `CmisMemMap` extends it by adding pages 0x01, 0x02, 0x10–0x13, 0x2F,
    and 0x9F.
  - `getaddr` now takes `page` as an explicit parameter and clamps `bank`
    to 0 for non-banked pages (00h–0Fh) and CDB pages (9Fh–AFh).
- Registers the new `cmis_pages` package in `setup.py`.

Net diff: `cmis.py` shrinks from ~720 lines to ~120 lines, with the field
definitions redistributed across the per-page modules.

#### Motivation and Context

The previous `CmisFlatMemMap` / `CmisMemMap` implementation defined every
field for every CMIS page inside a single ~700-line class. This made it
hard to:

- Locate the fields belonging to a given page.
- Compose alternative memory maps (e.g. flat vs. paged, per-bank variants)
  without duplicating large blocks of field definitions.
- Re-map an entire page (for vendor-specific layouts or future CMIS
  revisions) without editing the monolithic class.

Splitting the map into per-page classes makes each page self-contained
and addressable, lets `CmisFlatMemMap` and `CmisMemMap` declare their
contents by listing the pages they include, and allows shared field
groups that cross page boundaries (like `ADMIN_INFO_FIELD` spanning lower
and upper page 00h, or `ADVERTISING_FIELD` spanning pages 01h and 11h)
to be merged automatically by `register_fields`.

Exposing the page number as a constructor parameter on each page class
also enables entire pages to be remapped easily without duplicating field
declarations.

This change introduces no new fields or changes in behavior.
All the existing unit tests pass.

This change is intended to be behavior-preserving: the resulting
`CmisMemMap` exposes the same set of named fields at the same EEPROM
offsets as before the refactor. Only the internal organization of the
memory map definition changes.


sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/__init__.py
prgeor, line 1: @aditya-nexthop may I suggest to have following dir style?

../public/cmis/pages/

page00.py
page01.py
page02.py
...
...
...
page2f.py
page9f.py

../public/cmis/cmis.py

The page file anywas has comment to indicate what this memmap corresponds to.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/page.py
prgeor, line 1: @aditya-nexthop suggest to rename base.py -> page.py and keep cmis_linear_offset, get_field_from_pages within the CmisPage class

prgeor (commented):
@aditya-nexthop have we considered if any impact to link up time (fully loaded system), dom polling interval, link initialization when single optics insertion. A comparison will be good.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/page9f.py
prgeor, line 1: @aditya-nexthop What about https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/sonic_xcvr/mem_maps/public/cdb.py?

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/__init__.py
aditya-nexthop: updated

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/page.py
aditya-nexthop: yes, makes sense. updated.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/page9f.py
aditya-nexthop: also refactored CDB, C-CMIS, as well as the vendor-specific Amphenol and Credo to use the new page-based scheme.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/__init__.py
aditya-nexthop: I realized that this changes the location of `cmis.py` that holds the`CmisFlatMemMap` and `CmisMemMap` from `sonic_platform_base/sonic_xcvr/mem_maps/public/cmis.py` to `sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/cmis.py`.

We now have a divergence where, `cmis.py` is inside the `cmis` directory while `cdb.py` and `c_cmis.py` is still in `sonic_platform_base/sonic_xcvr/mem_maps/public/`. I recommend all of them are at the same depth.

Where should we put them? Inside `cmis` or outside?

bgallagher-nexthop:
> @aditya-nexthop have we considered if any impact to link up time (fully loaded system), dom polling interval, link initialization when single optics insertion. A comparison will be good.

We shouldn't expect to see any performance impact here for xcvrd from this refactor since the field lookup logic remains the same, so performance should not change after memory map construction. I thought it might be possible that xcvrd startup time could change (affecting link up time) due to the interpreter needing to do more module importing so I did some basic testing with SONiC images before and after this change on a fully loaded system and detected no regression:
- link up time was roughly the same
- the DOM polling loop was still reliably completing within its configured 30s time

prgeor:
@aditya-nexthop can you address copilot comments. Also the PR description is outdated. Thanks for keeping the file nam simple :)

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/page9f.py
```diff
@@ -0,0 +1,21 @@
+"""
+    pg_9f_cdb_message.py
```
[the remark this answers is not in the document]
aditya-nexthop: resolved

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/consts.py
```diff
@@ -0,0 +1,22 @@
+"""
+    cmis_page_consts.py
```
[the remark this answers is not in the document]
aditya-nexthop: resolved

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/cdb.py
```diff
@@ -86,6 +33,12 @@ def __init__(self, codes):
         self.cdb1_write_epl_block_cmd = CdbWriteEplBlock()
         self.cdb1_enter_password_cmd = CdbEnterPassword()
 
+    def add_pages(self, *pages):
+        """Append pages to self.pages and register their fields onto self."""
+        self.pages.extend(pages)
+        for page in pages:
+            page.register_fields(self)
```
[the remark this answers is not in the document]
aditya-nexthop: resolved

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/cmis.py
```diff
@@ -0,0 +1,65 @@
...
+            CmisAdministrativeLowerPage(codes),
+            CmisAdministrativeUpperPage(codes),
+        )
+
+    def add_pages(self, *pages):
+        """Append pages to self.pages and register their fields onto self."""
+        self.pages.extend(pages)
+        for page in pages:
+            page.register_fields(self)
+
```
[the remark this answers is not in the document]
aditya-nexthop: resolved

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/page.py
```diff
@@ -0,0 +1,85 @@
...
+"""
+
+from typing import Dict
+from ....xcvr_mem_map import XcvrMemMap
+from .....fields.xcvr_field import RegField, RegGroupField
+from .consts import CMIS_ARCH_PAGES, CMIS_NUM_NON_BANKED_PAGES
+
+
+class CmisPage(XcvrMemMap):
+    fields: Dict[str, list[RegField]]  # This is a Dictionary of list of fields
```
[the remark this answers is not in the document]
aditya-nexthop: resolved

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/cmis.py
```diff
@@ -0,0 +1,68 @@
...
+        # XcvrMemMap caches _fields on first get_field(); invalidate so newly
+        # registered RegGroupFields are picked up on the next lookup.
+        self._fields = None
+
+    @property
+    def bank(self):
+        """Returns the bank number (read-only)."""
+        return self._bank
+
+class CmisMemMap(CmisFlatMemMap):
```
micas-net, line 52: Can we support distinguish the version of CMIS, I mean cmis5.2, cmis5.3, etc.

bgallagher-nexthop: While I can see the value of defining different memory maps for each CMIS revision, I think this might be a significant refactoring effort that is probably better to tackle in a separate PR and probably have its own HLD.

This PR just aims to refactor the current memory maps to instead be composed from various page-specific classes -- there is intended to be no functional difference in how memory maps behave after this refactor.

micas-net (approved):
It seems good for me.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/cmis.py
aditya-nexthop: Opened https://github.com/sonic-net/sonic-platform-common/issues/677 for tracking this

prgeor (approved):

PR #675, opened 2026-05-25
[BMC] Wait the BMC to be up after BMC update and restart

benle7:
#### Description

Update bmc_fw_update.py to return only after BMC becomes reachable again.

#### Motivation and Context

The BMC update flow includes update and restart commands.
After the restart command succeeds, we need to wait for the BMC restart to complete,
since the completion indication should only be returned once the BMC is reachable again.

config platform firmware install chassis component BMC fw PATH

### DIFF
--- a/sonic_platform_base/bmc_fw_update.py
+++ b/sonic_platform_base/bmc_fw_update.py
@@ -46,12 +46,13 @@ def main():
             # Wait for BMC to become operational
             max_retries = 5
             bmc_is_up = False
-            for _ in range(max_retries):
+            for retry in range(max_retries):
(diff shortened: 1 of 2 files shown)


ganglyu (dismissed):
lgtm

ganglyu (approved):

stephenxs (approved):

liat-grozovik:
This is needed only when the Main CPU (aka host) is managing the BMC installation via the FW util. this is not relevant when sonic to sonic installer is used in cased sonic is running on the BMC. as of that, see no problem with taking this PR.

PR #555, opened 2025-03-30
Support for Custom Amphenol 800G Backplane catridge

prgeor:
#### Description
Added support for reading custom EEPROM of Amphenol 800G Backplane catridge

#### Motivation and Context
For custom application this catridge eeprom information parsing is required

#### How Has This Been Tested?
Tested on custom switch

PR #676, opened 2026-05-26
Add CPO module identifier (0x80) to SFF-8024 XCVR identifier tables

tshalvi:
#### Description
Add the SFF-8024 Identifier value `0x80` (decimal `128`) for **Co-Packaged Optics (CPO)** modules to the `Sff8024` code class in `sonic_platform_base/sonic_xcvr/codes/public/sff8024.py`:

- `XCVR_IDENTIFIERS[128] = 'CPO'`
- `XCVR_IDENTIFIER_ABBRV[128] = 'CPO'`

#### Motivation and Context
CPO is a new module form factor whose SFF-8024 Identifier (`0x80`) was not present in SONiC's `XCVR_IDENTIFIERS` / `XCVR_IDENTIFIER_ABBRV` tables. As a result, CPO modules were decoded as "Unknown", and `type` / `type_abbrv_name` in `TRANSCEIVER_INFO` were left empty, which prevented the rest of the stack from recognising them as a known form factor.

Adding this mapping lets SONiC correctly identify CPO modules end-to-end, and is the prerequisite that allows them to be picked up by the CMIS state machine in `xcvrd`

Manual testing.

### DIFF
--- a/sonic_platform_base/sonic_xcvr/codes/public/sff8024.py
+++ b/sonic_platform_base/sonic_xcvr/codes/public/sff8024.py
@@ -40,6 +40,7 @@ class Sff8024(XcvrCodes):
         29: 'x8 MiniLink',
         30: 'QSFP+ or later with CMIS',
        126: 'Backplane Cartridge',
+       128: 'CPO',  # Vendor-specific allocation for Co-Packaged Optics
     }
 
     XCVR_IDENTIFIER_ABBRV = {
@@ -75,6 +76,7 @@ class Sff8024(XcvrCodes):
         29: 'Link-x8',
         30: 'QSFP+C',
         126: 'BP',  # Backplane Cartridge
+        128: 'CPO',
     }
 
     CONNECTORS = {


Junchao-Mellanox (approved):

aditya-nexthop (commented):
Based on the discussion today, CPO will have its own factories with platform specified CPO IDs. Would that approach make this change unnecessary?

sonic_platform_base/sonic_xcvr/codes/public/sff8024.py
```diff
@@ -75,6 +76,7 @@ class Sff8024(XcvrCodes):
         29: 'Link-x8',
         30: 'QSFP+C',
         126: 'BP',  # Backplane Cartridge
+        128: 'CPO',
```
prgeor, line 79: @tshalvi which SFF spec version are you using?

tshalvi: Baseline is SFF-8024 Rev 4.8a. Table 4-1 defines 0x80 as `vendor-specific`, not as a named module type- we use that value for CPO. This mapping is so identifier decode returns something meaningful for downstream handling (Which is believe is the same idea as the 0x7E<-->`Backplane Cartridge` mapping introduced in #555). 
Returning a generic "Vendor Specific" string here wouldn't be useful for the CMIS logic that keys off the decoded module type. Let me know if you'd prefer we keep the decode as the generic vendor-specific label instead.

prgeor (approved):

PR #682, opened 2026-05-29
Platform Support for CPO

bgallagher-nexthop:
#### Description
This PR makes the following changes:
- Creates mixin classes that contain EEPROM read/write functionality and optoe functionality.
- Refactors SfpBase and SfpOptoeBase to use these mixin classes.
- Adds CPO base classes that leverage the above EEPROM read/write classes: CpoBase, OeBase, ElsfpBase, OptoeOeBase, OptoeElsfpBase.
- Adds a CPO API factory class that handles instantiation of the appropriate API based on the type of CPO hardware.

#### Motivation and Context
This change is required to implement a platform data model for CPO hardware that SONiC daemons can use.

Unit-tests have been added. Later on, when vendors extend the CPO API factory code with changes for their own platforms, the unit-tests will be extended to be more useful.


sonic_platform_base/sonic_xcvr/optoe_eeprom_rw.py
```diff
@@ -0,0 +1,80 @@
...
+CMIS_MODULE_IDS = (0x18, 0x19, 0x1b, 0x1e)
+# Lower-memory byte 2 bit 7: 1 = flat memory (lower + upper page 00h only), 0 = paged.
+CMIS_FLAT_MEM_FILE_OFFSET = 2
+CMIS_FLAT_MEM_BIT_MASK = 0x80
+# Page 01h byte 142 in the optoe linear EEPROM file at bank-0 stride.
+CMIS_BANKS_SUPPORTED_FILE_OFFSET = 270
+# CMIS AdvBnkSupport (page 01h byte 142, bits 0-1): 00b->1, 01b->2, 10b->4 banks.
+CMIS_BANKS_SUPPORTED_TO_MAX_BANK_SIZE = {0: 0, 1: 2, 2: 4}
+
+class OptoeEepromReadWriteMixin(EepromReadWriteMixin, ABC):
```
prgeor, line 8: @bgallagher-nexthop what does `Mixin` mean?

sonic_platform_base/sonic_xcvr/optoe_eeprom_rw.py
```diff
@@ -0,0 +1,80 @@
...
+SFP_OPTOE_PAGE_SIZE = 128
+
+CMIS_MODULE_IDS = (0x18, 0x19, 0x1b, 0x1e)
+# Lower-memory byte 2 bit 7: 1 = flat memory (lower + upper page 00h only), 0 = paged.
+CMIS_FLAT_MEM_FILE_OFFSET = 2
+CMIS_FLAT_MEM_BIT_MASK = 0x80
+# Page 01h byte 142 in the optoe linear EEPROM file at bank-0 stride.
+CMIS_BANKS_SUPPORTED_FILE_OFFSET = 270
+# CMIS AdvBnkSupport (page 01h byte 142, bits 0-1): 00b->1, 01b->2, 10b->4 banks.
+CMIS_BANKS_SUPPORTED_TO_MAX_BANK_SIZE = {0: 0, 1: 2, 2: 4}
```
prgeor: @bgallagher-nexthop can we remove this if NOT used?

bgallagher-nexthop: Removed 👍

sonic_platform_base/sonic_xcvr/optoe_eeprom_rw.py
bgallagher-nexthop: https://en.wikipedia.org/wiki/Mixin

It is just a pattern for including functionality in a class without being a standalone instantiable class itself.

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
```diff
@@ -0,0 +1,73 @@
...
+        self._oe = oe
+
+    def _create_api(self, codes_class, mem_map_class, api_class):
+        mem_map = mem_map_class(codes_class, self._oe.bank)
+        oe_eeprom = XcvrEeprom(self._oe.read_eeprom, self._oe.write_eeprom, mem_map)
+        return api_class(oe_eeprom)
+
+    def create_oe_api(self):
+        # if self._oe.hardware_id.oe_id == OeId.EXAMPLE:
+        #     self._create_api(...)
```
Natgerbi: What is the benefit of having _create_api take the full list of vendors? We already know the oe vendor from the platform.
For ELS, I can understand it, since it may be used across different platforms.

sonic_platform_base/sonic_xcvr/cpo/cpo_base.py
```diff
@@ -0,0 +1,54 @@
...
+        self.hardware_id = hardware_id
+        self._oe_api_factory = OeApiFactory(self)
+
+    def refresh_oe_api(self):
+        self._oe_api = self._oe_api_factory.create_oe_api()
+
+    def get_oe_api(self):
+        if self._oe_api is None:
+            self.refresh_oe_api()
+        return self._oe_api
```
Natgerbi: Why not create a common parent for both the OE and ELSFP bases? Instead of having _oe_api_factory, we could use a generic _api_factory for both. Then create_oe_api and create_elsfp_api could be replaced with a single create_api, making both refresh and get-api logic transparent at the OE and ELSFP base layers.

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
```diff
@@ -0,0 +1,73 @@
+from typing import Optional
+from enum import Enum
+from dataclasses import dataclass
+
+from sonic_platform_base.sonic_xcvr.xcvr_eeprom import XcvrEeprom
+from sonic_platform_base.sonic_xcvr.eeprom_rw import get_vendor_name, get_vendor_part_num
+
+
```
Natgerbi: I think it would be better to keep OeId and OeApiFactory in an OE-related file, and to do the same for ELSFP.

sonic_platform_base/sonic_xcvr/sfp_optoe_base.py
```diff
@@ -6,21 +6,9 @@
 """
 
 from ..sfp_base import SfpBase
+from .optoe_eeprom_rw import OptoeEepromReadWriteMixin
```
prgeor, line 9: @bgallagher-nexthop what is `Mixin`?

bgallagher-nexthop: https://en.wikipedia.org/wiki/Mixin

It is just a pattern for including functionality in a class without being a standalone instantiable class itself.

sonic_platform_base/sonic_xcvr/eeprom_rw.py
```diff
@@ -0,0 +1,69 @@
...
+def get_vendor_name(reader):
+   name_data = reader(VENDOR_NAME_OFFSET, VENDOR_NAME_LENGTH)
+   if name_data is None:
+       return None
+   vendor_name = name_data.decode('utf-8', errors='ignore')
+   return vendor_name.strip()
+
+def get_vendor_part_num(reader):
+   part_num = reader(VENDOR_PART_NUM_OFFSET, VENDOR_PART_NUM_LENGTH)
+   if part_num is None:
```
prgeor: @bgallagher-nexthop can we avoid these standalone functions? Instead lets define an appropritate class? ModuleEepromInfo class?

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
prgeor, line 1: @bgallagher-nexthop would it be better to keep these cpo_<filename>.py under sonic_xcvr/cpo/ dir?

sonic_platform_base/sonic_xcvr/optoe_eeprom_rw.py
```diff
@@ -0,0 +1,71 @@
...
+            with open(sys_path, mode='w') as f:
+                f.write(str(write_timeout))
+        except (OSError, IOError):
+            pass
+
+    def set_page0(self):
+        self.write_eeprom(SFP_OPTOE_PAGE_SELECT_OFFSET, 1, bytearray([0x00]))
+
+    def get_optoe_current_page(self):
+        return self.read_eeprom(SFP_OPTOE_PAGE_SELECT_OFFSET, 1)[0]
```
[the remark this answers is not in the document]
bgallagher-nexthop: I won't change this in order to avoid regressions. This code is pre-existing and just copy pasted from its original home in SfpOptoeBase.

sonic_platform_base/sonic_xcvr/optoe_eeprom_rw.py
```diff
@@ -0,0 +1,71 @@
...
+
+    def read_eeprom(self, offset, num_bytes):
+        try:
+            with open(self.get_eeprom_path(), mode='rb', buffering=0) as f:
+                if offset >= SFP_OPTOE_UPPER_PAGE0_OFFSET  and \
+                    offset < (SFP_OPTOE_UPPER_PAGE0_OFFSET+SFP_OPTOE_PAGE_SIZE) and \
+                        self.get_optoe_current_page() != 0:
+                    # Restoring the page to 0 helps in cases where the optoe driver failed to restore
+                    # the page when say the module was busy with CDB command processing
+                   self.set_page0()
```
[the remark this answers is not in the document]
bgallagher-nexthop: Same as above, this code was just moved, so I'll avoid changing it.

sonic_platform_base/sonic_xcvr/eeprom_rw.py
```diff
@@ -0,0 +1,69 @@
...
+       return None
+   vendor_pn = part_num.decode('utf-8', errors='ignore')
+   return vendor_pn.strip()
```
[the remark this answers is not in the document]
bgallagher-nexthop: Fixed.

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
```diff
@@ -0,0 +1,73 @@
...
+            vendor_part_number=get_vendor_part_num(self._elsfp.read_eeprom),
+        )
+
+    def create_elsfp_api(self):
+        if self._elsfp.hardware_id.elsfp_id is None:
+            # Read vendor name & part number from EEPROM
+            # and determine the correct memory map to use
+            # based on that information.
+            elsfp_info = self._get_elsfp_info()
+
```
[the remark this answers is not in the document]
bgallagher-nexthop: This is intentionally unused. Follow up PRs will use this information.

sonic_platform_base/sonic_xcvr/eeprom_rw.py
bgallagher-nexthop: Done

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
bgallagher-nexthop: Makes sense, moved to a cpo subdir

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
```diff
@@ -0,0 +1,74 @@
...
+    def _create_api(self, codes_class, mem_map_class, api_class):
+        mem_map = mem_map_class(codes_class, self._elsfp.bank)
+        elsfp_eeprom = XcvrEeprom(self._elsfp.read_eeprom, self._elsfp.write_eeprom, mem_map)
+        return api_class(elsfp_eeprom)
+
+    def _get_elsfp_info(self) -> ElsfpInfo:
+        eeprom_info = ModuleEepromInfo(self._elsfp.read_eeprom)
+        return ElsfpInfo(
+            vendor_name=eeprom_info.get_vendor_name(),
+            vendor_part_number=eeprom_info.get_vendor_part_num(),
```
Natgerbi: If we’re talking about joint mode, the EEPROM read operation is the same for both OE and ELS, but the vendor name page and offset may differ. I think we should also pass the offset as an optional parameter.

sonic_platform_base/sonic_xcvr/cpo/cpo_base.py
bgallagher-nexthop: I've refactored the factory and OE/ELSFP classes to have superclasses that share some of the duplicate logic 👍

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
bgallagher-nexthop: Done, there are now oe.py and elsfp.py files that contain most OE/ELSFP specific logic. However, note that I have left OeId and ElsfpId in cpo_base.py to prevent circular dependencies.

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
bgallagher-nexthop: Good point, I've added support for passing an offset to `ModuleEepromInfo`. The offset is retrieved based on the OE hardware ID the factory is using.

sonic_platform_base/sonic_xcvr/cpo/cpo_api_factory.py
bgallagher-nexthop: This is primarily to keep the same consistent approach for instantiating an API between the OE and the ELSFP. It also keeps the logic for API instantiation in common, public code as opposed to hidden in vendor platform code.

The vendor will pass in the CpoHardwareId from their code, then sonic-platform-common code knows how to handle the rest.

cc @prgeor as the original suggester of this ID-based approach, in case you have any additional motivation for this approach you'd like to mention

sonic_platform_base/sonic_xcvr/eeprom_rw.py
```diff
@@ -0,0 +1,84 @@
...
+from typing import Optional
+
+
+class ModuleEepromInfo:
+    VENDOR_NAME_OFFSET = 129
+    VENDOR_PART_NUM_OFFSET = 148
+    VENDOR_NAME_LENGTH = 16
+    VENDOR_PART_NUM_LENGTH = 16
+
+    def __init__(self, reader, offset: Optional[int] = None):
```
Natgerbi: I think you can init it with 0 and then avoid the if statement in _translate

sonic_platform_base/sonic_xcvr/eeprom_rw.py
```diff
@@ -0,0 +1,84 @@
...
+        # EEPROM has been mapped into a different address space.
+        # For instance, an ELSFP EEPROM can be mapped into a
+        # part of another I2C device's address space.
+        self.offset = offset
+
+    def _translate(self, initial_offset: int) -> int:
+        return initial_offset + self.offset if self.offset else initial_offset
+
+    def get_id(self):
+        id_byte_raw = self.reader(self._translate(0), 1)
```
Natgerbi: can we please use some global like VENDOR_NAME?

sonic_platform_base/sonic_xcvr/eeprom_rw.py
```diff
@@ -0,0 +1,84 @@
...
+        if id_byte_raw is None:
+            return None
+        return id_byte_raw[0]
+
+    def get_revision_compliance(self):
+        id_byte_raw = self.reader(self._translate(1), 1)
```
Natgerbi: can we please use some global like VENDOR_NAME?

sonic_platform_base/sonic_xcvr/cpo/elsfp.py
```diff
@@ -0,0 +1,49 @@
...
+
+@dataclass
+class ElsfpInfo:
+    vendor_name: str
+    vendor_part_number: str
+
+
+class ElsfpApiFactory(CpoApiFactory):
+    def _get_elsfp_page_start_offset(self) -> Optional[int]:
+        offsets = {}
```
Natgerbi: this should contain the first page of elsfp mirror in joint mode? it will be map between OE_ID:PAGE_ID? is it well defined that everyone must follow the design where in joint mode all pages most be sequentially ? because theoretically some vendor can decide to map it in different approach if it is not well define.

bgallagher-nexthop: `ModuleEepromInfo` is not really designed or intended to provide full access to upper/paged memory (that is the job of the memory map and API abstractions) -- instead, it is really just there to give a home to methods that allow direct reading/writing of lower memory for bootstrapping purposes, like the logic here to decide what API to instantiate based on vendor information of the hardware. As a result, it should not be assumed pages are re-mapped sequentially. We are only really trying to find lower memory with this logic.

I have renamed this to `_get_elsfp_lower_mem_offset` to make this more clear that this offset is intended to find lower mem only, not really find the entire ELSFP pageset. Let me know if there is anything else you think we should do to make it more clear.

sonic_platform_base/sonic_xcvr/eeprom_rw.py
bgallagher-nexthop: Done

sonic_platform_base/sonic_xcvr/eeprom_rw.py
bgallagher-nexthop: Updated to use constants

sonic_platform_base/sonic_xcvr/eeprom_rw.py
bgallagher-nexthop: Done

sonic_platform_base/sonic_xcvr/cpo/elsfp.py
bgallagher-nexthop: I have also renamed `ModuleEepromInfo` to `ModuleEepromLowerMemoryInfo` to make this more clear and added a docstring.

sonic_platform_base/sonic_xcvr/cpo/oe.py
```diff
@@ -0,0 +1,16 @@
+from sonic_platform_base.sonic_xcvr.cpo.cpo_base import CpoApiFactory, CpoDeviceBase
+
+
+class OeApiFactory(CpoApiFactory):
```
micas-net, line 4: As we know, current module api are all about OE, but here, we define a new oe class, such as oe api, oe factory, it seems that they are deplicated. I think we can treat oe same, in normal module or in CPO module.

abelamit (dismissed):

sonic_platform_base/sonic_xcvr/cpo/oe.py
bgallagher-nexthop: While I agree that the OE heavily resembles a traditional CMIS transceiver from software's perspective, I think we should stick with an OE object as agreed upon in our latest joint design meeting:
- we decided to commit to a separate API (`get_presence` which is declared on a SfpBase object won't make sense for an OE object, for instance).
- we have a different API factory approach than traditional transceivers, based on the CPO hardware ID that is provided by vendor platform code.

prgeor:
@bgallagher-nexthop can you check the merge conflict?

bgallagher-nexthop:
> @bgallagher-nexthop can you check the merge conflict?

I have fixed the merge conflict.

bgallagher-nexthop:
Also fixed the Bailly CPO unit-tests that were failing

abelamit (dismissed):

micas-net (dismissed):
All comments are resolved, lgtm.

sonic_platform_base/sonic_xcvr/cpo/cpo_base.py
```diff
@@ -0,0 +1,69 @@
...
+from sonic_platform_base.sonic_xcvr.eeprom_rw import EepromReadWriteMixin
+
+
+class OeId(Enum):
+    pass
+
+
+class ElsfpId(Enum):
+    pass
+
```
prgeor, line 16: @bgallagher-nexthop I guess these are for the vendor specific IDs. If so can we put some examples?

sonic_platform_base/sonic_xcvr/cpo/cpo_optoe_base.py
prgeor, line 1: @bgallagher-nexthop What is the intention? Seems like both class are not used?

sonic_platform_base/sonic_xcvr/cpo/elsfp.py
```diff
@@ -0,0 +1,48 @@
...
+@dataclass
+class ElsfpInfo:
+    vendor_name: str
+    vendor_part_number: str
+
+
+class ElsfpApiFactory(CpoApiFactory):
+    def _get_elsfp_lower_mem_offset(self) -> int:
+        offsets = {}
+        return offsets.get(self._device.hardware_id.oe_id, 0)
```
prgeor: @bgallagher-nexthop will always return 0 because offset = {} ?

sonic_platform_base/sonic_xcvr/cpo/cpo_base.py
bgallagher-nexthop: I am intending to fill these in with a follow-up PR for Davisson (I will open a PR like [this one](https://github.com/nexthop-ai/sonic-platform-common/pull/9) against sonic-net:master after this merges).

sonic_platform_base/sonic_xcvr/cpo/cpo_optoe_base.py
bgallagher-nexthop: The only difference in these classes is that they inherit from the optoe eeprom read/write classes, so they get all the optoe functionality. Vendors using optoe will inherit from these base classes.

sonic_platform_base/sonic_xcvr/cpo/elsfp.py
bgallagher-nexthop: This will be filled in by later PRs that add support for vendor hardware. For instance, when I post the PR for Broadcom Davisson, this will be filled in with the offset where the first ELSFP page resides.

sonic_platform_base/sonic_xcvr/cpo/elsfp.py
```diff
@@ -0,0 +1,48 @@
...
+
+        raise ValueError(
+            f"Could not determine what ELSFP API to use for CPO HW ID. "
+            f"OE ID: {self._device.hardware_id.oe_id}, ELSFP ID: {self._device.hardware_id.elsfp_id}"
+        )
+
+
+class ElsfpBase(CpoDeviceBase):
+    def _make_api_factory(self) -> CpoApiFactory:
+        return ElsfpApiFactory(self)
```
prgeor, line 46: @bgallagher-nexthop please check that CpoDeviceBase.refresh_api is expecting Elsp API class whereas here you are returning the factory class.

bgallagher-nexthop: it should be okay, OeApiFactory and ElsfpApiFactory  are both subclasses of CpoApiFactory

CpoDeviceBase only expects a CpoApiFactory object (`create_api()` is implemented)

prgeor (approved):

PR #684, opened 2026-06-02
Implement ELSFP CMIS API

bgallagher-nexthop:
#### Description

Add API methods to interact with an ELSFP module, specifically adding methods to cover the information exposed in pages 1Ah and 1Bh of the ELSFP EEPROM.

ELSFP CMIS Spec Document:
https://www.oiforum.com/wp-content/uploads/OIF-ELSFP-CMIS-01.0.pdf

#### Motivation and Context

This is required to expose the information in the ELSFP memory map to application code.

This PR contains new unit-tests that exercises each new API method using an in-memory EEPROM.


sonic_platform_base/sonic_xcvr/api/public/elsfp.py
```diff
@@ -0,0 +1,503 @@
...
+    #                                                           #
+    #   Pages 10h, 11h, 12h, 13h, 2Fh are not present in the    #
+    #   ELSFP memory map and the methods below reflect that.    #
+    #   Note: we leave the VDM methods available for use,       #
+    #   since a vendor could conceivably implement VDM          #
+    #   observables for an ELSFP despite VDM not being          #
+    #   mentioned in the spec.                                  #
+    #############################################################
+
+    # Page 10h — Lane Datapath Config
```
prgeor: @bgallagher-nexthop can we remove these APIs below  which are NOT applicable for ELSP?

sonic_platform_base/sonic_xcvr/api/public/elsfp.py
```diff
@@ -0,0 +1,503 @@
+import typing
+
+from .cmis import CmisApi
+import sonic_platform_base.sonic_xcvr.fields.elsfp_consts as elsfp_consts
+
+class ElsfpApi(CmisApi):
+
+    def _unpack_lane_bits(self, field: str, num_lanes: int = 8) -> list[int]:
```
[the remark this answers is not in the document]
bgallagher-nexthop: This change will not be backported to older releases. Therefore, we don't need to make this backwards compatible with Python 3.7.

tests/sonic_xcvr/test_elsfp.py
```diff
@@ -1,10 +1,16 @@
 import pytest
+import struct
```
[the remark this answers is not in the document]
bgallagher-nexthop: This change will not be backported to releases. Therefore, we don't need to make this backwards compatible with Python 3.7.

sonic_platform_base/sonic_xcvr/api/public/elsfp.py
bgallagher-nexthop: Good point, these methods (and the decision to inherit from `CmisApi`) was made back before we decided on  the latest approach for the CPO platform API structure.

I've removed all of these and `ElsfpApi` no longer inherits from `CmisApi`.

sonic_platform_base/sonic_xcvr/api/public/elsfp.py
```diff
@@ -0,0 +1,346 @@
...
+        Returns:
+            List of num_lanes integers (0 or 1), where index 0 corresponds to
+            lane 1 (bit 0 of the raw value), index 1 to lane 2, etc.
+        """
+        raw = self.xcvr_eeprom.read(field)
+        if raw is None:
+            return None
+        return [(raw >> lane) & 1 for lane in range(num_lanes)]
+
+    def _set_lane_bits(self, field: str, lane_mask: int, value: bool) -> bool:
```
micas-net, line 24: _pack_lane_bits ?

bgallagher-nexthop: It makes sense because it is the inverse of the above `_unpack_lane_bits`, but the "packing" is technically already done since the function accepts the "packed" integer as the `lane_mask` parameter.

I've renamed `_unpack_lane_bits` to `_read_lane_bits` so that it matches `_set_lane_bits` better.

sonic_platform_base/sonic_xcvr/api/public/elsfp.py
```diff
@@ -0,0 +1,346 @@
+import typing
+
+from ..xcvr_api import XcvrApi
+import sonic_platform_base.sonic_xcvr.fields.elsfp_consts as elsfp_consts
+
+class ElsfpApi(XcvrApi):
```
Natgerbi, line 6: Some of the functions return a dictionary, some return a list, and some return only a float.

I think this should already be organized as a table-specific function(like for example get_transceiver_status for OE), and each function should return a dictionary, similar to get_transceiver_status, for example.

Then, once we decide on the STATE_DB structure, whether to use a new table for ELS or the existing table, it will be very easy to connect it to the relevant function and add it to the relevant table.

bgallagher-nexthop: I agree that we should use the same dictionary-based aggregate functions that existing CMIS and SFF APIs provide to `xcvrd` 👍

My plan here was that this PR provides all the primitives to create those aggregate functions that return a dictionary containing a bunch of information (like `get_transceiver_status`, `get_transceiver_status_flags`, `get_transceiver_dom_real_value`, etc) later on when the code PR for CPO DOM support is being authored, so that work is deferred to a time when we are more certain about the STATE_DB schema we will use for CPO.

```
class ElsfpApi:
    def get_transceiver_dom_real_value(self):
        # 1. call various functions added in this PR
        # 2. assemble all of the results into a dictionary
        # 3. return that dictionary to the caller, like xcvrd, for easy publishing to STATE_DB
```

Does that sound okay to you?

micas-net (approved):
LGTM.

sonic_platform_base/sonic_xcvr/api/public/elsfp.py
Natgerbi: yes sounds good ..

abelamit (approved):

abelamit (approved):

prgeor (approved):

PR #686, opened 2026-06-04
Fix hardcoded lane masks when applying loopbacks

abhi-nexthop:
#### Description

Certain transceivers expect exact lane masks for the active appsel when setting Media Side and Host side loopbacks respectively.
Currently, we have them hardcoded as 0xFF which works on certain transceivers but not others. On a 1 media lane transceiver, trying to disable loopback on it with 0xFF might fail depending on the device. We noticed this on several transceivers

Tested on INPHI CORP IN-Q3JZ1-TC-M1 and Acacia DP04QSDD-E20-001

Acacia works with and without the fix
IN-Q3JZ1-TC-M1 fails to set any kind of media loopback without the fix

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -1183,6 +1183,72 @@ class CmisApi(CmisCdbFw, XcvrApi):
         active_apsel_code = self.xcvr_eeprom.read(consts.ACTIVE_APSEL_CODE)
         return defaultdict(lambda: 'N/A') if not active_apsel_code else active_apsel_code
 
+    def _get_active_appsel_descriptor(self):
+        '''
+        Returns the application-advertisement descriptor for the currently
+        active AppSel on host lane 1.
+
+        Raises RuntimeError if the active AppSel cannot be determined --
+        callers must not silently fall back to the default-application
+        descriptor, since that defeats the purpose of resolving the
+        runtime-configured application.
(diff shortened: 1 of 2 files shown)


sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1183,6 +1183,72 @@ def get_active_apsel_hostlane(self):
...
+            raise RuntimeError(
+                'Cannot determine active AppSel host lane count: '
+                'AppSel %d descriptor %r has no host_lane_count'
+                % (active_apsel, descriptor))
+        return descriptor['host_lane_count']
+
+    def get_active_appsel_media_lane_count(self):
+        '''
+        Returns the media lane count for the currently active application.
+
```
prgeor: @abhi-nexthop we don't have existing APIs for these? cc @mihirpat1

prgeor:
@abhi-nexthop are you working on resolving the comments?

sonic_platform_base/sonic_xcvr/api/public/cmis.py
abhi-nexthop: I have repurposed the existing APIs to correctly get active appsel values. For Media we did have it but for host it just returned appsel 0 values.

abhi-nexthop:
@prgeor Sorry Prince I was away and missed the response email notification. Fixed and tested

prgeor (approved):

PR #700, opened 2026-06-25
[CPO] Extend ChassisBase to support CPO ports

bgallagher-nexthop:
#### Description
The ["Port Mapping for CPO" HLD](https://github.com/sonic-net/SONiC/pull/2211) proposes changes to the Chassis object for CPO hardware platforms, where a function called `construct_cpo_devices` is called when an `cpo.json` file is present for the SONiC device. The intent is for vendors to implement `construct_cpo_devices` and create the CPO platform API objects required for each interface on the vendor's switch.

Example usage, from the HLD:
```python3
class VendorCpoChassis(ChassisBase):
  ...
  def construct_cpo_devices(self, cpo_data):
      for interface_name, interface in cpo_data["interfaces"].items():
          associated_devices = interface["associated_devices"]
          assert len(associated_devices) == 2, \
              "We expect 2 devices per interface on this CPO hardware platform"

          # Construct OE/ELSFP objects
          oe = None
          elsfp = None
          for dvc_info in associated_devices:
              device = cpo_data["devices"][dvc_info["device_id"]]
              if device["device_type"] == "optical_engine":
                  oe = VendorOe(hardware_id=self.hw_id, bank=dvc_info["bank"])
              elif device["device_type"] == "external_laser_source":
                  elsfp = VendorElsfp(hardware_id=self.hw_id, bank=dvc_info["bank"])

          # Create a CpoBase-derived object to represent this interface
          assert oe, "No optical engine found for this interface"
          assert elsfp, "No ELSFP found for this interface"
          # VendorCpoSfp would be a subclass of CpoSfpBase here.
          self._sfp_list.append(VendorCpo(hardware_id=self.hw_id, oe=oe, els=elsfp))
```

This PR adds a `construct_cpo_devices` stub function for vendors to implement and calls this function when `cpo.json` is present. It also adds function to store and retrieve CpoBase objects on a Chassis object. Additionally, it changes CPO Platform objects to inherit from DeviceBase so that they adhere to the DeviceBase interface.

Note: this PR requires https://github.com/sonic-net/sonic-buildimage/pull/28077 to be merged first.

#### Motivation and Context
This change is required to instantiate the required platform API objects (OE, ELSFP, etc) for a Chassis that models CPO hardware.

Unit-tests have been added.

### DIFF
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -91,7 +91,7 @@ class ChassisBase(device_base.DeviceBase):
         from sonic_py_common import device_info
         optical_device_data = device_info.get_optical_devices_data()
         if optical_device_data:
-            self.construct_sfp_list_for_topology(optical_device_data)
+            self.construct_optical_devices(optical_device_data)
 
     def get_base_mac(self):
         """
@@ -714,11 +714,11 @@ class ChassisBase(device_base.DeviceBase):
     # SFP methods
     ##############################################
 
-    def construct_sfp_list_for_topology(self, optical_device_data):
+    def construct_optical_devices(self, optical_device_data):
         """
         Construct objects representing the devices driving traffic through
         a front panel port on the chassis based on topology data in
-        optical_devices.json and appends them to self._sfp_list
+        optical_devices.json
 
         Subclasses should implement this method on platforms that provide
         an optical_devices.json file.
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -80,6 +80,15 @@ class ChassisBase(device_base.DeviceBase):
         # SED (Self-Encrypting Drive) password management
         self._sed_mgmt = None
 
+        # On platforms that provide an optical_devices.json file (e.g.
+        # platforms using co-packaged optics, where each port can be driven
+        # by multiple devices), populate the SFP list based on the optical
+        # device topology described in that file
+        from sonic_py_common import device_info
+        optical_device_data = device_info.get_optical_devices_data()
+        if optical_device_data:
+            self.construct_sfp_list_for_topology(optical_device_data)
+
     def get_base_mac(self):
         """
         Retrieves the base MAC address for the chassis
@@ -700,6 +709,21 @@ class ChassisBase(device_base.DeviceBase):
     ##############################################
     # SFP methods
     ##############################################
+
+    def construct_sfp_list_for_topology(self, optical_device_data):
+        """
+        Construct objects representing the devices driving traffic through
+        a front panel port on the chassis based on topology data in
+        optical_devices.json and appends them to self._sfp_list
+
+        Subclasses should implement this method on platforms that provide
+        an optical_devices.json file.
+
+        Args:
+            optical_device_data: Optical device topology data parsed from
+                optical_devices.json
+        """
+        raise NotImplementedError
 
     def get_num_sfps(self):
         """
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -84,14 +84,12 @@ class ChassisBase(device_base.DeviceBase):
         # SED (Self-Encrypting Drive) password management
         self._sed_mgmt = None
 
-        # On platforms that provide an optical_devices.json file (e.g.
-        # platforms using co-packaged optics, where each port can be driven
-        # by multiple devices), populate the SFP list based on the optical
-        # device topology described in that file
+        # On platforms that provide a cpo.json file, populate self._cpo_list
+        # based on the device topology described in that file
         from sonic_py_common import device_info
-        optical_device_data = device_info.get_optical_devices_data()
-        if optical_device_data:
-            self.construct_optical_devices(optical_device_data)
+        cpo_data = device_info.get_cpo_data()
+        if cpo_data:
+            self.construct_cpo_devices(cpo_data)
 
     def get_base_mac(self):
         """
@@ -714,18 +712,17 @@ class ChassisBase(device_base.DeviceBase):
     # SFP methods
     ##############################################
 
-    def construct_optical_devices(self, optical_device_data):
+    def construct_cpo_devices(self, cpo_data):
         """
         Construct objects representing the devices driving traffic through
         a front panel port on the chassis based on topology data in
-        optical_devices.json
+        cpo.json
 
         Subclasses should implement this method on platforms that provide
-        an optical_devices.json file.
+        an cpo.json file.
 
         Args:
-            optical_device_data: Optical device topology data parsed from
-                optical_devices.json
+            cpo_data: device topology data parsed from cpo.json
         """
         raise NotImplementedError
 
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -65,6 +65,10 @@ class ChassisBase(device_base.DeviceBase):
         # available on the chassis
         self._sfp_list = []
 
+        # List of CpoBase-derived objects representing all CPO ports
+        # available on the chassis, indexed by physical port.
+        self._cpo_list = []
+
         # Object derived from WatchdogBase for interacting with hardware watchdog
         self._watchdog = None
 
@@ -768,6 +772,41 @@ class ChassisBase(device_base.DeviceBase):
 
         return sfp
 
+    def get_num_cpos(self):
+        """
+        Retrieves the number of CPO ports available on this chassis
+
+        Returns:
+            An integer, the number of CPO ports available on this chassis
+        """
+        return len(self._cpo_list)
+
+    def get_all_cpos(self):
+        """
+        Retrieves all CPO ports available on this chassis
+
+        Returns:
+            A list of objects derived from CpoBase representing all CPO ports
+            available on this chassis
+        """
+        return [cpo for cpo in self._cpo_list if cpo is not None]
+
+    def get_cpo(self, index):
+        """
+        Retrieves the CPO port corresponding to physical port <index>, if
+        that port is driven by CPO devices.
+
+        Args:
+            index: An integer (>=0), the physical port index (same indexing as
+                   get_sfp()).
+
+        Returns:
+            An object derived from CpoBase representing the specified CPO port,
+            or None if the port is not a CPO port.
+        """
+        if 0 <= index < len(self._cpo_list):
+            return self._cpo_list[index]
+        return None
 
     def get_port_or_cage_type(self, index):
         """
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -776,7 +776,7 @@ class ChassisBase(device_base.DeviceBase):
         Returns:
             An integer, the number of CPO ports available on this chassis
         """
-        return len(self._cpo_list)
+        return sum(1 for cpo in self._cpo_list if cpo is not None)
 
     def get_all_cpos(self):
         """
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -62,12 +62,17 @@ class ChassisBase(device_base.DeviceBase):
         self._current_sensor_list = []
 
         # List of SfpBase-derived objects representing all sfps
-        # available on the chassis
+        # available on the chassis, indexed by physical port.
         self._sfp_list = []
 
         # List of CpoBase-derived objects representing all CPO ports
         # available on the chassis, indexed by physical port.
         self._cpo_list = []
+
+        # Set once the port lists have been checked for consistency. The check
+        # is deferred until the lists are first accessed, since subclasses
+        # populate them after ChassisBase.__init__ has returned.
+        self._port_lists_validated = False
 
         # Object derived from WatchdogBase for interacting with hardware watchdog
         self._watchdog = None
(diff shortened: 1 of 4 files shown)


sonic_platform_base/chassis_base.py
```diff
@@ -65,6 +65,10 @@ def __init__(self):
         # available on the chassis
         self._sfp_list = []
 
+        # List of CpoBase-derived objects representing all CPO ports
+        # available on the chassis, indexed by physical port.
+        self._cpo_list = []
```
micas-net, line 71: Any set function for _cpo_list ? Can we reuse _sfp_list?

sonic_platform_base/chassis_base.py
```diff
@@ -80,6 +84,15 @@ def __init__(self):
...
         self._sed_mgmt = None
 
+        # On platforms that provide an optical_devices.json file (e.g.
+        # platforms using co-packaged optics, where each port can be driven
+        # by multiple devices), populate the SFP list based on the optical
+        # device topology described in that file
+        from sonic_py_common import device_info
+        optical_device_data = device_info.get_optical_devices_data()
+        if optical_device_data:
+            self.construct_sfp_list_for_topology(optical_device_data)
```
prgeor: @bgallagher-nexthop where is `construct_sfp_list_for_topology` defined?

sonic_platform_base/chassis_base.py
prgeor: @micas-net i think, we should separate sfp objects from cpo objects. A CPO switch may have mix of CPO based data port and SFP based data port or auxiliary ports.

sonic_platform_base/chassis_base.py
```diff
@@ -65,6 +65,10 @@ def __init__(self):
         # available on the chassis
         self._sfp_list = []
 
+        # List of CpoBase-derived objects representing all CPO ports
```
KroosMicas, line 69: The _sfp_list array is created in the order of panel port indexes, and _cpo_list is also an ordered array. How can we guarantee the sequence consistency when a device has both CPO ports and traditional SFP transceiver ports?
Example port layout: port_list: [cpo, cpo, cpo, cpo, sfp, sfp]

sonic_platform_base/chassis_base.py
bgallagher-nexthop: The expectation is that vendor subclasses of `ChassisBase` implement this function to create the CPO objects for each interface defined in `optical_devices.json`.

sonic_platform_base/chassis_base.py
bgallagher-nexthop: This PR separates CPO objects (stored in `_cpo_list`) from SFP objects (stored in `_sfp_list`). Accessors are provided via `get_cpo`, `get_all_cpos` and `get_num_cpos` functions.

sonic_platform_base/chassis_base.py
bgallagher-nexthop: I have renamed this function to `construct_optical_devices`

sonic_platform_base/chassis_base.py
bgallagher-nexthop: On a CPO switch with a mix of both types of ports, there should be gaps in each list.

Both lists are indexed by the physical front panel port, so in your example of a front panel port layout of:

`[cpo, cpo, cpo, cpo, sfp, sfp]`

`_cpo_list` would be `[CpoObject, CpoObject, CpoObject, CpoObject, None, None]`
and `_sfp_list` would be `[None, None, None, None, SfpObject, SfpObject]`.

sonic_platform_base/chassis_base.py
```diff
@@ -80,6 +84,13 @@ def __init__(self):
         # SED (Self-Encrypting Drive) password management
         self._sed_mgmt = None
 
+        # On platforms that provide a cpo.json file, populate self._cpo_list
+        # based on the device topology described in that file
+        from sonic_py_common import device_info
+        cpo_data = device_info.get_cpo_data()
+        if cpo_data:
+            self.construct_cpo_devices(cpo_data)
```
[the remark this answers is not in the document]
bgallagher-nexthop: This is fine. It should only require the `self._cpo_list` instance attribute, which is defined above this.

sonic_platform_base/chassis_base.py
```diff
@@ -701,6 +712,20 @@ def get_current_sensor(self, index):
...
     ##############################################
 
+    def construct_cpo_devices(self, cpo_data):
+        """
+        Construct objects representing the devices driving traffic through
+        a front panel port on the chassis based on topology data in
+        cpo.json
+
+        Subclasses should implement this method on platforms that provide
+        an cpo.json file.
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

sonic_platform_base/chassis_base.py
```diff
@@ -744,6 +769,41 @@ def get_sfp(self, index):
...
+            index: An integer (>=0), the physical port index (same indexing as
+                   get_sfp()).
+
+        Returns:
+            An object derived from CpoBase representing the specified CPO port,
+            or None if the port is not a CPO port.
+        """
+        if 0 <= index < len(self._cpo_list):
+            return self._cpo_list[index]
+        return None
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

tests/chassis_base_test.py
```diff
@@ -161,3 +171,39 @@ def test_pdbs_multiple_and_negative_index(self, capsys):
...
+        chassis = ChassisBase()
+        self.mock_get_cpo_data.assert_called_once()
+        assert chassis.get_num_cpos() == 0
+
+    def test_cpo_data_base_class_not_implemented(self):
+        self.mock_get_cpo_data.return_value = {"devices": {}, "interfaces": {}}
+        with pytest.raises(NotImplementedError):
+            ChassisBase()
+
+    def test_construct_sfp_list_for_topology(self):
```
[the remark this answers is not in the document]
bgallagher-nexthop: Done

sonic_platform_base/chassis_base.py
```diff
@@ -744,6 +769,45 @@ def get_sfp(self, index):
...
         return sfp
 
+    def get_num_cpos(self):
+        """
+        Retrieves the number of CPO ports available on this chassis
+
+        Returns:
+            An integer, the number of CPO ports available on this chassis
+        """
+        return len(self._cpo_list)
```
Natgerbi: based on 

Image


https://github.com/nexthop-ai/SONiC/blob/bf16f4d462af7102e3f68b7c16c6ebc6d54ade26/doc/platform/port_mapping_for_cpo.md

I think we will count here also the "None" entries that replace the sfp objects

sonic_platform_base/chassis_base.py
```diff
@@ -744,6 +769,45 @@ def get_sfp(self, index):
...
+            index: An integer (>=0), the physical port index (same indexing as
+                   get_sfp()).
+
+        Returns:
+            An object derived from CpoBase representing the specified CPO port,
+            or None if the port is not a CPO port.
+        """
+        cpo = None
+        try:
+            cpo = self._cpo_list[index]
```
Natgerbi, line 850: do we want to return an error in case the entry in specific index is "None"?

sonic_platform_base/chassis_base.py
Natgerbi, line 97: is the construct_cpo_devices responsible also to add the "None" for the sfps?

bgallagher-nexthop: Correct, the intention is that the implementation of this method should add None for any non-CPO ports

sonic_platform_base/chassis_base.py
bgallagher-nexthop: Good catch, I have updated the implementation to correctly count non-None elements

sonic_platform_base/chassis_base.py
bgallagher-nexthop: I have copied the behaviour of `get_sfp` here, which is to just log an error and return `None` if an out-of-bounds index is used.

So for the two scenarios where an incorrect access occurs, we return `None` in both cases:
- index is out of bounds -> `None` returned
- index is for a non-CPO port -> `None` returned

sonic_platform_base/chassis_base.py
prgeor: @bgallagher-nexthop can we add an assert that len(sfp_list) == len(cpo_list) ? otherwise people don't know what we want to intend? Not here but at some init or approprite place

bgallagher-nexthop: I've updated the PR with the following changes:
- Add validation that `_sfp_list` and `_cpo_list` have the shape we expect when a platform has both SFP and CPO ports. This validation runs at first access to any sfp/cpo method on `ChassisBase` (we cannot do at `__init__` since vendor subclasses will populate sfp_list after the fact usually).
- Change `get_num_sfps` to only count non-`None` objects in the list, now that we can have "gaps" in the list for CPO.

micas-net (dismissed):
LGTM!

sonic_platform_base/chassis_base.py
```diff
@@ -80,6 +89,13 @@ def __init__(self):
         # SED (Self-Encrypting Drive) password management
         self._sed_mgmt = None
 
+        # On platforms that provide a cpo.json file, populate self._cpo_list
+        # based on the device topology described in that file
+        from sonic_py_common import device_info
+        cpo_data = device_info.get_cpo_data()
```
Junchao-Mellanox, line 95: should we do this in a lazy way?

bgallagher-nexthop: The lazy import was just following convention I saw in other parts of platform common code [here](https://github.com/sonic-net/sonic-platform-common/blob/bde6102ce27763d075f87544300450ddb384a3b5/sonic_platform_base/bmc_base.py#L282) and [here](https://github.com/sonic-net/sonic-platform-common/blob/bde6102ce27763d075f87544300450ddb384a3b5/sonic_platform_base/module_base.py#L516)

abelamit (dismissed):

prgeor (dismissed):

sonic_platform_base/chassis_base.py
Junchao-Mellanox: I mean, should we call device_info.get_cpo_data() here eagerly? I think this should be only called when user is trying to access cpo object

bgallagher-nexthop: `get_cpo_data` basically acts as a presence check for `cpo.json`, since it returns `None` if that file does not exist (which should be the case for any non-CPO hardware platforms). As a result, it is called eagerly since it determines whether we are dealing with CPO hardware or not

Junchao-Mellanox: what if user create a chassis object to access the fan list only?

bgallagher-nexthop: performance-wise it should not have much effect, the module import will be the most expensive and realistically that will only take ~25-30 milliseconds from some minimal measurements I performed

Is it the dependency that is the primary concern, that ChassisBase cannot be used without swsscommon installed now? I wonder if there is a use-case there I am not considering

prgeor (approved):

sonic_platform_base/chassis_base.py
Junchao-Mellanox: It will make any CLIs depending on `chassis` object to delay ~25-30 miliseconds on a CPO capable system. For example: `sfputil`, `fwutil` and so on. Even if those CLIs do not access CPO at all.

bgallagher-nexthop: Both [`sfputil`](https://github.com/sonic-net/sonic-utilities/blob/9f5ee85c8cf79657259a954830fb75862c13e492/sfputil/main.py#L24) and [`fwutil`](https://github.com/sonic-net/sonic-utilities/blob/9f5ee85c8cf79657259a954830fb75862c13e492/fwutil/lib.py#L21) already have this dependency, so in practice adding this import won't actually cost anything. Even `show` commands [import it already](https://github.com/sonic-net/sonic-utilities/blob/9f5ee85c8cf79657259a954830fb75862c13e492/show/main.py#L14).

Junchao-Mellanox: i am not saying the import, I am saying the "cpo_data = device_info.get_cpo_data()". We should get it lazy way.

bgallagher-nexthop: For non-CPO platforms, `get_cpo_data()` will search for `cpo.json` in the hwsku and platform directories and [immediately return](https://github.com/sonic-net/sonic-buildimage/blob/c4d8e5888ef18563b9e2779ab3baab307bc877cc/src/sonic-py-common/sonic_py_common/device_info.py#L222-L224) when it is found to not exist. Since `cpo.json` is the canonical way to check whether the platform is CPO/non-CPO, I don't think there is any lower overhead way to achieve this in the current design. If it proves costly, we can always revisit this and devise some lower overhead mechanism if necessary.

yijingyan2:
Hi @bgallagher-nexthop, this change seems to cause the build failure in https://github.com/sonic-net/sonic-buildimage/pull/28794
`ChassisBase.__init__()` now eagerly calls `device_info.get_cpo_data()`. This breaks:
• `sensormond` tests because platform files are unavailable.
•  `bmcctld` tests because its mocked `sonic_py_common` lacks `device_info`.
Could you please help investigate? here is the build log: https://dev.azure.com/mssonic/build/_build/results?buildId=1188779&view=logs&jobId=618cdd86-3a55-5536-cbf2-915e654a5cd8&j=618cdd86-3a55-5536-cbf2-915e654a5cd8&t=21f5e61c-7c7c-5fa7-6b79-94888d7a2c64. Thanks!

bgallagher-nexthop:
> Hi @bgallagher-nexthop, this change seems to cause the build failure in https://github.com/sonic-net/sonic-buildimage/pull/28794

Thank you for the heads up. I am working on a fix for this in https://github.com/sonic-net/sonic-platform-daemons/pull/871 and I am currently validating the fix addresses the build failures.

PR #702, opened 2026-06-29
Add BMCWatchdog base class to be used in SONiC BMC

chander-nexthop:
in SONiC BMC, the watchdog service would be modeled along the classic linux watchdog daemon model, where one daemon would be the owner of /dev/watchdog0. However watchdog util expects the watchdog class to implement the SONiC watchdog API's like arm/disarm etc. So create a BMCWatchdog that implements all these API's as IPC calls to the daemon.

#### Description
BMCWatchdog implements the Linux watchdog API as an IPC client to the hw-watchdog-mgrd daemon (the sole owner of /dev/watchdog0), with a read-only sysfs fallback for is_armed() when the daemon is unreachable. socket and sysfs paths are configurable via the constructor. Includes unit tests.

#### Motivation and Context
By having the BMCWatchdog, all vendors that have their BMC based on aspeed 2720 SoC will instantiate this class within their Chassis object. The watchdog daemon is already part of the aspeed SONiC image.

Unit tests have been added.

Refer to https://github.com/sonic-net/sonic-buildimage/pull/27776 for the discussions.


sonic_platform_base/bmc_watchdog.py
```diff
@@ -0,0 +1,170 @@
...
+import os
+import socket
+import syslog
+
+from sonic_platform_base.watchdog_base import WatchdogBase
+
+
+# Default IPC socket served by the hw-watchdog-mgrd daemon.  This must match
+# SOCKET_PATH in the platform's hw-watchdog-mgrd.py daemon.
+SOCKET_PATH = "/run/hw-watchdog-mgrd.sock"
```
judyjoseph: @chander-nexthop not sure if you already working on this .. but we need a way to mount this path in pmon docker to be able to using platform_api infra to arm/unarm or other operations on watchdog

chander-nexthop: I wasn't. In fact had asked you about this here at https://github.com/sonic-net/sonic-buildimage/issues/28715#issuecomment-5133094224

I shall take this up, but don't think this PR needs to be gated for that.

chander-nexthop: Raised https://github.com/sonic-net/sonic-buildimage/pull/28803

chander-nexthop: Will fix the path in this PR after that PR merges.

judyjoseph (dismissed):

judyjoseph:
@chander-nexthop could you take care of this issue as well : https://github.com/sonic-net/sonic-buildimage/issues/28185. We need to add/update sonic-mgmt tests for bmc watchdog

chander-nexthop:
> @chander-nexthop could you take care of this issue as well : [sonic-net/sonic-buildimage#28185](https://github.com/sonic-net/sonic-buildimage/issues/28185). We need to add/update sonic-mgmt tests for bmc watchdog

Raised https://github.com/sonic-net/sonic-mgmt/pull/27106 for this. Please take a look

judyjoseph (approved):
LGTM

PR #708, opened 2026-07-09
CPO SW Control Vmodule Statistics platform changes

Natgerbi:


sonic_platform_base/sonic_xcvr/api/public/elsfp_cmis.py
```diff
@@ -0,0 +1,236 @@
...
+from ..cdb_capable_mixin import CdbCapableMixin
+from ..xcvr_api import XcvrApi
+
+logger = logging.getLogger(__name__)
+logger.addHandler(logging.NullHandler())
+
+NUM_ELSFP_LANES = 8
+
+
+class ElsfpCmisApi(XcvrApi, CdbCapableMixin):
```
bgallagher-nexthop, line 21: There already exists an `ElsfpApi`. It does not contain some of the aggregate methods like `get_transceiver_dom_real_value` or have CDB support, but is designed to be easily extended to contain that type of functionality.

Could you extend that class so we have one `ElsfpApi` defined in the platform layer?

sonic_platform_base/sonic_xcvr/api/public/cpo.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Each ``get_transceiver_*`` method merges OE and ELS results into a freshly
+    owned dict (a shallow copy of the OE result) so the backend dicts are never
+    mutated; ELS keys win on collisions. ``None`` from either bank is treated as
+    no contribution (so a missing CMIS page on one side does not suppress the
+    other).
+    """
+
+    def __init__(self, optical_engine_xcvr_api, external_laser_source_xcvr_api) -> None:
+        super().__init__(optical_engine_xcvr_api.xcvr_eeprom)
```
bgallagher-nexthop, line 17: I was initially thinking this approach is the best strategy for aggregating/combining OE and ELSFP data, but I think it might be simpler to just do that work in `xcvrd` code as opposed to the platform API layer.

**The xcvr API lifecycle is complicated in this approach.** The aggregate `CpoApi` and ELSFP API both need to be destroyed and re-built upon an ELSFP plug-in and plug-out event in order to ensure they are not stale. That means `CpoBase` needs its own xcvr lifecycle events with a correct implementation that destroys the ELSFP API and `CpoApi` in `remove_xcvr_api` and re-creates both in the `refresh_xcvr_api` functions.

Since we've already committed in `xcvrd` to directly use the new CPO API objects, I think we can just do the aggregation in `xcvrd` code.

CpoBase can provide the OE API to `xcvrd` by default -- this ensures backwards compatibility in CMIS state machine logic and other places:
```python3
class CpoBase(device_base.DeviceBase):
    def __init__(self, hardware_id: CpoHardwareInfo, oe: "OeBase", elsfp: "ElsfpBase"):
        self.hardware_id = hardware_id
        self.oe = oe
        self.elsfp = elsfp

    def refresh_xcvr_api(self):
        self.oe.refresh_api()
        self.elsfp.refresh_api()

    def get_xcvr_api(self):
        # We always default to the OE API for CPO. If the ELSFP API is required,
        # then that can be accessed via self.elsfp.get_api() directly.
        return self.oe.get_api()

    def remove_xcvr_api(self):
        self.oe.remove_api()
        self.elsfp.remove_api()
```

Then, in the CPO-specific areas of `xcvrd`, we can just directly interact with the OE and ELSFP APIs (and perform aggregation optionally if needed):
```python3
    def post_cpo_port_dom_sensor_info_to_db(self, logical_port_name, db_cache=None):
        asic_index = self.port_mapping.get_asic_id_for_logical_port(logical_port_name)
        if asic_index is None:
            self.logger.log_error(f"CPO: Post port dom sensor info to db failed for {logical_port_name} "
                                  "as no asic index found")
            return

        oe_result = self.post_diagnostic_values_to_db(logical_port_name,
                                                 self.xcvr_table_helper.get_dom_tbl(asic_index),
                                                 self.cpo_dom_utils.get_oe_transceiver_dom_sensor_real_value,
                                                 db_cache=db_cache,
                                                 beautify_func=self._beautify_dom_info_dict)
        elsfp_result = self.post_diagnostic_values_to_db(logical_port_name,
                                                 self.xcvr_table_helper.get_dom_tbl(asic_index),
                                                 self.cpo_dom_utils.get_elsfp_transceiver_dom_sensor_real_value,
                                                 db_cache=db_cache,
                                                 beautify_func=self._beautify_dom_info_dict)
        return oe_result and elsfp_result
```

This also eliminates the problem of merging the results and having to care about key collisions and such.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -75,7 +83,8 @@ class VdmSubtypeIndex(Enum):
     "Errored Frames Minimum Host Input" : "errored_frames_min_host_input",
     "Errored Frames Maximum Host Input" : "errored_frames_max_host_input",
     "Errored Frames Average Host Input" : "errored_frames_avg_host_input",
-    "Errored Frames Current Value Host Input" : "errored_frames_curr_host_input"
+    "Errored Frames Current Value Host Input" : "errored_frames_curr_host_input",
+    "ELS Input Power [dBm]" : "els_input_power"
```
bgallagher-nexthop, line 87: This would probably be better housed in some elsfp specific file, since it is not necessarily related to the regular CmisApi

sonic_platform_base/sonic_xcvr/codes/public/cmis.py
```diff
@@ -181,4 +181,54 @@ class CmisCodes(Sff8024):
         2: 4
     }
 
+    CONTROL_MODE = {
```
bgallagher-nexthop, line 184: There exists an ELSFP-specific codes class that already contains these: https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/sonic_xcvr/codes/public/elsfp.py

bgallagher-nexthop (commented):
There seems to be some unrelated changes possibly in this PR (e.g cmis page changes, c-cmis changes), i assume these might be old changes

PR #709, opened 2026-07-13
[module_base]: Add 'recovery' transition type to _TRANSITION_TIMEOUT_DEFAULTS

vvolam:
#### Description

Add a `"recovery": 600` (10 min) entry to `ModuleBase._TRANSITION_TIMEOUT_DEFAULTS`.

`set_module_state_transition()` validates the requested `transition_type` against the keys of `_TRANSITION_TIMEOUT_DEFAULTS`. chassisd's DPU power-cycle auto-recovery acquires the module state-transition lock via `set_module_state_transition(name, 'recovery')`; because `'recovery'` was missing from the defaults, every recovery acquisition was rejected ("Cannot acquire state transition lock") and the power-cycle was silently skipped — blocking DPU auto-recovery.

#### Motivation and Context

The `'recovery'` transition type is used by chassisd (sonic-platform-daemons) to power-cycle an unresponsive DPU. Without a matching entry in `_TRANSITION_TIMEOUT_DEFAULTS`, the transition is never accepted, so a DPU that loses its control plane is never power-cycled. This unblocks the DPU auto-recovery path in sonic-net/sonic-platform-daemons#829.

`recovery` has no `platform.json` override (only `startup`/`shutdown`/`reboot`/`halt_services` are overridable in `_load_transition_timeouts()`), so the 600 s default applies uniformly. The timeout only bounds the power-cycle critical section (pre_shutdown → admin down → admin up → post_startup), which is cleared in `finally`; 600 s is ample headroom.

- Unit tests: updated the timeout-loading tests in `tests/module_base_test.py` to expect the new key. `python3 -m pytest tests/module_base_test.py` → **115 passed**.
- On-device with both master and 202605 image: with the fix, a control-plane-down DPU was power-cycled (reboot-cause "Switch rebooted DPU") and recovered to `ready_status=true`.

Companion to sonic-net/sonic-platform-daemons#829 (chassisd DPU auto-recovery robustness), which issues the `'recovery'` transition this default enables.

### DIFF
--- a/sonic_platform_base/module_base.py
+++ b/sonic_platform_base/module_base.py
@@ -488,7 +488,8 @@ class ModuleBase(device_base.DeviceBase):
         "startup": 300,      # 5 mins
         "shutdown": 180,     # 3 mins
         "reboot": 240,       # 4 mins
-        "halt_services": 60  # 1 min
+        "halt_services": 60, # 1 min
+        "recovery": 600      # 10 mins
     }
 
     _TRANSITION_TIMEOUTS_CACHE = None
--- a/tests/module_base_test.py
+++ b/tests/module_base_test.py
@@ -839,6 +839,21 @@ class TestModuleBase:
             call(self._key("DPU0"), "transition_start_time", "1000"),
         ])
 
+    def test_set_module_state_transition_recovery(self):
+        """The 'recovery' transition type must be accepted (added to defaults)."""
+        db = MagicMock()
+        self.module.state_db = db
+        db.hget.return_value = None
+        with patch.object(self.module, "get_name", return_value="DPU0"), \
+             patch.object(self.module, "_transition_operation_lock"), \
+             patch("time.time", return_value=1000):
+            assert self.module.set_module_state_transition("dpu0", "recovery") is True
+        db.hset.assert_has_calls([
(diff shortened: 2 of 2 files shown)


gpunathilell (dismissed):

rameshraghupathy:
@vvolam LGTM  Consider adding a focused unit test that verifies the new transition type is accepted:

assert module.set_module_state_transition("DPU0", "recovery") is True

vvolam:
> @vvolam LGTM Consider adding a focused unit test that verifies the new transition type is accepted:
> 
> assert module.set_module_state_transition("DPU0", "recovery") is True

@rameshraghupathy Added a new testcase for a happy path too. Please review again

rameshraghupathy:
@vvolam Thanks, the new happy-path test covers my earlier comment. LGTM

rameshraghupathy (approved):

vaibhavhd:
Please add results or the description of what kind of tests are run from 202605

vvolam:
> Please add results or the description of what kind of tests are run from 202605

@vaibhavhd Tests ran on both master and 202605 image on a smartswitch. Updated the PR description with the same.

PR #710, opened 2026-07-14
[sonic_xcvr] Cache static CMIS support/identity fields to optimize EEPROM reads

aditya-nexthop:
#### Description
Cache static CMIS support/identity/capability fields so they are read from the module EEPROM once and reused, instead of being re-read on every DOM/status poll cycle.

Changes:

- **Cache static getters** by applying the `@read_only_cached_api_return`
  decorator (caches until the getter returns a non-`None`, non-empty value) to
  the CMIS advertisement getters that expose static, read-only fields

- **Add a cached `_get_tx_bias_scale_raw()` helper.** The TX bias scaling
  exponent is a static per-module advertisement, but it was read from EEPROM on
  every `get_tx_bias()` and `get_transceiver_threshold_info()` call. Both paths
  now go through the cached helper.

- **Return `None` on failed reads instead of freezing a default.** Because the
  cache only stores non-`None` results, getters now return the raw `None` on a
  transient read failure so the value is re-read on the next call rather than
  caching a wrong default.

- **Update callers for the new `get_supported_power_config` contract**
  (`c_cmis.py`): `set_tx_power()` and `get_transceiver_info()` now use
  `self.get_supported_power_config() or (None, None)` to safely unpack when the read fails.

#### Motivation and Context
These fields (support flags, module identity/interface, and capability
advertisements) are static for the lifetime of a plugged-in module, yet they
were being re-read from the EEPROM on every DOM/status polling cycle. On a
platform with many transceivers this adds redundant I2C/EEPROM traffic and
latency to the hot polling loop.

Caching them behind `@read_only_cached_api_return` reads each field once and serves subsequent calls from memory, cutting EEPROM reads in the per-cycle DOM path. Returning `None` on a failed read (instead of a hard-coded `False` / `(None, None)`) ensures a transient read failure is not cached permanently — the cache re-reads on the next call once the module responds.

- Updated existing unit tests in `tests/sonic_xcvr/test_cmis.py` to reflect the
  new `None`-on-failed-read behavior:
  - `test_get_diag_page_support`: a `None` read now expects `None` (previously
    `False`).
  - `test_is_cdb_supported`: a `None` read now expects `None` (previously
    `False`).
- Covered by the caching unit tests in `tests/sonic_xcvr/test_cmis_cache.py`,
  which validate that decorated getters are cached and can be cleared.
- Full `sonic_xcvr` pytest suite run locally.


sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -2552,85 +2581,94 @@ def get_tx_adaptive_eq_fail_flag(self):
...
 
+    @read_only_cached_api_return
     def get_rx_ouput_amp_ctrl_supported(self):
         '''
         This function returns the supported RX output amp control field
         '''
         rx_amp_support = self.xcvr_eeprom.read(consts.RX_OUTPUT_AMP_CTRL_SUPPORT_FIELD)
-        if not rx_amp_support or rx_amp_support is None:
-            return False
+        if rx_amp_support is None:
```
prgeor: @aditya-nexthop is the if check really required?

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -2552,85 +2581,94 @@ def get_tx_adaptive_eq_fail_flag(self):
...
 
+    @read_only_cached_api_return
     def get_tx_input_recall_buf2_supported(self):
         '''
         This function returns the supported TX input recall buf2 field
         '''
         tx_recall_buf2_support = self.xcvr_eeprom.read(consts.TX_INPUT_EQ_RECALL_BUF2_SUPPORT_FIELD)
-        if not tx_recall_buf2_support or tx_recall_buf2_support is None:
-            return False
+        if tx_recall_buf2_support is None:
```
prgeor: @aditya-nexthop is the if check really required? return the same value?

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -2552,85 +2581,94 @@ def get_tx_adaptive_eq_fail_flag(self):
...
 
+    @read_only_cached_api_return
     def get_rx_output_eq_post_ctrl_supported(self):
         '''
         This function returns the supported RX output eq post control field
         '''
         rx_post_support = self.xcvr_eeprom.read(consts.RX_OUTPUT_EQ_POST_CTRL_SUPPORT_FIELD)
-        if not rx_post_support or rx_post_support is None:
-            return False
+        if rx_post_support is None:
```
prgeor: @aditya-nexthop same as above other comments

sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
```diff
@@ -182,7 +182,7 @@ def set_tx_power(self, tx_power):
         This function sets the TX output power. Unit in dBm
         Return True if the provision succeeds, False if it fails
         '''
-        min_prog_tx_output_power, max_prog_tx_output_power = self.get_supported_power_config()
+        min_prog_tx_output_power, max_prog_tx_output_power = self.get_supported_power_config() or (None, None)
```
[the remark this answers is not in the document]
aditya-nexthop: removed the read altogether!

sonic_platform_base/sonic_xcvr/api/public/cmis.py
aditya-nexthop: no, this was superfluous, removed it.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
aditya-nexthop: no, this was superfluous, removed it.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
aditya-nexthop: no, this was superfluous, removed it.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1465,13 +1490,16 @@ def set_lpmode(self, lpmode, wait_state_change = True):
                     return True
         return False
 
+    @read_only_cached_api_return
     def get_diag_page_support(self):
         '''
         This function returns whether the module supports diagnostic pages
         '''
```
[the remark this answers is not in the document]
aditya-nexthop: fixed

prgeor (approved):

PR #713, opened 2026-07-16
[xcvr]: Fix appl_advt dict-lookup for non-sequential AppSel indices

gs1571:
#### Description
`get_host_lane_count()`, `get_media_lane_count()` and `get_media_lane_assignment_option()`
in `cmis.py` guard their dict lookup into the application advertisement with
`len(appl_advt) >= appl`. That holds only when the advertised Applications form a
contiguous `1..N` range, and returns `0` for a valid Application on a **sparse**
advertisement, where `appl in appl_advt` is true but `len(appl_advt) < appl`.

Changed the guard to `appl in appl_advt`, matching the pattern already used by
`get_host_lane_assignment_option()` in the same file.

#### Motivation and Context
`get_application_advertisement()` is best-effort: it skips any Application whose
host/media interface code cannot be decoded (an unknown SFF-8024 code decodes to
`Unknown` and is dropped). The advertisement dict is therefore allowed to be
sparse / non-contiguous, so consumers must look Applications up by key rather
than assume the keys are `1..N`. Under the old guard, when a module advertises an
Application at an index higher than the number of decoded entries (e.g. `appl=9`
with `len(appl_advt) == 4`), these getters return `0` for a valid Application.

Added regression cases with a sparse `appl_advt` (single high-index entry) for all
three affected functions in `tests/sonic_xcvr/test_cmis.py`, plus a case confirming
a genuinely-absent index still returns `0`.

### DIFF
--- a/tests/sonic_xcvr/test_cmis.py
+++ b/tests/sonic_xcvr/test_cmis.py
@@ -1098,9 +1098,9 @@ class TestCmis(object):
         assert result == expected
 
     @pytest.mark.parametrize("appl, appl_advt, expected", [
-        # Regression: appl present but with an index higher than
-        # len(appl_advt). A previous `len(appl_advt) >= appl` check
-        # incorrectly returned 0 here even though appl_advt[9] exists.
+        # Sparse advertisement: only a high-index App is present (lower ones
+        # were skipped as undecodable SFF-8024 codes), so the App must be
+        # looked up by key - appl_advt[9] exists and returns its count.
         (9, {9: {'media_lane_count': 4}}, 4),
         # appl genuinely absent from the advertisement still returns 0.
         (5, {1: {'media_lane_count': 1}}, 0),
@@ -1178,9 +1178,9 @@ class TestCmis(object):
         assert result == expected
 
     @pytest.mark.parametrize("appl, appl_advt, expected", [
-        # Regression: appl present but with an index higher than
-        # len(appl_advt). A previous `len(appl_advt) >= appl` check
-        # incorrectly returned 0 here even though appl_advt[9] exists.
+        # Sparse advertisement: only a high-index App is present (lower ones
+        # were skipped as undecodable SFF-8024 codes), so the App must be
+        # looked up by key - appl_advt[9] exists and returns its value.
         (9, {9: {'media_lane_assignment_options': 1}}, 1),
         # appl genuinely absent from the advertisement still returns 0.
         (5, {1: {'media_lane_assignment_options': 1}}, 0),
@@ -1220,10 +1220,9 @@ class TestCmis(object):
             (0, False, {1: {'host_lane_count': 1}}, 0),
             # appl not in advertisement returns 0.
             (5, False, {1: {'host_lane_count': 1}}, 0),
-            # Regression: appl present but with an index higher than
-            # len(appl_advt) (e.g. App index 9 from Page 01h NAD advertised
-            # alongside a single App in the Lower Page). A previous
-            # `len(appl_advt) >= appl` check incorrectly returned 0 here.
+            # Sparse advertisement: only a high-index App is present (lower
+            # ones were skipped as undecodable SFF-8024 codes), so the App
+            # must be looked up by key - appl_advt[9] returns its count.
             (9, False, {9: {'host_lane_count': 8}}, 8),
         ]
     )
--- a/tests/sonic_xcvr/test_cmis.py
+++ b/tests/sonic_xcvr/test_cmis.py
@@ -1097,6 +1097,19 @@ class TestCmis(object):
         result = self.api.get_media_lane_count(appl)
         assert result == expected
 
+    @pytest.mark.parametrize("appl, appl_advt, expected", [
+        # Regression: appl present but with an index higher than
+        # len(appl_advt). A previous `len(appl_advt) >= appl` check
+        # incorrectly returned 0 here even though appl_advt[9] exists.
+        (9, {9: {'media_lane_count': 4}}, 4),
+        # appl genuinely absent from the advertisement still returns 0.
+        (5, {1: {'media_lane_count': 1}}, 0),
+    ])
+    def test_get_media_lane_count_sparse_appl_advt(self, appl, appl_advt, expected):
+        self.api.is_flat_memory = MagicMock(return_value=False)
+        with patch.object(self.api, 'get_application_advertisement', return_value=appl_advt):
+            assert self.api.get_media_lane_count(appl) == expected
+
     @pytest.mark.parametrize("mock_response, expected", [
         ('C-band tunable laser', 'C-band tunable laser')
     ])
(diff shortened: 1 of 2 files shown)


bgallagher-nexthop (approved):

bgallagher-nexthop:
Thinking further: a module that advertises only a single application, and it is not the first application descriptor (app sel code 1) is not compliant with the CMIS spec:

*"A module advertises at least one supported Application as its power-up default Application in the first Application Descriptor (identified by AppSel code 1), and it advertises any additionally supported types of Applications consecutively..."*

Looking at the code for `get_application_advertisements`, it seems it is possible that we would skip applications we have no mapping for because [decoding can pass back "Unknown" for missing entries](https://github.com/sonic-net/sonic-platform-common/blob/c95220d57521f29c9cd4d3c646021f08658e95f9/sonic_platform_base/sonic_xcvr/fields/xcvr_field.py#L247) and the `get_application_advertisements` code seems like it could [skip those](https://github.com/sonic-net/sonic-platform-common/blob/c95220d57521f29c9cd4d3c646021f08658e95f9/sonic_platform_base/sonic_xcvr/api/public/cmis.py#L2387-L2388) cases and pass back a dictionary with gaps in the app sel codes.

I don't think it seems likely that the current fix addresses the root cause as result.

It seems like either one of the two is the root cause:
- we are missing entries in our [sff-8024 codes](https://github.com/sonic-net/sonic-platform-common/blob/master/sonic_platform_base/sonic_xcvr/codes/public/sff8024.py) for host/media interfaces
- the module firmware is potentially buggy

bgallagher-nexthop (changes requested):
(Marking request changes to dismiss my approval based on above comment)

aditya-nexthop (commented):
> Found while bringing up an 800G ZR+ coherent module (CMIS 5.3) whose only
> advertised Lower-Memory Application is at a non-1 index

Please can you explain how a module does this? If the only advertised Lower-Memory Application is at a non-1 index, what's in the application advertisements prior (1 to n) to this application?

It would help if you can post either an EEPROM dump of the advertising region or a list of applications this module advertises.

gs1571:
@bgallagher-nexthop @aditya-nexthop

Went back and tried to reproduce this on the actual module. What I could reproduce is a different, and I think more likely, cause — the one @bgallagher-nexthop suggested: our SFF-8024 table not covering all of the module's MediaIF codes.

Setup: an 800G ZR+ coherent module (CMIS 5.3) that, with full SFF-8024 coverage, advertises a clean, contiguous `appl_advt` for apps 1-12, no gaps.

I removed 2 of the 3 SFF-8024 entries this module needs (kept the third), then restarted xcvrd. Result:
```
appl_advt = {
    9:  {'module_media_interface_id': '<code>', 'host_lane_count': 8, ...},
    10: {...},
    11: {...},
    12: {...},
}
```
Apps 1-8 disappeared — `get_application_advertisements()` decoded their MediaIF codes to `Unknown` and skipped them, exactly as you described. Apps 9-12 stayed because their code was still in the table.

xcvrd acted on this dict directly:
```
Setting appl=9
Setting host_lanemask=0xff
Setting media_lanemask=0x1
```

`len(appl_advt) == 4` here. Under the old guard, `4 >= 9` is `False` — same for 10, 11, 12. So `get_media_lane_count()` / `get_host_lane_assignment_option()` would return 0 for all four Applications xcvrd was actually using.

So: I can't prove the module itself is non-compliant, and I'm dropping that claim. But an incomplete SFF-8024 table — which is a moving target, given how often new codes get added (#601, #687, #701 alone span about 9 months) — produces the exact same non-contiguous `appl_advt` shape, and the current guard breaks on it the same way. I'll update the PR description to reflect this instead of the original claim. Let me know if you still want changes to the fix itself, or if the corrected motivation resolves the concern.

tests/sonic_xcvr/test_cmis.py
```diff
@@ -1194,6 +1220,11 @@ def test_get_active_apsel_hostlane(self, mock_response, expected):
             (0, False, {1: {'host_lane_count': 1}}, 0),
             # appl not in advertisement returns 0.
             (5, False, {1: {'host_lane_count': 1}}, 0),
+            # Regression: appl present but with an index higher than
+            # len(appl_advt) (e.g. App index 9 from Page 01h NAD advertised
+            # alongside a single App in the Lower Page). A previous
+            # `len(appl_advt) >= appl` check incorrectly returned 0 here.
```
aditya-nexthop: please don't leave comments about code that no longer exists. It only confuses readers.

tests/sonic_xcvr/test_cmis.py
```diff
@@ -1164,6 +1177,19 @@ def test_get_media_lane_assignment_option(self, appl, expected):
         result = self.api.get_media_lane_assignment_option(appl)
         assert result == expected
 
+    @pytest.mark.parametrize("appl, appl_advt, expected", [
+        # Regression: appl present but with an index higher than
+        # len(appl_advt). A previous `len(appl_advt) >= appl` check
+        # incorrectly returned 0 here even though appl_advt[9] exists.
```
aditya-nexthop: please don't leave comments about code that no longer exists. It only confuses readers.

tests/sonic_xcvr/test_cmis.py
```diff
@@ -1097,6 +1097,19 @@ def test_get_media_lane_count(self, appl, expected):
         result = self.api.get_media_lane_count(appl)
         assert result == expected
 
+    @pytest.mark.parametrize("appl, appl_advt, expected", [
+        # Regression: appl present but with an index higher than
+        # len(appl_advt). A previous `len(appl_advt) >= appl` check
+        # incorrectly returned 0 here even though appl_advt[9] exists.
```
aditya-nexthop: same

aditya-nexthop (approved):
LGTM

bgallagher-nexthop (approved):
The fix makes sense to me when framed as striving for best-effort application advertisement parsing in the presence of an incomplete/outdated SFF-8024 identifier set (software allows a sparse app advertisement dict since it skips applications it could not successfully decode). Please also update the PR description to reflect this reasoning for the fix.

gs1571:
Thanks for your comments. Addressed both asks from this round: reworded the three sparse-`appl_advt` test comments to drop references to the removed `len(appl_advt) >= appl` guard, and updated the PR description to frame the fix as best-effort parsing of a sparse application advertisement. The fix itself (`appl in appl_advt` in the three getters) is unchanged; all `test_cmis.py` tests pass. 

No changes to the fix itself (`appl in appl_advt` in the three getters). Let me know if anything else is needed.

gs1571:
Hi @prgeor, reaching out to you specifically — saw you recently approved/merged related sonic_xcvr PRs (#701, #717, #722). This one is ready: CI green, no conflicts, both community reviewers approved, no open questions. Could you take a look when you have time?

gs1571:
@aditya-nexthop / @bgallagher-nexthop — would one of you mind re-approving on the
current head so the status reflects your approval?

gs1571:
Hi @prgeor — thanks for merging #720. Could you take a look at #713 as well?

It's a small one and ready to go:
- Fix is a `appl in appl_advt` guard in the three getters — best-effort parsing of a sparse application advertisement (module advertises non-contiguous AppSel indices, e.g. when the SFF-8024 table doesn't cover all of the module's MediaIF codes).
- Only touches the three getters; no behavior change for contiguous advertisements.
- All `test_cmis.py` tests pass, CI is green, no conflicts.

PTAL when you have a moment.

gs1571:
Hi @prgeor, could you please take another look at this PR when you have a chance? The review comments have been addressed and all checks are green. Thanks!

gordon-nexthop (approved):
Thank you, was seeing this on our side as well!

PR #718, opened 2026-07-16
[xcvr]: Support 150GHz grid in get_laser_config_freq/set_laser_freq

gs1571:
#### Description
`get_freq_grid()` already maps `GridSpacing` code `8` to `150` (GHz), but the
two functions that actually use the frequency grid for laser tuning were never
updated to match:

- `get_laser_config_freq()` fell through to the generic `else` branch and
  computed `channel * 150` instead of `channel * 25` — producing a frequency
  up to several hundred GHz off, and sometimes outside the module's own
  supported range.
- `set_laser_freq()` had no `grid == 150` branch at all and unconditionally
  returned `False` for any request on a 150GHz-only module.

Fixed both: `get_laser_config_freq()` now treats grid 150 the same as grid 75
(channel step is 25GHz for both — a 150GHz channel is simply every 6th 25GHz
step). `set_laser_freq()` gained a `grid == 150` branch (`GridSpacingTx1` code
`0x80`, channel number required to be a multiple of 6).

#### Motivation and Context
Found bringing up an 800G ZR+ coherent module (CMIS 5.3) that only supports
the 150GHz DWDM grid. Without this fix such a module can never be tuned to a
target frequency — `set_laser_freq()` always fails, and even reading back the
configured frequency is wrong.

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
@@ -88,9 +88,15 @@ class CCmisApi(CmisApi):
         '''
         freq_grid = self.get_freq_grid()
         channel = self.xcvr_eeprom.read(consts.LASER_CONFIG_CHANNEL)
-        if freq_grid in (75, 150):
+        # OIF-CMIS 5.3 Table 8-66: 75GHz is 193.1 + n x 0.025 THz, but 150GHz
+        # is 193.1 + (n+3) x 0.025 THz - the two grids are not interchangeable.
+        if freq_grid == 75:
             config_freq = 193100 + channel * 25
+        elif freq_grid == 150:
+            config_freq = 193100 + (channel + 3) * 25
         else:
+            # All other grids (100/50/25/12.5/6.25/3.125GHz) use a plain
+            # 193.1 + n x grid formula with no additive offset.
             config_freq = 193100 + channel * freq_grid
         return config_freq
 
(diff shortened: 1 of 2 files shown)


sonic_platform_base/sonic_xcvr/api/public/c_cmis.py
```diff
@@ -169,6 +169,10 @@ def set_laser_freq(self, freq, grid):
             assert grid_supported_100GHz
             freq_grid = 0x50
             channel_number = int(round((freq - 193100)/100))
+        elif grid == 150:
+            freq_grid = 0x80
+            channel_number = int(round((freq - 193100)/25))
```
aditya-nexthop: Not sure, but both here and here:
```
        if freq_grid in (75, 150):
            config_freq = 193100 + channel * 25
```
we haven't accounted for the `+3` in the spec:

Image

gs1571: My bad, missed the offset entirely — should have checked the spec table more carefully before submitting. Fixed both directions: get_laser_config_freq() adds 3 to the channel before scaling, set_laser_freq() subtracts 3 before the % 6 check and the write.

One thing worth flagging: this shifts which frequencies are valid on the 150GHz grid — 193175/193325/193475/... instead of 193100/193250/193400/.... Any config targeting the old values will now correctly fail the channel_number % 6 == 0 check, since those were never real grid points; the old code just silently accepted them at the wrong frequency.

Pushed the fix with updated/added tests in test_ccmis.py, including an explicit assertion on the channel number actually written.

aditya-nexthop (approved):

gs1571:
Hi @prgeor, reaching out to you specifically — saw you recently approved/merged related sonic_xcvr PRs (#701, #717, #722). This one is ready: CI green, no conflicts, approved by @aditya-nexthop on current HEAD, no open questions. Could you take a look when you have time?

prgeor (approved):

PR #720, opened 2026-07-17
[xcvr]: Detect coherent modules via CoherentPagesSupported bit

gs1571:
#### Description
`is_coherent_module()` decided coherency purely by searching for the substring
'ZR' in the free-text module media interface name. This breaks for any
coherent module whose active application advertises a media interface name
without 'ZR' in it (e.g. FOIC-only names such as
`'FOIC1.4-DO (G.709.3/Y.1331.3)'`).

Added the `CoherentPagesSupported` bit — Page 01h byte 142 bit 4, defined in
**OIF-CMIS-05.3, section 8.4.4 "Supported Pages Advertisement", Table 8-46
"Supported Pages Advertising (Page 01h)"** ("Banked Pages 30h-4Fh (partially)
supported") — to the Page 01h mem map, alongside the existing
`VdmSupported`/`DiagPageSupportAdvtField` bits already defined on that same
byte. Spec: https://www.oiforum.com/wp-content/uploads/OIF-CMIS-05.3.pdf

`is_coherent_module()` now reads this bit first, since it is defined by spec
rather than inferred from a name, and falls back to matching `'ZR'`/`'FOIC'`
in the media interface name when the module/CMIS revision does not expose
the bit (read returns `None`).

#### Motivation and Context
Found bringing up an 800G ZR+ coherent module (CMIS 5.3) whose default active
Application advertises a media interface name containing `'FOIC'` but not
`'ZR'` — `is_coherent_module()` returned `False`, so `CCmisApi` was never
instantiated and the module's laser could never be configured.


sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1051,9 +1051,18 @@ def get_module_media_interface(self):
...
         '''
         Returns True if the module follow C-CMIS spec, False otherwise
+
+        Prefers the CoherentPagesSupported bit (Page 01h byte 142 bit 4,
+        OIF-CMIS 5.x+) when the module advertises it, since it is defined by
+        spec rather than inferred from a free-text media interface name.
+        Falls back to matching 'ZR' or 'FOIC' in the media interface name for
+        modules/CMIS revisions that do not expose this bit.
         '''
+        coherent_pages_supported = self.xcvr_eeprom.read(consts.COHERENT_PAGES_SUPPORTED)
```
aditya-nexthop: Should we also be checking the CMIS revision (byte 1) here, since this field was reserved in CMIS 5.2 and earlier?

tests/sonic_xcvr/test_cmis.py
```diff
@@ -965,13 +965,36 @@ def test_get_module_media_interface(self, mock_response1, mock_response2, expect
...
+        # are coherent too; only reachable when CoherentPagesSupported is
+        # unavailable (mocked read() below defaults to None).
+        ('FOIC1.4-DO (G.709.3/Y.1331.3)', True),
     ])
     def test_is_coherent_module(self, mock_response, expected):
+        self.clear_cache('is_coherent_module')
+        # CoherentPagesSupported unavailable: force the string-matching
+        # fallback path regardless of what earlier tests left behind on the
+        # shared self.api.xcvr_eeprom mock.
+        self.api.xcvr_eeprom.read = MagicMock(return_value=None)
```
aditya-nexthop, line 978: On a live system, I think the read returns `None` only upon an EEPROM read failure. We will still read byte 142, the value in a CMIS 5.2 or earlier coherent module will likely be `0` as it is reserved and we might end up breaking coherent detection altogether.

aditya-nexthop (commented):
Please can you post in the testing section whether your change still correctly detects coherent (ZR or otherwise) modules advertising CMIS 5.2 or earlier with the Coherent pages supported bit set to both 0 and 1 ?

sonic_platform_base/sonic_xcvr/api/public/cmis.py
gs1571: Yes, you're right — pushed a fix that gates the CoherentPagesSupported check on the module's actual reported CMIS revision. It now reads CmisMajorRevision/CmisMinorRevision (byte 1) first and only trusts the bit when (major, minor) >= (5, 3); every module below that unconditionally falls back to the 'ZR'/'FOIC' string match, exactly as it worked before this bit existed.

tests/sonic_xcvr/test_cmis.py
gs1571: Confirmed, and that's exactly the regression this exposed — XcvrEeprom.read() only returns None on an actual EEPROM/I2C failure, not based on CMIS revision, so a real CMIS 5.2-or-earlier module would read the reserved bit as a genuine 0 and short-circuit straight to False, skipping the fallback and breaking detection for currently-supported coherent modules. Fixed by requiring CMIS 5.3+ before trusting the bit (see reply on the other comment). Added test cases: CMIS 5.2 module with the reserved bit read as 0 and as 1 (both now correctly fall through to string matching), plus unreadable CMIS revision fields.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1052,15 +1052,20 @@ def is_coherent_module(self):
...
+        OIF-CMIS starting at revision 5.3; on earlier revisions that bit is
+        Reserved, and real modules are not guaranteed to report it as 0, so
+        it cannot be trusted to mean "not coherent" there. Only honor the
+        bit on modules that report CMIS 5.3 or later; every other module
+        keeps using the 'ZR'/'FOIC' substring match against the media
+        interface name, exactly as before this bit existed.
+        '''
+        cmis_major = self.xcvr_eeprom.read(consts.CMIS_MAJOR_REVISION)
+        cmis_minor = self.xcvr_eeprom.read(consts.CMIS_MINOR_REVISION)
+        if cmis_major is not None and cmis_minor is not None and (cmis_major, cmis_minor) >= (5, 3):
```
aditya-nexthop: we can just use `get_cmis_rev` here?

Also, being pedantic here, but if a current SONiC release runs a CMIS 5.3+ coherent module that does not correctly advertise `COHERENT_PAGES_SUPPORTED` bit, such a module would be detected as coherent before, but will not be detected as coherent after this change.

We could claim though, that in that case, that the module is not CMIS-compliant due to the mis-advertisement. It is a change of behavior though.

gs1571: You're right that
returning `bool(CoherentPagesSupported)` directly on 5.3+ was a regression: a
coherent module that reports the bit as `0` while still naming a `ZR`/`FOIC`
media interface would flip from coherent to non-coherent.

Reworked `is_coherent_module()` so detection is the **union** of the two signals
rather than the bit overriding the name. The name match returns `True` on its own (same as before the bit existed), and the bit now only adds coherent modules whose name contains neither keyword.
It can no longer drop a module the name match already covered, so the behavior
you flagged is gone. Added a test for a 5.3+ module reporting the bit as `0`
with a coherent media interface name.

**`get_cmis_rev`.** I looked at reusing it, but it returns a formatted string
(`"5.3"`, or `"None.None"` when the read fails), so it can't feed the numeric
`(major, minor) >= (5, 3)` comparison without parsing it back — and a
lexicographic version compare would be unsafe (`"5.10" < "5.3"`). I kept the two
explicit int reads for that reason. Happy to switch to a small numeric helper
if you'd prefer that instead.

gs1571:
Hi @aditya-nexthop,

Just checking in on this one — following your last comment, I pushed `2acf42e` which reworks `is_coherent_module()` so detection is additive (name match OR bit), instead of the bit overriding the name on CMIS 5.3+. This should fix the regression you pointed out: a coherent module reporting `CoherentPagesSupported=0` while still naming a `ZR`/`FOIC` media interface will now still be detected correctly, since the name match alone is enough on its own.

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -1050,10 +1050,27 @@ def get_module_media_interface(self):
...
+            bit existed, and
+          * CoherentPagesSupported (Page 01h byte 142 bit 4) is set. That
+            bit is only defined from OIF-CMIS 5.3 on (Reserved before, where
+            a real module may report it as a stray 1), so it is only honored
+            when the module reports CMIS 5.3 or later.
         '''
         mintf = self.get_module_media_interface()
-        return False if 'ZR' not in mintf else True
+        if any(kw in mintf for kw in ('ZR', 'FOIC')):
+            return True
```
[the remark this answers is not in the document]
gs1571: Restored the `None` guard so a module whose media-interface name can't be read still falls through
to the bit check.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/page01.py
```diff
@@ -85,6 +85,7 @@ def __init__(self, codes, page=ADVERTISING_PAGE):
             NumberRegField(consts.PAGE_SUPPORT_ADVT_FIELD, self.getaddr(142),
                 RegBitField(consts.VDM_SUPPORTED, 6),
                 RegBitField(consts.DIAG_PAGE_SUPPORT_ADVT_FIELD, 5),
+                RegBitField(consts.COHERENT_PAGES_SUPPORTED, 4),
             ),
```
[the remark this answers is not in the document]
gs1571: Agreed, this is backward-incompatible: `NumberRegField.decode()` right-shifts by
`min(bitpos)` of its sub-fields, so adding bit 4 moves `Supported Pages Advertisement`
from `>>5` to `>>4` and widens its mask — the same EEPROM byte now decodes to a
different number for any existing consumer. Moved `CoherentPagesSupported` to its own
standalone `RegBitField` at the same offset (142), mirroring the existing single-bit
`TUNING_IN_PROGRESS`/`WAVELENGTH_UNLOCKED` fields, so the neighbouring field is untouched:

```python
NumberRegField(consts.PAGE_SUPPORT_ADVT_FIELD, self.getaddr(142),
    RegBitField(consts.VDM_SUPPORTED, 6),
    RegBitField(consts.DIAG_PAGE_SUPPORT_ADVT_FIELD, 5),
),
RegBitField(consts.COHERENT_PAGES_SUPPORTED, offset=self.getaddr(142), bitpos=4),
```

This also makes `read(COHERENT_PAGES_SUPPORTED)` resolve directly to the bit. Added two
tests: `is_coherent_module()` returning `True` when the media-interface name is `None`
but the coherent bit is set on CMIS ≥ 5.3, and a decode-stability check that
`PAGE_SUPPORT_ADVT_FIELD` is unchanged when bit 4 is set.

aditya-nexthop (approved):

gs1571:
Hi @prgeor — this one is ready to merge. The two Copilot review comments were addressed in adc0dbf — a None-guard
in is_coherent_module(), and moving CoherentPagesSupported into its own RegBitField
so the neighbouring PAGE_SUPPORT_ADVT_FIELD decode stays unchanged. This is the last
PR in the same coherent-optics series as #718, which you merged on Jul 28. Could you
take a look when you have a moment? Thanks!

sonic_platform_base/sonic_xcvr/api/public/cmis.py
prgeor: @gs1571 may I suggest as follows?

```
    @read_only_cached_api_return
    def is_coherent_module(self):
        '''
        Returns True if the module follows the C-CMIS spec, False otherwise.

        Detection is the union of two independent signals, so the bit can
        only add coherent modules and never drops one that used to be
        detected:
          * the media interface name contains 'ZR' or 'FOIC' - works on
            every CMIS revision and matches the behavior from before this
            bit existed, and
          * CoherentPagesSupported (Page 01h byte 142 bit 4) is set. That
            bit is only defined from OIF-CMIS 5.3 on (Reserved before, where
            a real module may report it as a stray 1), so it is only honored
            when the module reports CMIS 5.3 or later.
        '''
        mintf = self.get_module_media_interface()
        if mintf is not None and any(kw in mintf for kw in ('ZR', 'FOIC')):
            return True
        try:
            cmis_rev = self.get_cmis_rev()
            if tuple(int(x) for x in cmis_rev.split('.')) >= (5, 3):
                return bool(self.xcvr_eeprom.read(consts.COHERENT_PAGES_SUPPORTED))
        except (ValueError, AttributeError):
            pass
        return False
```

gs1571: Thanks @prgeor — adopted in 0c78a71. is_coherent_module() now gates the
CoherentPagesSupported bit via get_cmis_rev(), parsing the revision into an int
tuple (so 5.10 >= 5.3 stays correct) with a try/except that falls back to
name-only detection if the revision read fails ("None.None"). Behaviour is
unchanged and the full cmis/ccmis unit test suite passes. PTAL.

sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages/page01.py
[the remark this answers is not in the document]
prgeor: @gs1571 you can force the decoder to decode individual bit fields by doing this

```
     NumberRegField(consts.PAGE_SUPPORT_ADVT_FIELD, self.getaddr(142),
                RegBitField(consts.VDM_SUPPORTED, 6),
                RegBitField(consts.DIAG_PAGE_SUPPORT_ADVT_FIELD, 5),
                RegBitField(consts.COHERENT_PAGES_SUPPORTED, 4),
                bitdecode=True,
            ),
```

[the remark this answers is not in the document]
gs1571: Thanks @prgeor — adopted in adf6fda: bitdecode=True on PAGE_SUPPORT_ADVT_FIELD.
Adding bit 4 no longer changes the field's decode, and CoherentPagesSupported stays
readable via read(). Tests pass. PTAL.

prgeor (approved):

PR #724, opened 2026-07-21
[cmis] Fix enter_password to include standard password entry method

Junchao-Mellanox:
#### Why I did it
CMIS defines two ways to deliver the host password that unlocks protected CDB/EEPROM access:

1. CDB command 0001h (CdbCmdHandler.enter_password()), and
2. writing the 4-byte password MSB-first to the Password Entry Area at page 00h bytes 122-125.

CDB command 0001h is not honored by all CMIS modules. On modules that unlock only via the Password Entry Area, password entry silently fails, which blocks every password-protected CDB/EEPROM operation — firmware download/run/commit and SED-protected access among them. This change makes password entry work across both module families without changing behavior for modules that already unlock via CDB.

#### How I did it
- `cdb.py` — Reworked `enter_password()` to first write the password to the Password Entry Area (page 00h bytes - api/public/cmis.py — New CmisApi.enter_password_via_memory(password), which writes the password to the Password Entry Area and then interprets the result:
  - _supports_password_cmd_result() gates on CMIS revision — PasswordCmdResult (00h:42.3-0) was introduced in CMIS 5.3, so on earlier modules (or when the revision can't be parsed) those bits are reserved and are never interpreted.
  - _read_password_cmd_result() polls PasswordCmdResult past the transient "validation in progress" / unreadable states, bounded by a 1000 ms timeout at 20 ms intervals (per CMIS 8.2.14 the module updates the result within tWRITE).
  - Verdict handling: HOST_ACCEPTED/MODULE_ACCEPTED → True; NOT_ACCEPTED → False; pre-5.3, NOT_SUPPORTED, or undetermined → best-effort True on a successful write.
- api/public/cdb_fw.py — New CmisCdbFw._enter_password() becomes the single entry point for all password entry. It validates the password (int in 0..0xFFFFFFFF), tries CDB 0001h first because that path is synchronous (send_cmd waits for and checks CDB status), and falls back to the Password Entry Area write on failure — or goes straight to it when no CDB handler is available. All existing call sites (get_module_fw_info, FW run/commit/download-start password retries, and cdb_enter_host_password) now go through it.
- cdb/cdb.py — CdbCmdHandler.enter_password() keeps its original CDB-only behavior; only the docstring is updated to state that the fallback is the caller's responsibility.
- fields/consts.py — Added PASSWORD_ENTRY (offset 122, size 4), PASSWORD_CMD_RESULT, the PASSWORD_RESULT_* code constants, PASSWORD_RESULT_MIN_CMIS_REV = (5, 3), and the poll interval/timeout constants, with comments citing CMIS 8.2.14.
- mem_maps/.../page00_lower.py — Registered the PasswordEntry field (NumberRegField, format=">I", writable — the format packs MSB-first) and the 4-bit PasswordCmdResult field at 00h:42.

#### How to verify it
Unit tests in tests/sonic_xcvr/test_cmis.py and tests/sonic_xcvr/test_cdb.py:

- _supports_password_cmd_result gating across CMIS 4.0 / 5.2 / 5.3 / 5.4 / 6.0 and unparseable revisions.
- _read_password_cmd_result polling past IN_PROGRESS and None reads, plus the timeout case.
- enter_password_via_memory: write failure, pre-5.3 best-effort (no register read), all four PasswordCmdResult codes, and the undetermined-result case.
- _enter_password: CDB success short-circuits the fallback; fallback is invoked and its result propagated on CDB failure; no-CDB-handler path; invalid passwords ("not_an_int", -1, 0x100000000, 12.5, None) rejected without touching either path.
- CdbCmdHandler.enter_password retains coverage for CDB failure propagation and invalid-input rejection without sending a command.

### DIFF
--- a/sonic_platform_base/sonic_xcvr/api/public/cdb_fw.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cdb_fw.py
@@ -44,6 +44,35 @@ class CmisCdbFw:
         self._init_cdb_fw_handler = False
         return None
 
+    def _enter_password(self, password=cdb_consts.CDB_DEFAULT_PASSWORD):
+        """
+        Enter the CMIS host password to unlock protected CDB/EEPROM access.
+
+        The password is entered via CDB command 0001h first, since that path is
+        synchronous (send_cmd waits for and checks the CDB status). If it fails
+        -- or CDB is not available at all -- fall back to the non-CDB Password
+        Entry Area write (the register is in the CMIS memory map, not the CDB
+        one).
+
+        Returns True if the password was accepted (or delivered best-effort),
+        False otherwise.
+        """
+        if not isinstance(password, int) or \
+                password < 0 or password > 0xFFFFFFFF:
+            log.log_notice('Invalid password: must be an integer in range 0..0xFFFFFFFF')
+            return False
+
+        if self.cdb_fw_hdlr is None:
+            log.log_notice('CDB not available; entering password via Password Entry Area write')
+            return self.enter_password_via_memory(password) is True
+
+        if self.cdb_fw_hdlr.enter_password(password) is True:
+            return True
+
+        log.log_notice('CDB command 0001h password entry failed; '
+                       'falling back to Password Entry Area write')
+        return self.enter_password_via_memory(password) is True
+
     def get_module_fw_mgmt_feature(self, verbose = False):
         """
         This function obtains CDB features supported by the module from CDB command 0041h,
@@ -105,7 +134,7 @@ class CmisCdbFw:
         if fw_info is False or fw_info is None:
             if self.get_status_code() == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
                 log.log_notice('Get module FW info: Need to enter password')
-                self.cdb_fw_hdlr.enter_password()
+                self._enter_password()
                 fw_info = self.cdb_fw_hdlr.get_firmware_info()
 
         if fw_info is False or fw_info is None:
@@ -238,7 +267,7 @@ class CmisCdbFw:
             if fw_run_status == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
                 string = 'Module FW run: Need to enter password\n'
                 log.log_notice(string)
-                self.cdb_fw_hdlr.enter_password()
+                self._enter_password()
                 result = self.cdb_fw_hdlr.run_fw_image(mode)
                 if result is not True:
                     txt += 'Module FW run: Fail after password retry\n'
@@ -276,7 +305,7 @@ class CmisCdbFw:
             if fw_commit_status == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
                 string = 'Module FW commit: Need to enter password\n'
                 log.log_notice(string)
-                self.cdb_fw_hdlr.enter_password()
+                self._enter_password()
                 result = self.cdb_fw_hdlr.commit_fw_image()
                 if result is not True:
                     txt += 'Module FW commit: Fail after password retry\n'
@@ -339,7 +368,7 @@ class CmisCdbFw:
     def cdb_enter_host_password(self, password):
         if self.cdb_fw_hdlr is None:
             return 0
-        if self.cdb_fw_hdlr.enter_password(password) is True:
+        if self._enter_password(password) is True:
             log.log_notice('CDB host auth status: Success')
             return 1
         status = self.get_status_code()
@@ -373,7 +402,7 @@ class CmisCdbFw:
         # password error - retry with default password
         if fw_start_status == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
             log.log_notice('Start module FW download: Need to enter password\n')
-            self.cdb_fw_hdlr.enter_password()
+            self._enter_password()
             if self.cdb_fw_hdlr.start_fw_download(imagepath) is True:
                 return True, ''
             txt = 'Start module FW download: Fail after password retry\n'
--- a/sonic_platform_base/sonic_xcvr/api/public/cdb_fw.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cdb_fw.py
@@ -44,35 +44,6 @@ class CmisCdbFw:
         self._init_cdb_fw_handler = False
         return None
 
-    def _enter_password(self, password=cdb_consts.CDB_DEFAULT_PASSWORD):
-        """
-        Enter the CMIS host password to unlock protected CDB/EEPROM access.
-
-        The password is entered via CDB command 0001h first, since that path is
-        synchronous (send_cmd waits for and checks the CDB status). If it fails
-        -- or CDB is not available at all -- fall back to the non-CDB Password
-        Entry Area write (the register is in the CMIS memory map, not the CDB
-        one).
-
-        Returns True if the password was accepted (or delivered best-effort),
-        False otherwise.
-        """
-        if not isinstance(password, int) or \
-                password < 0 or password > 0xFFFFFFFF:
-            log.log_notice('Invalid password: must be an integer in range 0..0xFFFFFFFF')
-            return False
-
-        if self.cdb_fw_hdlr is None:
-            log.log_notice('CDB not available; entering password via Password Entry Area write')
-            return self.enter_password_via_memory(password) is True
-
-        if self.cdb_fw_hdlr.enter_password(password) is True:
-            return True
-
-        log.log_notice('CDB command 0001h password entry failed; '
-                       'falling back to Password Entry Area write')
-        return self.enter_password_via_memory(password) is True
-
     def get_module_fw_mgmt_feature(self, verbose = False):
         """
         This function obtains CDB features supported by the module from CDB command 0041h,
@@ -134,8 +105,17 @@ class CmisCdbFw:
         if fw_info is False or fw_info is None:
             if self.get_status_code() == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
                 log.log_notice('Get module FW info: Need to enter password')
-                self._enter_password()
+                self.cdb_fw_hdlr.enter_password()
                 fw_info = self.cdb_fw_hdlr.get_firmware_info()
+
+            # Some modules do not unlock via CDB command 0001h. Fall back to the
+            # non-CDB Password Entry Area write (EEPROM page 00h bytes 122-125).
+            if (fw_info is False or fw_info is None) and \
+                    self.get_status_code() == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
+                log.log_notice('Get module FW info: CDB password entry failed, '
+                               'falling back to EEPROM password entry')
+                if self.enter_password_via_memory(cdb_consts.CDB_DEFAULT_PASSWORD) is True:
+                    fw_info = self.cdb_fw_hdlr.get_firmware_info()
 
         if fw_info is False or fw_info is None:
             # Return 0 distinguishes busy/command failure and interface fail from unsupported CDB
@@ -267,7 +247,7 @@ class CmisCdbFw:
             if fw_run_status == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
                 string = 'Module FW run: Need to enter password\n'
                 log.log_notice(string)
-                self._enter_password()
+                self.cdb_fw_hdlr.enter_password()
                 result = self.cdb_fw_hdlr.run_fw_image(mode)
                 if result is not True:
                     txt += 'Module FW run: Fail after password retry\n'
@@ -305,7 +285,7 @@ class CmisCdbFw:
             if fw_commit_status == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
                 string = 'Module FW commit: Need to enter password\n'
                 log.log_notice(string)
-                self._enter_password()
+                self.cdb_fw_hdlr.enter_password()
                 result = self.cdb_fw_hdlr.commit_fw_image()
                 if result is not True:
                     txt += 'Module FW commit: Fail after password retry\n'
@@ -368,7 +348,7 @@ class CmisCdbFw:
     def cdb_enter_host_password(self, password):
         if self.cdb_fw_hdlr is None:
             return 0
-        if self._enter_password(password) is True:
+        if self.cdb_fw_hdlr.enter_password(password) is True:
             log.log_notice('CDB host auth status: Success')
             return 1
         status = self.get_status_code()
@@ -402,7 +382,7 @@ class CmisCdbFw:
         # password error - retry with default password
         if fw_start_status == cdb_consts.CDB_PASSWORD_ERROR_STATUS:
             log.log_notice('Start module FW download: Need to enter password\n')
-            self._enter_password()
+            self.cdb_fw_hdlr.enter_password()
             if self.cdb_fw_hdlr.start_fw_download(imagepath) is True:
                 return True, ''
             txt = 'Start module FW download: Fail after password retry\n'
--- a/sonic_platform_base/sonic_xcvr/api/public/cmis.py
+++ b/sonic_platform_base/sonic_xcvr/api/public/cmis.py
@@ -323,6 +323,90 @@ class CmisApi(CmisCdbFw, XcvrApi):
         cmis_rev = [str(num) for num in [cmis_major, cmis_minor]]
         return '.'.join(cmis_rev)
 
+    def _supports_password_cmd_result(self):
+        """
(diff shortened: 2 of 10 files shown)


Junchao-Mellanox:
Hi @prgeor , @mihirpat1 , could you please help review this? Thanks!

stephenxs (dismissed):

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -135,12 +136,30 @@ def get_cmd_status_code(self):
...
         """
-        if not isinstance(password, int) or password < 0 or password > 0xFFFFFFFF:
+        if not isinstance(password, int) or \
+                password < 0 or password > 0xFFFFFFFF:
             log.log_notice("Invalid password: must be an integer in range 0..0xFFFFFFFF")
             return False
+
+        # Preferred: write the password to the Password Entry Area (MSB first).
+        pwd_bytes = bytearray(struct.pack(">I", password))
+        if self.write_raw(cdb_consts.CDB_HOST_PASSWORD_ENTRY_OFFSET,
```
bgallagher-nexthop: I think this lower memory password logic is probably better housed in `CmisApi`, since transceiver modules with no CDB support can still provide password protected features via these registers in lower memory. This CDB handler object would never be instantiated for those modules, meaning password protection via this mechanism would not be possible.

Something like:
```
class CmisApi:
    ...
    def enter_password(self, password):
        # 1. attempt lower memory password write
        # 2. attempt CDB password entry via the cdb object
        self.cdb_enter_host_password(password)
```

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -135,12 +136,30 @@ def get_cmd_status_code(self):
...
+                                  cdb_consts.CDB_HOST_PASSWORD_ENTRY_SIZE, pwd_bytes):
+            return True
+
+        # Fallback: enter the password via CDB command 0001h.
```
bgallagher-nexthop: I am not confident fallback should be based on only the result of the write.

The spec (5.3 onwards) says the following in section "8.2.14 Password Entry and Change":

> "The result of the last password entry or of the last password change by a WRITE to one of the following registers is eventually indicated in the PasswordCmdResult register 00h:42.3-0. While password validation is still in progress (indicated in PasswordCmdResult) and until the validation result is available, the host should refrain from using password entry registers. The module may also reject READ access to the PasswordCmdResult register until the result has been determined and updated by the module within a period not exceeding tWRITE."

I feel like we'd need to do something like the below process in order to support a combination of both lower mem and CDB-backed password entry:
1. Attempt write to lower memory
2. if write fails, fallback to CDB method
3. read PasswordCmdResult register to understand if password was accepted. if not, fall back to CDB (?)

sonic_platform_base/sonic_xcvr/cdb/cdb.py
Junchao-Mellanox: I prefer to keep the current implementation for the following reasons:

- There is currently no function in CmisApi that calls enter_password.
- Even for CDB-capable modules, sometimes the lower memory password write method is still required to enter the module password. If enter_password is implemented in CmisApi, functions in CdbCmdHandler would need access to a CmisApi object to invoke it, creating a bidirectional dependency.

bgallagher-nexthop: Good points, that makes sense 👍 

If password support for non CDB capable modules is ever required it can always be added as a new function in `CmisApi` and the shared logic between the CDB and non-CDB function factored out to some shared place, but I agree that can be treated as a separate concern to this PR

sonic_platform_base/sonic_xcvr/cdb/cdb.py
Junchao-Mellanox: Make sense to me, trying to fix this

sonic_platform_base/sonic_xcvr/fields/cdb_consts.py
```diff
@@ -73,6 +73,34 @@
...
+CDB_PASSWORD_CMD_RESULT_CODE = "CdbPasswordCmdResultCode"
+CDB_PASSWORD_RESULT_NOT_SUPPORTED = 0x0    # not supported (legacy before CMIS 5.3)
+CDB_PASSWORD_RESULT_MODULE_ACCEPTED = 0x1  # module password entry/change accepted
+CDB_PASSWORD_RESULT_HOST_ACCEPTED = 0x2    # host password entry/change accepted
+CDB_PASSWORD_RESULT_NOT_ACCEPTED = 0x3     # password entry not accepted
+CDB_PASSWORD_RESULT_IN_PROGRESS = 0x8      # password validation in progress
+# Poll bound for PasswordCmdResult after writing the Password Entry Area. The
+# module updates the result within tWRITE and may reject reads (or report
+# "in progress") until then; give it a small margin.
+CDB_PASSWORD_RESULT_POLL_INTERVAL = 100    # msec
```
bgallagher-nexthop: This is fine as is IMO, but just a quick FYI that `tWRITE` is 10ms so this could possibly be set lower in theory.

Image

bgallagher-nexthop (approved):
LGTM 👍

sonic_platform_base/sonic_xcvr/fields/cdb_consts.py
Junchao-Mellanox: will set it to 20ms, thanks for the comments

keboliu (dismissed):

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -133,16 +134,126 @@ def get_cmd_status_code(self):
...
+
     def enter_password(self, password=cdb_consts.CDB_DEFAULT_PASSWORD):
         """
-        Enter host password via CDB command 0001h.
-        Returns True if password accepted, False/None otherwise.
+        Enter the CMIS host password to unlock protected CDB/EEPROM access.
+
+        Per CMIS 8.2.14 the password is entered by writing the 4-byte value
+        (MSB first) to the Password Entry Area at page 00h bytes 122-125. The
+        write succeeding at the transport level does NOT mean the password was
```
prgeor: @Junchao-Mellanox can we ensure backward compatibility? Like if CDB command fails, fall back to new approach?

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -133,16 +134,126 @@ def get_cmd_status_code(self):
...
             return False
-        payload = {"password": password}
-        return self.send_cmd(cdb_consts.CDB_ENTER_PASSWORD_CMD, payload)
+
+        # Preferred: write the password to the Password Entry Area (MSB first).
+        pwd_bytes = bytearray(struct.pack(">I", password))
+        if not self.write_raw(cdb_consts.CDB_HOST_PASSWORD_ENTRY_OFFSET,
+                              cdb_consts.CDB_HOST_PASSWORD_ENTRY_SIZE, pwd_bytes):
+            log.log_notice("Password Entry Area write failed; falling back to CDB command 0001h")
+            return self._enter_password_via_cdb(password)
```
prgeor: @Junchao-Mellanox please avoid write_raw, instead develop the proper cmis memap eeprom 
https://github.com/sonic-net/sonic-platform-common/tree/master/sonic_platform_base/sonic_xcvr/mem_maps/public/cmis/pages

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -133,16 +134,126 @@ def get_cmd_status_code(self):
         """
         return self.last_cmd_status
 
+    def _get_cmis_rev(self):
```
prgeor: @Junchao-Mellanox please use existing API from 
src/sonic-platform-common/sonic_platform_base/sonic_xcvr/api/public/cmis.py

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -133,16 +134,126 @@ def get_cmd_status_code(self):
...
+        """
+        Whether the PasswordCmdResult register (00h:42.3-0) is defined for this
+        module. It was introduced in CMIS 5.3; on earlier modules those bits are
+        reserved, so their value must not be used to judge password acceptance.
+
+        Returns False if the CMIS revision cannot be determined, so the code
+        does not interpret a reserved register on a legacy module.
+        """
+        rev = self._get_cmis_rev()
+        return rev is not None and rev >= cdb_consts.CDB_PASSWORD_RESULT_MIN_CMIS_REV
```
prgeor: @Junchao-Mellanox please add proper decoding logic. check for revision is good but not guaranteed to be supported

Image

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -133,16 +134,126 @@ def get_cmd_status_code(self):
...
-        return self.send_cmd(cdb_consts.CDB_ENTER_PASSWORD_CMD, payload)
+
+        # Preferred: write the password to the Password Entry Area (MSB first).
+        pwd_bytes = bytearray(struct.pack(">I", password))
+        if not self.write_raw(cdb_consts.CDB_HOST_PASSWORD_ENTRY_OFFSET,
+                              cdb_consts.CDB_HOST_PASSWORD_ENTRY_SIZE, pwd_bytes):
+            log.log_notice("Password Entry Area write failed; falling back to CDB command 0001h")
+            return self._enter_password_via_cdb(password)
+
+        # PasswordCmdResult (00h:42.3-0) only exists on CMIS 5.3+. On earlier
```
prgeor: @Junchao-Mellanox The reason initially we used CDB method for password was because CDB method give MORE accurate feedback if the module accepted or rejected the password. Please see the CMIS spec


Image


Writing to EEPROM directly gives weak feedback in `PasswordCmdResult ` register .i.e it DOESNOT tell you if the host password is accepted/not or module password is accepted/not. Whereas CDB method is MORE reliable


1. CDB commands are checksum protected, so you know your payload i.e PASSWORD is checksum protected. So you need to check for  password status only if CDB command is successfully which tells you exactly WHICH password is successfully accepted

Image

Writing the EEPROM with password if the i2c payload is corrupted will give wrong feedback to the user. I would suggest to keep the CDB approach as the default

Junchao-Mellanox: Fine, will do this:
1. Enter password via CDB
2. Fallback to directly write if step 1 fails

sonic_platform_base/sonic_xcvr/cdb/cdb_fw.py
```diff
@@ -14,8 +14,8 @@
 log.logger.propagate = False
 
 class CdbFwHandler(CdbCmdHandler):
-    def __init__(self, reader, writer, mem_map):
-        super(CdbFwHandler, self).__init__(reader, writer, mem_map)
+    def __init__(self, reader, writer, mem_map, cmis_api):
```
prgeor: @Junchao-Mellanox please avoid polluting the namespace by introducing cmis_api into the CdbFwHandler. The handler is a service provider....the API is the user lets not inter-twine both of them.

Junchao-Mellanox: Hi @prgeor , we cannot achieve following 2 items together:
1. Not putting cmis_api to CdbFwHandler (Your current review comment)
2. Put all memory password related logic to CmisAPI, cmis memory map, cmis consts. (Your previous review comment)

If we choose 1, we will need to put all memory password related logic to Cdb scope, including some duplicate logic such as implementing cdb.get_cmis_rev ; otherwise, we will need to put cmis_api to CdbFwHandler. I don't see other option. Please decide what to do.

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -14,9 +15,24 @@
...
+        # owns (and caches) this handler, so a strong back-reference here would
+        # form CmisApi -> handler -> CmisApi. The CmisApi always outlives the
+        # handler (it created and holds it), so dereferencing is safe in
+        # practice; self.cmis_api returns None only once the owning API is gone.
+        self._cmis_api_ref = weakref.ref(cmis_api)
+
+    @property
+    def cmis_api(self):
+        """The owning CmisApi, or None if it has already been collected."""
+        return self._cmis_api_ref()
```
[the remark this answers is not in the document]
prgeor: @Junchao-Mellanox

sonic_platform_base/sonic_xcvr/api/public/cdb_fw.py
```diff
@@ -44,6 +44,35 @@ def _create_cdb_fw_handler(self):
         self._init_cdb_fw_handler = False
         return None
 
+    def _enter_password(self, password=cdb_consts.CDB_DEFAULT_PASSWORD):
```
prgeor, line 47: @Junchao-Mellanox how does password access via EEPROM relates to the CDB firmware handler. this file is for cdb_fw handling.

sonic_platform_base/sonic_xcvr/cdb/cdb.py
```diff
@@ -135,15 +135,22 @@ def get_cmd_status_code(self):
 
     def enter_password(self, password=cdb_consts.CDB_DEFAULT_PASSWORD):
         """
-        Enter host password via CDB command 0001h.
-        Returns True if password accepted, False/None otherwise.
+        Enter the host password via CDB command 0001h. This path is synchronous:
```
prgeor, line 138: @Junchao-Mellanox keep this file changes limited to CDB protocol

sonic_platform_base/sonic_xcvr/api/public/cmis.py
```diff
@@ -323,6 +323,90 @@ def get_cmis_rev(self):
         cmis_rev = [str(num) for num in [cmis_major, cmis_minor]]
         return '.'.join(cmis_rev)
 
+    def _supports_password_cmd_result(self):
```
prgeor, line 326: @Junchao-Mellanox to maintain backward compatibility, I would suggest to add a new option in the CLI where user can exercise the password via eeprom write. keep the default password to CDB.

sonic_platform_base/sonic_xcvr/cdb/cdb.py
Junchao-Mellanox: There is no code logic change in this file. I will revert this.

sonic_platform_base/sonic_xcvr/api/public/cdb_fw.py
Junchao-Mellanox: As we discussed, it needs to fallback to standard memory way if CDB password way does not work. This file is the place to call enter_password, so it should handle the logic like this:

1. Try CDB password way first; otherwise
2. Fall back to EEPROM write password

sonic_platform_base/sonic_xcvr/api/public/cmis.py
Junchao-Mellanox: Currently, there is automatically fallback. Do you think we still need to add an extra option in CLI?

prgeor: @Junchao-Mellanox yes, lets add new option. As mentioned earlier, the EEPROM approach is not reliable  compared to the CDB approach. Let the user decide which option to choose.

sonic_platform_base/sonic_xcvr/api/public/cdb_fw.py
prgeor: @Junchao-Mellanox see my comments below. A new options in the cli can avoid this.

sonic_platform_base/sonic_xcvr/cdb/cdb.py
Junchao-Mellanox: Hi @prgeor , I am fine to add a new CLI option. However, the auto fallback logic is for xcvrd who might periodically call `CmisApi.get_transceiver_info_firmware_versions` which in turn calls `api.get_module_fw_info()`. In this flow, it might require entering password. As this is an auto flow, CLI option could not help here.

sonic_platform_base/sonic_xcvr/api/public/cdb_fw.py
Junchao-Mellanox: @prgeor , there is a auto flow in xcvrd, which cannot rely on the CLI option.

keboliu:
@prgeor  would you please review again?

keboliu (approved):

PR #729, opened 2026-07-31
cmisVDM: cache the static VDM descriptor pages

aditya-nexthop:
#### Description

Cache the static VDM descriptor pages (0x20-0x23) on the `CmisVdmApi` object so
each page is read from EEPROM once instead of on every DOM cycle.

- Adds `CmisVdmApi._read_vdm_descriptor_page(page)`, which returns a raw
  descriptor page and memoizes it in a `self._vdm_descriptor` dict keyed by page.
- `get_vdm_page()` now obtains the descriptor through this helper instead of
  calling `xcvr_eeprom.read_raw()` directly.
- Only a non-empty read is cached, so a transient read failure is retried on the
  next cycle rather than leaving VDM unavailable for the life of the api object.
- Pages are cached independently and populated lazily, so only the descriptor
  pages a module actually advertises are read.

#### Motivation and Context

The VDM descriptors -- observable Type ID, threshold-set ID and monitored-lane
assignment for each of the 64 slots in a page -- are a static, read-only module
advertisement in CMIS. They describe how the VDM value and threshold pages are
laid out and do not change while the module is powered and plugged in, yet
`get_vdm_page()` re-read them on every DOM cycle.

Each descriptor page is a 128 B read and `get_vdm_allpage()` walks every page the
module advertises (up to four, 0x20-0x23), so this removes up to 4 x 128 B of
redundant EEPROM traffic per port per DOM cycle.

The cache lives on the api object, which xcvrd recreates when a module is
re-inserted, so it needs no explicit invalidation.

Full `pytest` suite: 1388 passed, no failures.

New unit tests in `tests/sonic_xcvr/test_cmisVDM.py`:

- `test_vdm_descriptor_page_cached` -- the descriptor page is read once across
  three `get_vdm_page()` cycles, while the value page is re-read each cycle.
- `test_vdm_descriptor_cache_is_per_page` -- pages 0x20 and 0x21 are cached
  independently.
- `test_vdm_descriptor_empty_read_not_cached` -- an empty descriptor read is
  retried rather than cached.

`TestVDM` builds its api in `setup_method` so each test starts with an empty
descriptor cache.

Descriptor caching is unconditional. It is not gated by
`CmisApi.set_cache_enabled()`, which continues to control only the
`read_only_cached_api_return` fields in `cmis.py`.


sonic_platform_base/sonic_xcvr/api/public/cmisVDM.py
```diff
@@ -29,9 +30,35 @@ class CmisVdmApi(XcvrApi):
...
 
+    # Default caching disabled; CmisApi wires cache_enabled onto the instance.
+    cache_enabled = False
+
     def __init__(self, xcvr_eeprom):
         super(CmisVdmApi, self).__init__(xcvr_eeprom)
-    
+
+    @read_only_cached_api_return
+    def _read_vdm_descriptor_page(self, page):
```
prgeor, line 38: @aditya-nexthop do we really need to implement cache for this? we can initialize a self._vdm_descriptor( in the constructor)which will be created every time the CMIS API is instantiated during OIR/reboot.?

prgeor (dismissed):

sonic_platform_base/sonic_xcvr/api/public/cmisVDM.py
aditya-nexthop: We don't need to - we can maintain a local dictionary instead like this:
```
  def __init__(self, xcvr_eeprom):
      super().__init__(xcvr_eeprom)
      self._vdm_descriptor = {}

  def _read_vdm_descriptor_page(self, page):
      offset = page * PAGE_SIZE + PAGE_OFFSET
      if not self.cache_enabled:
          return self.xcvr_eeprom.read_raw(offset, PAGE_SIZE)
      if not self._vdm_descriptor.get(page):        # retry, don't cache a bad read
          self._vdm_descriptor[page] = self.xcvr_eeprom.read_raw(offset, PAGE_SIZE)
      return self._vdm_descriptor[page]
```
Please let me know if you prefer this, and I'll make the change.

prgeor: @aditya-nexthop would prefer this as this is more readable.

aditya-nexthop: fixed

sonic_platform_base/sonic_xcvr/api/public/cmisVDM.py
```diff
@@ -29,9 +29,41 @@ class CmisVdmApi(XcvrApi):
     VDM_OBSERVABLE_STATISTIC = 0x2  # Statistic (min/max/avg) observable types
     VDM_OBSERVABLE_ALL = 0x3        # Both basic and statistic
 
+    # Default caching disabled; CmisApi wires cache_enabled onto the instance.
```
prgeor: @aditya-nexthop shouldn't this be by default cached? I do'nt see any usecase to NOT cache the VDM descriptor

aditya-nexthop: made the change, no need to have a flag for cache_enabled at all!

prgeor (approved):

PR #731, opened 2026-08-04
Provide BMC defaults for system status LED methods in ChassisBase

shreyansh-nexthop:
#### Description

Add BMC defaults for the system status LED methods on `ChassisBase`, gated on `is_switch_bmc()`:
- `initizalize_system_led()` returns `True`
- `set_status_led()` returns `False`
- `get_status_led()` returns `"N/A"`

Non-BMC platforms are unaffected: they still `raise NotImplementedError` (or use their own platform override). The methods use the module-level `device_info` import that `chassis_base.py` already has, and guard it with `if device_info and ...` so they still raise `NotImplementedError` rather than `AttributeError` when `sonic_py_common` is shadowed by a partial mock and `device_info` is `None`.

#### Motivation and Context

`show system-health` (sonic-utilities) calls `chassis.initizalize_system_led()`, `set_status_led()` and `get_status_led()` unconditionally. BMC platforms have no controllable system status LED, and `ChassisBase` left `set`/`get` as `NotImplementedError` with no `initizalize_system_led()` at all, so the CLI failed on BMC images. Handling this once in the common base, instead of duplicating no-op stubs in each vendor's `chassis.py`, keeps the behavior consistent across all BMC platforms.

The method name `initizalize_system_led` intentionally keeps the existing (misspelled) name used by the `show system-health` caller and current platform implementations.

Non-BMC behavior is unchanged (still `NotImplementedError`, since `is_switch_bmc()` is false there). On a BMC image the expected result is that `show system-health summary` no longer errors and reports `System status LED  N/A`.

### DIFF
--- a/sonic_platform_base/chassis_base.py
+++ b/sonic_platform_base/chassis_base.py
@@ -884,6 +884,18 @@ class ChassisBase(device_base.DeviceBase):
     # System LED methods
     ##############################################
 
+    def initizalize_system_led(self):
+        """
+        Initialize the system status LED.
+
+        Returns:
+            bool: True if the system LED was initialized successfully.
+        """
+        if device_info and device_info.is_switch_bmc():
+            # BMC platforms have no controllable system LED, nothing to initialize.
+            return True
+        raise NotImplementedError
+
     def set_status_led(self, color):
         """
         Sets the state of the system LED
@@ -895,6 +907,9 @@ class ChassisBase(device_base.DeviceBase):
         Returns:
             bool: True if system LED state is set successfully, False if not
         """
+        if device_info and device_info.is_switch_bmc():
+            # BMC platforms have no controllable system LED.
+            return False
         raise NotImplementedError
 
     def get_status_led(self):
@@ -905,6 +920,9 @@ class ChassisBase(device_base.DeviceBase):
             A string, one of the valid LED color strings which could be vendor
             specified.
         """
+        if device_info and device_info.is_switch_bmc():
+            # BMC platforms have no controllable system LED.
+            return "N/A"
         raise NotImplementedError
 
     ##############################################
--- a/tests/chassis_base_test.py
+++ b/tests/chassis_base_test.py
@@ -58,6 +58,61 @@ class TestChassisBase:
 
             assert exception_raised
 
+    @mock.patch('sonic_py_common.device_info.is_switch_bmc', return_value=True)
+    def test_system_led_bmc(self, _mock_is_switch_bmc):
+        # BMC platforms have no controllable system LED, so the base class
+        # provides no-op defaults instead of raising NotImplementedError.
+        chassis = ChassisBase()
+        assert(chassis.initizalize_system_led() == True)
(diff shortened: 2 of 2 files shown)


judyjoseph:
@shreyansh-nexthop  is this ready for review, please set once you complete tests - thx

shreyansh-nexthop:
> @shreyansh-nexthop is this ready for review, please set once you complete tests - thx

@judyjoseph, the PR is now ready for review. Thanks!

judyjoseph (approved):
LGTM

judyjoseph:
@oleksandrivantsiv @vivekrnv for a quick review

judyjoseph (dismissed):

judyjoseph:
### 1. Missing `None` guard on `device_info`

`chassis_base.py` deliberately tolerates `device_info` being `None` (when `sonic_py_common` is shadowed by a partial mock). The existing code guards for it at line 102:

```python
cpo_data = device_info.get_cpo_data() if device_info else None
```

There's even a dedicated test (`test_device_info_import_failure`) asserting `chassis_base.device_info is None` works. But the three new methods call `device_info.is_switch_bmc()` unguarded:

```python
if device_info.is_switch_bmc():   # AttributeError if device_info is None
```

If `device_info is None`, these raise `AttributeError` instead of the intended `NotImplementedError`. Suggested fix:

```python
if device_info and device_info.is_switch_bmc():
```

### 2. PR description doesn't match the code

The description states that `device_info` is imported lazily inside each method, matching the existing pattern in `module_base.py` and `bmc_base.py`. The code actually uses the module-level import (line 10), not a lazy per-method import. `module_base.py`/`bmc_base.py` genuinely do lazy imports (lines 516, 282); this PR does not. Either update the description, or switch to lazy imports (which would also side-step issue 1 by re-importing fresh each call).

### 3. Test gap

Neither new test exercises the `device_info is None` case for these methods — worth adding if you apply fix 1.

shreyansh-nexthop:
Thanks @judyjoseph for the review, I have taken care of all the 3 comments.

yxieca (approved):
Reviewed. Verified @judyjoseph's 3 change requests are addressed at head 9c81faf1b:

1. `device_info` None-guard — all three methods now use `if device_info and device_info.is_switch_bmc():` (chassis_base.py:894/910/923), so they raise NotImplementedError rather than AttributeError when `device_info` is None.
2. Import approach reconciled — module-level import with try/except None-fallback retained, and the None-guard resolves the description/code contradiction.
3. Test coverage added — `test_system_led_no_device_info` (device_info=None) plus `test_system_led_bmc` / `test_system_led_non_bmc` cover the full matrix.

The `initizalize_system_led` spelling matches the established platform-API contract (caller show/system_health.py:40; 90 in-tree definitions vs 0 correctly-spelled), so it must stay as-is. LGTM.

Note: no blocking issues. Backport to msft-202608 still needs the Tested-branch/Test-result info populated per the mssonicbld bot before cherry-pick.

judyjoseph (approved):

PR #735, opened 2026-08-13
Add vendor-independent leak test API

chinmoy-nexthop:
Add a common API for injecting a simulated leak into the leak detection path, so the reporting chain can be validated without wetting hardware.

Injection is non-destructive: an injected leak is published like any other leak, and is additionally flagged through LeakageSensorBase is_test_leak() so consumers must not take a mitigation action on it.

- leakage_sensor_test_base.py: new LeakageSensorTestBase defining is_leak_test_supported(), set_test_leak(), is_test_leak_enabled() and clear_test_leaks()
- liquid_cooling_base.py: add is_test_leak() on LeakageSensorBase and get_leak_sensor_test() on LiquidCoolingBase, defaulting to False and None so platforms without injection support are unaffected
- tests/leak_test_api_base.py: LeakTestApiBase, a reusable conformance suite a platform subclasses to validate its implementation against the common contract


judyjoseph:
@fraserg-arista to review as well, this could be first step on https://github.com/sonic-net/SONiC/pull/2441

nikamirrr (commented):
_(Duplicate submission from a retried API call — superseded by the review below, which carries all 15 inline comments. Please disregard this entry.)_

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -87,16 +88,32 @@ def get_leak_severity(self) -> LeakSeverity:
...
+        from a test injection rather than from the hardware.
+
+        A test leak is observable end to end (it is published like any other
+        leak) but must never be used to trigger a mitigation action.
+
+        Returns:
+            bool: True if the reported leak was injected by a leak test,
+                  False otherwise
+        """
+        return self.test_leak
```
nikamirrr: **`is_test_leak()` is decoupled from `is_leak()`, so a stale flag suppresses mitigation on a *real* leak.**

The docstring says this reports "whether the leak **currently reported by this sensor**
originates from a test injection", but the body never consults `self.leaking`. Once
`test_leak` is set it stays set until something explicitly resets it.

Verified by running it: inject on `s1`, take a withdraw path that clears `leaking` but not
`test_leak`, then let a genuine leak arrive on `s1`:

```
real leak -> is_leak() True   is_test_leak() True   => mitigation suppressed
```

A consumer written to the contract in this PR (`if s.is_leak() and not s.is_test_leak(): mitigate()`)
now skips power-off on actual coolant. Deriving the flag instead of storing it independently
removes the whole class of stale-state bugs:

```suggestion
        return self.leaking and self.test_leak
```

tests/leakage_sensor_test_base_test.py
```diff
@@ -0,0 +1,113 @@
...
+                sensor.leaking = enable
+                sensor.test_leak = enable
+                sensor.leak_severity = severity if enable else None
+                return True
+
+            def is_test_leak_enabled(self, sensor_name):
+                sensor = self._sensors.get(sensor_name)
+                return bool(sensor and sensor.test_leak)
+
+            def clear_test_leaks(self):
```
nikamirrr: **This reference `clear_test_leaks()` erases *real* leaks, and its `return True` is unconditional.**

This is the only worked implementation the PR ships, so it is what vendors will copy.
Two problems:

1. It loops over **every** sensor and calls `set_test_leak(name, False)`, which at line 74
   assigns `sensor.leaking = enable` — clearing sensors that never had an injection.
   Verified by running it: `sensor2` holds a genuine CRITICAL leak, a test leak is injected
   on `sensor1`, then `clear_test_leaks()` runs:

   ```
   before: leaking sensors ['sensor1', 'sensor2']
   after : leaking sensors []
           sensor2.is_leak() False   get_leak_severity() None
   ```

   Running the leak test disarms real leak detection. `leak_severity` is also left at `None`
   for the sensor's lifetime, so a later genuine leak reports `None` and any caller doing
   `.value` on it raises `AttributeError`.

2. Line 83 returns a literal `True` while discarding every per-sensor result, contradicting the
   ABC docstring ("True if all test leaks were withdrawn, False otherwise"). `leak_test_api_base.py:153`
   asserts `== True`, so that assertion can never catch it.

Withdrawal should only ever touch sensors this object actually injected, and should restore the
prior severity rather than `None`.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -87,16 +88,32 @@ def get_leak_severity(self) -> LeakSeverity:
         Retrieves the severity of leak
 
         Returns:
-            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None if no leak
+            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None
+            if no leak
         """
         return self.leak_severity
 
+    def is_test_leak(self) -> bool:
```
nikamirrr, line 104: **The "never trigger a mitigation action" contract has no enforcement point anywhere, and the
default path simulates the worst case.**

`is_test_leak()` is added here but has zero consumers, and no field carries it into STATE_DB, so
nothing downstream can distinguish an injected leak from a real one:

- `sonic-utilities` `show/platform.py` `leak_status` reads only
  `name` / `leaking` / `leak_sensor_status` / `type` / `leak_severity`.
- `sonic-buildimage` `src/system-health/health_checker/hardware_checker.py:341` reads only
  `leak_status`, then calls `set_object_not_ok(...)` and `publish_events(..., "sensor report leaking event")`
  under tag `liquid-cooling-leak`.
- `sonic-buildimage` `src/sonic-yang-models/yang-models/sonic-leak-control.yang:55-58`:

  ```
  leaf system_critical_leak_action {
      type leak-action;
      default power_off;
  ```

`set_test_leak`'s severity default is `LeakSeverity.CRITICAL`, and four of the five injection
call sites in `leak_test_api_base.py` omit the argument — so the default, least-thought path
injects the highest severity into a chain whose default action is to power the host off.
A "non-destructive" test powers the switch off.

Could the flag be plumbed into the published leak record (and honoured by the health checker /
leak policy) in this PR, before injection ships? `LeakSeverity.MINOR` would also be a much safer
default for a simulated leak.

tests/leak_test_api_base.py
```diff
@@ -0,0 +1,156 @@
...
+                assert leak_test.is_test_leak_enabled(name) == True
+
+                leaking = liquid_cooling.get_leak_sensor_status()
+                assert len(leaking) == 1
+                assert leaking[0].get_name() == name
+                assert leaking[0].is_leak() == True
+                assert leaking[0].is_test_leak() == True
+                assert leaking[0].get_leak_severity() is severity
+
+                leak_test.clear_test_leaks()
```
nikamirrr: **The conformance suite mutates the platform with no teardown, so a failing assertion leaves the
DUT injected.**

`clear_test_leaks()` sits *after* the assertions here (and at 123, 153) rather than in a fixture or
`try/finally`, and `test_withdraw_single_injection` never clears at all. If any assert at 98-107
fails — precisely what this suite exists to detect — line 109 never runs.

Because a correct implementation has to persist the injection outside the process, the box keeps
publishing a CRITICAL leak after pytest exits, and the configured `power_off` action can fire
minutes later on an unattended lab machine. Within the run it also cascades:
`test_no_leak_before_injection` then fails on stale state, even though `get_liquid_cooling()` is
documented at line 41 as returning a fresh object with no test leak injected.

An autouse yield-fixture that calls `clear_test_leaks()` in teardown would cover all four sites and
let the tests drop the trailing cleanup entirely.

sonic_platform_base/leakage_sensor_test_base.py
```diff
@@ -0,0 +1,71 @@
...
+        """
+        Retrieves whether this platform supports leak test injection.
+
+        Returns:
+            bool: True if leak test injection is supported, False otherwise
+        """
+        return False
+
+    @abstractmethod
+    def set_test_leak(self, sensor_name: str, enable: bool,
```
nikamirrr, line 41: **Nothing here requires a platform's `is_leak()` override to honour the injection — and on both
shipping platforms the PR's injection model is a silent no-op.**

The reference implementation injects by assigning `sensor.leaking` / `sensor.leak_severity`. Neither
in-tree platform reads those attributes:

- `sonic-buildimage` mlnx-platform-api `sonic_platform/liquid_cooling.py` — `LeakageSensor.is_leak()`
  re-reads sysfs and reassigns `self.leaking` on every call, wiping the injection before
  `get_leak_sensor_status()` sees it.
- `sonic-buildimage` nvidia-bmc ast2700 `sonic_platform/leakage_sensor.py:191-201` — `is_leak()`,
  `is_leak_sensor_ok()` and `get_leak_severity()` are computed purely from `_check_channel_value()`
  and never read those attributes at all.

Result on both: `is_test_leak_enabled()` returns True while `get_leak_sensor_status()` returns `[]`,
so `leak_test_api_base.py:103` fails on `assert len(leaking) == 1`.

There is also a process-boundary gap: a CLI or pytest process mutating its own object graph never
reaches the daemon that polls `LiquidCooling` and writes STATE_DB, so the reporting chain this PR
exists to validate is never actually exercised.

Could the ABC state where the injected state must live (sysfs / BMC / DB) and that `is_leak()`,
`get_leak_severity()` and `is_test_leak()` must all report it? Without that, "conforming" is
undefined and each vendor will pick a different layer.

sonic_platform_base/leakage_sensor_test_base.py
```diff
@@ -0,0 +1,71 @@
...
+            enable: True to inject the test leak, False to withdraw it
+            severity: severity the injected leak is reported with
+
+        Returns:
+            bool: True if the test leak state was applied, False otherwise
+        """
+        pass
+
+    @abstractmethod
+    def is_test_leak_enabled(self, sensor_name: str) -> bool:
```
nikamirrr, line 58: **Two independent sources of truth for one fact, and the conformance suite cannot detect drift
between them.**

This registry-style query and `LeakageSensorBase.is_test_leak()` both answer "is a test leak
injected on this sensor", with nothing keeping them in sync.

I wrote an implementation that is wrong in exactly the dangerous way — tracks injections in its own
set and, on withdraw, does `self._inj.discard(name); sensor.leaking = False` but forgets
`sensor.test_leak = False`. It passes the entire suite:

```
(A) sloppy impl that leaves sensor.test_leak stale: 8 pass / 0 fail
    real leak -> is_leak() True  is_test_leak() True  => mitigation suppressed: True
```

The post-withdrawal assertions only ever check `is_test_leak_enabled()` (the registry) and
`get_leak_sensor_status() == []` — which by construction returns only *leaking* sensors, so it can
never observe a stale flag on a non-leaking one. That platform ships, and the first real leak on
that sensor is silently exempted from mitigation.

Two fixes worth considering: make `is_test_leak_enabled()` a concrete base method that reads the
sensor (one source of truth), and add `assert sensor.is_test_leak() == False` after every withdrawal
in the suite.

sonic_platform_base/leakage_sensor_test_base.py
```diff
@@ -0,0 +1,71 @@
...
+        Retrieves whether this platform supports leak test injection.
+
+        Returns:
+            bool: True if leak test injection is supported, False otherwise
+        """
+        return False
+
+    @abstractmethod
+    def set_test_leak(self, sensor_name: str, enable: bool,
+                      severity: LeakSeverity = LeakSeverity.CRITICAL) -> bool:
```
nikamirrr, line 42: **A default value on an `@abstractmethod` is documentation only — Python does not inherit it into
the override, and `abc` does not check signatures.**

A platform writing the natural override gets no warning:

```python
def set_test_leak(self, sensor_name, enable, severity):   # no default
```

Verified: every 2-argument call then fails with

```
TypeError: NoDefault.set_test_leak() missing 1 required positional argument: 'severity'
```

`LeakTestApiBase` uses the 2-arg form at lines 85, 118, 134, 139 and 149, so four of the eight
conformance tests ERROR with a signature crash instead of a meaningful assertion — and any future
CLI caller crashes at runtime on that platform.

Making `severity` keyword-only, and/or enforcing the default in a concrete wrapper that delegates to
an abstract `_set_test_leak`, would make the default actually binding.

sonic_platform_base/leakage_sensor_test_base.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Provides a vendor-independent way to inject a simulated leak so that the
+    leak reporting path can be validated without wetting hardware. Injection is
+    non-destructive: a leak reported through this interface is flagged as a
+    test leak and must not cause any mitigation action to be taken.
+
+    Reached through LiquidCoolingBase.get_leak_sensor_test().
+    """
+
+    def is_leak_test_supported(self) -> bool:
```
nikamirrr, line 25: **A fully-conforming implementation reports itself unsupported, and "unsupported" is now encoded
twice with no stated precedence.**

This method is concrete and returns `False`, while its three siblings are `@abstractmethod`. So a
subclass that implements all three required methods correctly still reports `False` unless it also
remembers to override this one — verified:

```
fully-implemented subclass reports supported = False
```

`leak_test_api_base.py:68` hard-asserts `is_leak_test_supported() == True`, so a correct vendor
implementation fails conformance. Conversely a caller that only null-checks
`get_leak_sensor_test()` — the *other* encoding of the same fact, at `liquid_cooling_base.py:220` —
proceeds on a platform that declared no support.

The `False` default is also unreachable in the shape it seems designed for:
`class NoSupport(LeakageSensorTestBase): pass` raises `TypeError`, which is why
`leakage_sensor_test_base_test.py:31-42` has to stub all three methods just to observe the default.

Deriving support from `get_leak_sensor_test() is not None` would leave one source of truth.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -35,6 +35,7 @@ def __init__(self,
         self.leak_type: str = type or 'unknown'
         self.leak_location: str = location or 'unknown'
         self.leak_severity = severity
+        self.test_leak: bool = False
```
nikamirrr: **`test_leak` is instance-only, so the new public `is_test_leak()` raises `AttributeError` on any
subclass that does not chain `super().__init__()`.**

Verified — a subclass that sets `name` / `leaking` / `leak_sensor_ok` / `leak_type` /
`leak_location` / `leak_severity` itself (mirroring this `__init__`) gives:

```
AttributeError: 'NoSuper' object has no attribute 'test_leak'
```

The two in-tree platforms chain today, but this is a public vendor-plugin base class with
out-of-tree implementations; `LeakageSensorBase` itself never calls `super().__init__()` on
`SensorBase`, so there is no chaining precedent in the file; and ast2700's `LeakageSensor` already
overrides every other accessor.

This repo's own `.github/copilot-instructions.md` is explicit about it:

> **Backwards compatibility**: CRITICAL — don't break existing vendor implementations
> **Default implementations**: New methods MUST have safe defaults (return None, 'N/A', False, etc.)

Declaring `test_leak: bool = False` as a class attribute (next to `LEAK_SEVERITY_CRITICAL`) keeps the
instance assignment working and makes the accessor safe for every subclass.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -191,6 +208,17 @@ def get_all_profiles(self) -> List[LeakSensorProfileBase]:
...
+        Retrieves the leak test interface of this platform.
+
+        Returns:
+            LeakageSensorTestBase: the leak test interface defined in
+            leakage_sensor_test_base, or None if the platform does not support
+            leak test injection
+        """
+        return None
+
     def get_profile(self, type: str) -> LeakSensorProfileBase|None:
```
nikamirrr, line 230: **(line 226, just below) `get_profile()` calls `getattr()` on a dict, so this API is entirely
non-functional.**

`self.profiles` is a `Dict[str, LeakSensorProfileBase]` (line 154), but line 226 is:

```python
profile = getattr(self.profiles, type, None)
```

`getattr` looks up *attributes*, not keys. Verified with a `rope` profile registered:

- `get_profile('rope')` returns `None` and writes `Leakage sensor profile rope doesn't exist` to stderr
- `get_profile('items')` returns `<built-in method items of dict object>` (same for `keys`/`get`/`values`/`pop`),
  so `lc.get_profile(t).get_leak_max_minor_duration_sec()` raises `AttributeError` instead of the documented `None`

Should be `self.profiles.get(type)`.

Pre-existing rather than introduced here, but this PR inserts `get_leak_sensor_test()` directly above
it, and neither the existing tests nor the new conformance suite covers `get_profile` — worth fixing
while the file is open.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -87,16 +88,32 @@ def get_leak_severity(self) -> LeakSeverity:
         Retrieves the severity of leak
 
         Returns:
-            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None if no leak
+            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None
```
nikamirrr, line 99: **This docstring says "None if no leak", but a non-leaking sensor returns `CRITICAL`.**

`__init__` defaults `severity` to `LeakSeverity.CRITICAL` (line 37) and nothing ever clears it, so:

```
LeakageSensorBase('s1') -> is_leak() False, get_leak_severity() LeakSeverity.CRITICAL
```

Any consumer using severity as a leak indicator (`if s.get_leak_severity() is not None:`) escalates
every healthy sensor to CRITICAL. The `-> LeakSeverity` annotation also contradicts the "or None"
text.

The contradiction is baked into the tests too — `liquid_cooling_base_test.py:126-131` is docstring'd
"Test get_leak_severity default value is None" while asserting `is LeakSeverity.CRITICAL`. And both
real platforms follow the docstring rather than the base (ast2700 returns `None` when there is no
leak), so base and vendors disagree on the same documented API.

Since this PR is already reflowing these lines: either annotate `-> LeakSeverity|None` and have the
base return `None` when `not self.leaking`, or drop the "or None" claim.

nikamirrr (commented):
Automated review pass focused on correctness. Fifteen inline comments below.

Findings 1, 2, 6, 7, 8, 9, 10, 11, 14 and 15 were confirmed by **executing** the code rather than
by reading it — including running the full `LeakTestApiBase` suite against both a correct and a
deliberately-sloppy implementation.

The through-line is the PR's central safety promise: *an injected leak is published like a real one
but must never trigger a mitigation action*. Three independent gaps each break it on their own:

1. `is_test_leak()` is decoupled from `is_leak()`, so a stale flag exempts a **real** leak from
   mitigation.
2. The reference `clear_test_leaks()` erases **real** leaks on sensors that were never injected.
3. The flag reaches no consumer and no STATE_DB field, while `set_test_leak` defaults to `CRITICAL`
   and `system_critical_leak_action` defaults to `power_off` — so a "non-destructive" test can power
   the switch off.

The conformance suite cannot catch any of the three: I built an implementation that is wrong in
exactly the dangerous way and it passes 8/8.

Also worth a look, below the cut: a shared mutable default `leakage_sensors_list=[]` in
`LiquidCoolingBase.__init__`; `test_injection_is_non_destructive` being a tautology that passes when
injection is a no-op; and `@abstractmethod ... pass` against this repo's own written rule in
`.github/copilot-instructions.md` ("Abstract methods: Raise `NotImplementedError` in base class"),
which also makes `super().set_test_leak(...)` silently return a falsy `None`.

sonic_platform_base/leakage_sensor_test_base.py
nikamirrr, line 41: **The whole injection API is name-keyed, but the model underneath has no sensor-name uniqueness
invariant.**

`LiquidCoolingBase` only ever addresses sensors by index (`get_leak_sensor(index)`) and never checks
for duplicate names. Meanwhile nvidia-bmc's `LeakageSensor` takes its name from the sysfs
`channel_name` file (`ast2700/sonic_platform/leakage_sensor.py:56-59`), so two `(a2d, channel)` pairs
can legitimately carry the same label.

With two sensors both named `leakage`, the reference implementation's
`{s.get_name(): s for s in sensors}` silently collapses 2 entries into 1:

- `set_test_leak('leakage', True)` returns `True` but injects only the second sensor
- `is_test_leak_enabled('leakage')` returns `True`, hiding the partial injection
- `clear_test_leaks()` can never reach the shadowed sensor — its state is unclearable
- `test_clear_all_injections` then fails at line 150 with `len(status) == 1 != 2`, with no hint that
  a name collision is the cause

Either add `LiquidCoolingBase.get_leak_sensor_by_name()` with a documented uniqueness invariant, or
key injection by index like every other leak accessor.

tests/leak_test_api_base.py
```diff
@@ -0,0 +1,156 @@
...
+
+The file name deliberately avoids the *_test.py / test_* patterns so pytest
+does not collect this module directly — only the platform subclasses run.
+'''
+
+from sonic_platform_base.leakage_sensor_test_base import LeakageSensorTestBase
+from sonic_platform_base.liquid_cooling_base import LeakSeverity
+
+
+class LeakTestApiBase:
```
nikamirrr: **This "reusable conformance suite" is executed zero times by CI and cannot be imported by the
platform repos it was written for.**

Nothing in this repo subclasses `LeakTestApiBase`, and the module docstring's collection-avoidance
works a little too well: the filename matches neither `test_*.py` nor `*_test.py`, and the class name
does not match pytest's `python_classes = Test*`. So none of these 156 lines run — a typo or an
inverted assertion in the suite merges green.

For the downstream use case, `setup.py` uses an explicit `packages=[...]` list with no `tests` entry
and there is no `MANIFEST.in`, so the wheel omits this file entirely:
`from tests.leak_test_api_base import LeakTestApiBase` raises `ModuleNotFoundError`.

Adding `tests` to `packages` would not fix it either — mlnx-platform-api and ast2700 both already own
the top-level `tests` package name and import as `tests.*`, so the consumer's own package shadows it.

Moving the suite under `sonic_platform_base/` and adding one in-repo subclass would fix both halves.
The `PlatformLiquidCooling` fake already in `leakage_sensor_test_base_test.py:85` would do it in ~5
lines — and would let you delete the ~17 duplicated assertions at lines 97-113, which are a subset of
what this suite already checks.

tests/leak_test_api_base.py
```diff
@@ -0,0 +1,156 @@
...
+    '''
+    Collection of leak test API conformance checks.
+
+    Subclasses must override get_liquid_cooling() and SENSOR_NAMES.
+    '''
+
+    # Names of the leak sensors the platform exposes, in order.
+    SENSOR_NAMES = []
```
nikamirrr: **A `[]` default here turns "subclass forgot to override" into a mix of silently-vacuous PASSes and
bare `IndexError`s.**

Verified against a subclass that does not set `SENSOR_NAMES`:

- `test_injection_is_non_destructive` (line 118) and `test_withdraw_single_injection` (line 132)
  raise `IndexError: list index out of range` on `self.SENSOR_NAMES[0]`, from inside the shared base
  class, with no hint that the subclass is misconfigured
- `test_inject_every_severity` (line 93) and `test_clear_all_injections` (lines 148, 154) iterate an
  empty list, execute **zero** assertions, and report PASS

So a platform gets two confusing errors alongside two green "injection works" results that asserted
nothing.

Note `get_liquid_cooling()` gets a clear `NotImplementedError` at line 41 while the other half of the
required override gets no guard at all. `SENSOR_NAMES = None` plus an autouse fixture asserting it
was set would fail loudly and consistently.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -191,6 +208,17 @@ def get_all_profiles(self) -> List[LeakSensorProfileBase]:
         """
         return list(self.profiles.values())
 
+    def get_leak_sensor_test(self) -> 'LeakageSensorTestBase|None':
```
nikamirrr, line 219: **This forward reference can never be resolved — `LeakageSensorTestBase` is not imported into this
module and there is no `TYPE_CHECKING` guard.**

Verified:

```
typing.get_type_hints(LiquidCoolingBase.get_leak_sensor_test)
-> NameError: name 'LeakageSensorTestBase' is not defined
```

The two sibling annotations in this same file resolve fine (`get_leak_sensor -> LeakageSensorBase|None`,
`get_profile -> LeakSensorProfileBase|None`), so this one method is the odd one out. mypy/pyright
report the same undefined name and silently degrade the return type to `Any`, and anything that
resolves hints at runtime — sphinx `autodoc_typehints`, `inspect.signature(eval_str=True)`, or a
future `from __future__ import annotations` migration — breaks here.

A plain import would be circular (`leakage_sensor_test_base.py:10` imports `LeakSeverity` from this
module), so the standard fix applies:

```python
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from .leakage_sensor_test_base import LeakageSensorTestBase
```

chinmoy-nexthop:
> _(Duplicate submission from a retried API call — superseded by the review below, which carries all 15 inline comments. Please disregard this entry.)_

@nikamirrr  Thank you. 
It is absolutely fine. Maybe your agent was a little too verbose. I have looked into your feedback and taken care of a few things that I felt were appropriate.

Addressed the following:- 
- test_leak is now a class attribute, so is_test_leak() is safe on subclasses that don't chain super().__init__()
- is_test_leak() only reports True while the sensor is actually leaking, so a stale flag can't mask a real leak
- get_leak_severity() returns None when there's no leak, as documented
- Fixed get_profile() — it was using getattr() on a dict and could never find a profile
- Added a TYPE_CHECKING import so the LeakageSensorTestBase annotation resolves
- Moved the conformance suite into sonic_platform_base/ so it ships in the wheel, and added an in-repo subclass so it runs in CI
- Suite now clears injected leaks on teardown even when an assertion fails, fails loudly if SENSOR_NAMES isn't set, and verifies is_test_leak() is False after withdrawal
- Reference clear_test_leaks() only touches injected sensors, restores prior sensor state, and returns the aggregated result

nikamirrr (commented):
### Code review

Reviewed the new leak test API. Most of this is contract and robustness feedback rather than anything CI will catch — I ran `tests/liquid_cooling_base_test.py` and `tests/leakage_sensor_test_base_test.py` against 9ef9f23 and they pass (43 tests, all 8 conformance tests collected and green). Details inline.

One finding isn't on a changed line, so I couldn't leave it inline. `LiquidCoolingBase.__init__` takes mutable default arguments and assigns them straight to instance state:

https://github.com/sonic-net/sonic-platform-common/blob/9ef9f23afab5714c11dba89792e14c9bf86d1b13/sonic_platform_base/liquid_cooling_base.py#L155-L161

Every instance constructed without explicit lists shares one list object. This is pre-existing, but the new `test_get_profile` is the first thing to exercise it — it does `LiquidCoolingBase(profiles=[profile])`, whose `self.leakage_sensors` *is* the shared default. Anything that later appends to a `get_all_leak_sensors()` result would leak sensors into every other instance in the process. `= None` plus `or []` closes it.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -24,6 +29,10 @@ class LeakageSensorBase(SensorBase):
     LEAK_SEVERITY_CRITICAL = LeakSeverity.CRITICAL
     LEAK_SEVERITY_MINOR    = LeakSeverity.MINOR
 
+    # Class-level default so is_test_leak() is safe on vendor subclasses that
+    # do not chain super().__init__()
+    test_leak: bool = False
```
nikamirrr, line 34: This is the only `LeakageSensorBase` attribute declared as a class-level default — `leaking`, `leak_sensor_ok`, `leak_type`, `leak_location` and `leak_severity` are all set in `__init__`. The justification in the comment (safe on vendor subclasses that don't chain `super().__init__()`) applies equally to those five, and the new `test_is_test_leak_without_init_chaining` has to hand-set all of them precisely because they aren't class defaults. That leaves two initialization conventions in one class, and it means `LeakageSensorBase.test_leak = True` (class rather than instance) flips every sensor in the process. Setting it in `__init__` alongside the rest would be more consistent.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -82,21 +91,37 @@ def get_leak_sensor_location(self) -> str:
...
+    def get_leak_severity(self) -> LeakSeverity|None:
         """
         Retrieves the severity of leak
 
         Returns:
-            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None if no leak
+            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None
+            if no leak
+        """
+        return self.leak_severity if self.is_leak() else None
```
nikamirrr, line 102: No objection to returning `None` — the previous docstring already promised it, so this makes the implementation match. The concern is that `get_leak_severity()` now calls `is_leak()`, which on vendor subclasses is a hardware read rather than an attribute lookup:

- `platform/mellanox/mlnx-platform-api/sonic_platform/liquid_cooling.py` reads the sysfs file and mutates `self.leaking` as a side effect.
- `platform/aspeed/sonic-platform-modules-nvidia-bmc/ast2700/sonic_platform/leakage_sensor.py` re-reads the channel `input` file on every call.

Two consequences. A consumer doing `get_leak_sensor_status()` and then `get_leak_severity()` re-reads hardware between the two, so if the leak cleared in between it gets `None` and `.value` on the result raises. And Mellanox's `is_leak()` returns the string `'N/A'` on a failed read, which is truthy — so an unreadable sensor now reports its last-known severity rather than `None`. Reading `self.leaking` preserves the intent without the extra I/O or the truthiness trap.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -82,21 +91,37 @@ def get_leak_sensor_location(self) -> str:
...
+
+    def is_test_leak(self) -> bool:
```
nikamirrr, line 104: Same read-amplification point as above: with both this and `get_leak_severity()` routing through `is_leak()`, classifying a single leak now costs three hardware reads on aspeed where it previously cost one — and the three readings aren't guaranteed consistent with each other. Caching one evaluation per poll, or reading `self.leaking`, avoids it.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -82,21 +91,37 @@ def get_leak_sensor_location(self) -> str:
...
+
+        A test leak is observable end to end (it is published like any other
+        leak) but must never be used to trigger a mitigation action.
+
+        Returns:
+            bool: True if the reported leak was injected by a leak test,
+                  False otherwise
         """
-        return self.leak_severity
+        return self.is_leak() and self.test_leak
```
nikamirrr, line 116: The annotation is `-> bool`, but `self.is_leak() and self.test_leak` returns whatever `is_leak()` returned when that value is falsy. Neither vendor implementation trips this today - mellanox returns a real `False` on its no-leak branch and aspeed returns a bool - so this is latent rather than an active bug. But a subclass whose `is_leak()` returns a falsy non-bool (`None`, `0`) would make `is_test_leak()` return that value, and `assert sensor.is_test_leak() == False` in the conformance suite then fails on `None` with a message pointing at the test rather than at the sensor read error. Wrapping in `bool()` costs nothing.

(To correct an earlier version of this comment: mellanox's `'N/A'` is *truthy*, so it yields `self.test_leak` here rather than escaping as a non-bool. The truthiness problem shows up in `get_leak_severity()` instead, which I flagged above.)

sonic_platform_base/leak_test_api_base.py
```diff
@@ -0,0 +1,173 @@
+'''
```
nikamirrr, line 1: Verification in this module is 23 bare `assert` statements against a single `raise`. `assert` is the one statement the compiler is allowed to delete: under `python -O`, or with `PYTHONOPTIMIZE` set in the environment, asserts are not compiled into the bytecode at all. Running a reduced copy of `assert_no_leaks()` against a deliberately broken sensor:

```
normal interpreter   ->  FAILED (assertion fired)
python -O            ->  PASSED
PYTHONOPTIMIZE=1     ->  PASSED

assert_no_leaks: 21 bytecode instructions normally, 3 under -O
```

The helper's body is gone. Since `assert_no_leaks()` is the shared check behind several tests, most of the suite's coverage goes with it, and a platform whose `set_test_leak()` does nothing at all would still produce a green conformance run.

A file under `tests/` is only ever run by a developer typing `pytest`, which never sets `-O`. This one ships in the wheel and runs inside vendor build systems and CI harnesses, where an optimized interpreter is entirely plausible - so it is worth not depending on that.

Suggested fix - one helper, and route the checks through it:

```python
    @staticmethod
    def _check(condition, message):
        '''
        Raises on failure. Deliberately not an assert: this module runs in
        vendor environments where the interpreter may be started with -O,
        which strips assert statements entirely.
        '''
        if not condition:
            raise AssertionError(message)
```

so that, for example:

```python
    def assert_no_leaks(self):
        self._check(self.liquid_cooling.get_leak_sensor_status() == [],
                    "sensors report a leak before injection")
        for sensor in self.liquid_cooling.get_all_leak_sensors():
            self._check(sensor.is_leak() == False,
                        f"{sensor.get_name()} reports a leak")
            self._check(sensor.is_test_leak() == False,
                        f"{sensor.get_name()} reports a test leak")
        for name in self.SENSOR_NAMES:
            self._check(self.leak_test.is_test_leak_enabled(name) == False,
                        f"{name} has an armed injection")
```

`raise` survives `-O`, and naming the sensor in the message beats a bare assert's repr when a platform is debugging its own implementation.

sonic_platform_base/leak_test_api_base.py
```diff
@@ -0,0 +1,173 @@
...
+    '''
+    Collection of leak test API conformance checks.
+
+    Subclasses must override get_liquid_cooling() and SENSOR_NAMES.
+    '''
+
+    # Names of the leak sensors the platform exposes, in order. Deliberately
+    # None rather than an empty list, so a subclass that forgets to declare
+    # its sensors fails loudly instead of passing vacuously.
+    SENSOR_NAMES = None
```
nikamirrr, line 44: `tests/leakage_sensor_test_base_test.py:150` builds its sensors from `self.SENSOR_NAMES`, so in-repo `test_sensor_names` compares the list against itself and can never fail - the CI run of the suite does not exercise that check. Declaring them separately fixes it:

```python
    def get_liquid_cooling(self):
        # Declared independently of SENSOR_NAMES so test_sensor_names compares
        # two separately derived lists, as it does on a real platform.
        sensors = [LeakageSensorBase("sensor1"), LeakageSensorBase("sensor2")]
        return PlatformLiquidCooling(sensors)
```

sonic_platform_base/leak_test_api_base.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def setup_method(self, method):
+        '''
+        Builds the platform object each test runs against, failing loudly
+        when the platform subclass forgets to declare its sensor names.
+        '''
+        assert self.SENSOR_NAMES, \
+            "platform test must override SENSOR_NAMES with the leak sensor names"
+        self.liquid_cooling = self.get_liquid_cooling()
+        self.leak_test = self.liquid_cooling.get_leak_sensor_test()
```
nikamirrr, line 65: No check that this returned a non-`None` object. The base `get_leak_sensor_test()` returns `None` (`liquid_cooling_base.py:228`), and `teardown_method` on line 74 already guards `if self.leak_test is not None` — so the case is known to be reachable.

A platform that subclasses this without overriding `get_leak_sensor_test()` gets six `'NoneType' object has no attribute ...` failures with no indication of the actual cause. An assert here, next to the `SENSOR_NAMES` one, would say what's actually wrong.

sonic_platform_base/leak_test_api_base.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def teardown_method(self, method):
+        '''
+        Withdraws injected leaks on the same object the test ran against,
+        even when an assertion failed, so a failing test cannot leave the
+        platform reporting a leak.
+        '''
+        if self.leak_test is not None:
+            self.leak_test.clear_test_leaks()
```
nikamirrr, line 74: The return value is discarded. If `clear_test_leaks()` returns `False` because one sensor's cleanup failed, the suite still reports success, the next test's `setup_method` runs against a switch with an injected leak still latched, and after the last test the platform is left reporting a `CRITICAL` leak.

The docstring claims "a failing test cannot leave the platform reporting a leak" — asserting this return value (or at minimum logging it) is what makes that true.

sonic_platform_base/leak_test_api_base.py
```diff
@@ -0,0 +1,173 @@
...
+        assert self.liquid_cooling.get_num_leak_sensors() == \
+            len(self.SENSOR_NAMES)
+
+    def test_leak_test_interface_exposed(self):
+        '''
+        Test get_leak_sensor_test returns a LeakageSensorTestBase that reports
+        injection support
+        '''
+        assert isinstance(self.leak_test, LeakageSensorTestBase)
+        assert self.leak_test.is_leak_test_supported() == True
```
nikamirrr, line 103: `leakage_sensor_test_base.py` says a platform with conditional support may override `is_leak_test_supported()` to return `False`. This asserts it is `True`, so such a platform hard-fails here - and so do the other seven tests, since injection genuinely is not available.

Correcting my own earlier suggestion of `pytest.skip`: this module deliberately does not import pytest. `unittest.SkipTest` is stdlib and pytest honours it natively (verified on 9.1.1), so raising it from `setup_method` skips all eight in one place:

```python
        if not self.leak_test.is_leak_test_supported():
            raise unittest.SkipTest(
                "platform reports leak test injection is not supported")
```

`test_leak_test_interface_exposed` then keeps only its `isinstance` check. `teardown_method` correctly does not run on a skip, and `raise` survives `-O`.

sonic_platform_base/leak_test_api_base.py
```diff
@@ -0,0 +1,173 @@
...
+        Test a single injection can be withdrawn, and withdrawing an already
+        cleared sensor is not an error
+        '''
+        name = self.SENSOR_NAMES[0]
+
+        self.leak_test.set_test_leak(name, True)
+        assert self.leak_test.set_test_leak(name, False) == True
+        self.assert_no_leaks()
+
+        assert self.leak_test.set_test_leak(name, False) == True
```
nikamirrr, line 161: Line 161 is a second withdrawal on an already-clear sensor. The abstract contract says only "True if the test leak state was applied", and "applied" reads two ways: post-condition (no-op returns True, what this requires) or transition (no-op returns False, what `if sensor.test_leak == enable: return False` gives). The docstring is explicit about the neighbouring unknown-sensor case, so the omission looks deliberate to a vendor.

Correcting myself: I said the same applied to `clear_test_leaks()` at line 172, but that call does have injections to clear. The no-op case is in `teardown_method`, which passes today only because it discards the return value - so this couples to my line 74 comment.

Suggested wording in `leakage_sensor_test_base.py`:

```python
        Idempotent: the return value reports whether the sensor is left in the
        requested state, not whether a change was made. Injecting on a sensor
        that is already injected, or withdrawing from one that is already
        clear, must return True.
```

and for `clear_test_leaks()`: "True if no test leak remains injected on any sensor, including when none was injected to begin with."

tests/leakage_sensor_test_base_test.py
```diff
@@ -0,0 +1,152 @@
...
+        if enable:
+            if not sensor.test_leak:
+                self._saved[sensor_name] = (sensor.leaking,
+                                            sensor.leak_severity)
+            sensor.leaking = True
+            sensor.test_leak = True
+            sensor.leak_severity = severity
+        elif sensor.test_leak:
+            sensor.leaking, sensor.leak_severity = \
+                self._saved.pop(sensor_name)
```
nikamirrr, line 41: The injection state is split across two objects: `test_leak` lives on the sensor, `_saved` lives on the `PlatformLeakTest`. Nothing in `get_leak_sensor_test()`'s contract requires returning a stable instance, so a platform that builds a fresh `LiquidCooling` (and thus a fresh leak-test object) over cached or shared sensor objects — e.g. `Chassis().get_liquid_cooling()` — reaches this `pop` with `test_leak` set and an empty `_saved`, and the `KeyError` propagates out of `LeakTestApiBase.teardown_method`.

Since this class is presented as the reference that platforms will copy, the split-state design propagates with it. Keeping the saved state on the sensor, or `self._saved.pop(sensor_name, None)`, would close it.

sonic_platform_base/leak_test_api_base.py
```diff
@@ -0,0 +1,173 @@
...
+        Asserts no sensor reports a leak, a test leak, or an armed injection.
+        '''
+        assert self.liquid_cooling.get_leak_sensor_status() == []
+        for sensor in self.liquid_cooling.get_all_leak_sensors():
+            assert sensor.is_leak() == False
+            assert sensor.is_test_leak() == False
+        for name in self.SENSOR_NAMES:
+            assert self.leak_test.is_test_leak_enabled(name) == False
+
+    def test_sensor_names(self):
```
judyjoseph, line 87: Hi @chinmoy-nexthop I feel we really don't need the test infra which you are defining here in this class **LeakTestApiBase**. This can be defined and used as needed from the sonic-mgmt test framework - can leave it to platform test writer to how to call the platform APIs to test leak. Let me know your thoughts

The other changes in sonic_platform_base/leak_sensor_test_base.py, sonic_platform_base/liquid_cooling_base.py is defined as we discussed in community REF: https://github.com/sonic-net/SONiC/pull/2441.

chinmoy-nexthop: Absolutely!
Make sense @judyjoseph we should use this https://github.com/sonic-net/SONiC/pull/2441 .

parvathi-nexthop:
Thanks for the changes. The capability model in #2441 looks right to me, and there's one case worth making sure it keeps covering: platforms where the injected severity isn't selectable. On those, the severity a sensor reports is a fixed property of the sensor, so injection can assert a leak but not choose the severity it comes back as. #2441 handles this already — *"a fixed-severity binary sensor can expose only `LEAK_INJECTION`"* — but the `set_test_leak()` signature in this PR can't express it, since `severity` defaults to `CRITICAL` and is applied unconditionally. Keeping the capability negotiation would be worth it for that alone.

One case the design doesn't cover yet: injection mechanisms that can hold only one injection at a time. `get_test_capabilities()` is per-sensor with no concurrency dimension, so there's no defined answer for "inject on B while A is injected" — it would either need advertising, or a defined failure. Worth pinning down before platforms implement it, since test code that assumes it can inject on several sensors at once would silently mean different things on different hardware. Nexthop expects to have a platform in this shape which is why I'm rather than leaving theoretical.

parvathi-nexthop (changes requested):
Thanks for the changes. The capability model in #2441 looks right to me, and there's one case worth making sure it keeps covering: platforms where the injected severity isn't selectable. On those, the severity a sensor reports is a fixed property of the sensor, so injection can assert a leak but not choose the severity it comes back as. #2441 handles this already — "a fixed-severity binary sensor can expose only LEAK_INJECTION" — but the set_test_leak() signature in this PR can't express it, since severity defaults to CRITICAL and is applied unconditionally. Keeping the capability negotiation would be worth it for that alone.

One case the design doesn't cover yet: injection mechanisms that can hold only one injection at a time. get_test_capabilities() is per-sensor with no concurrency dimension, so there's no defined answer for "inject on B while A is injected" — it would either need advertising, or a defined failure. Worth pinning down before platforms implement it, since test code that assumes it can inject on several sensors at once would silently mean different things on different hardware. Nexthop expects to have a platform in this shape which is why I'm rather than leaving theoretical.

sonic_platform_base/liquid_cooling_base.py
nikamirrr: Double checked, please consider adding it to __init__ as well

sonic_platform_base/liquid_cooling_base.py
nikamirrr: Concrete suggestion, since the cleanest fix touches the `is_leak()` contract rather than just this line.

What the three implementations do today:

| | `is_leak()` | keeps `self.leaking` in sync? | `get_leak_severity()` | `is_test_leak()` |
|---|---|---|---|---|
| `LeakageSensorBase` | `return self.leaking` | n/a | base | base |
| mellanox `LeakageSensor` | sysfs read | yes on the 1/0 branches, no on the error branch (returns `'N/A'`) | base | base |
| aspeed `LeakageSensor` | sysfs read | never | overridden | base |

**Option 1 - tighten the `is_leak()` contract, then read the cached state.**

```python
    def is_leak(self) -> bool:
        """
        Retrieves the leak status of the sensor.
        The platform should apply debounce logic before reporting/clearing leak.

        Implementations that read hardware must record the result in
        self.leaking and must return a bool. The other accessors on this class
        read self.leaking rather than calling is_leak() again, so one is_leak()
        call per poll gives a consistent view of the sensor.

        Returns:
            bool: True if leak is detected, False if not
        """
        return self.leaking

    def get_leak_severity(self) -> LeakSeverity|None:
        return self.leak_severity if self.leaking else None

    def is_test_leak(self) -> bool:
        return bool(self.leaking and self.test_leak)
```

`get_leak_sensor_status()` already calls `is_leak()` on every sensor, so that call is the refresh and everything downstream then reads a consistent snapshot. This removes the extra hardware read, the TOCTOU window and the `'N/A'`-is-truthy trap in one go, and makes the `-> bool` annotation true.

The cost is that it needs two one-line vendor follow-ups landing alongside it, because neither implementation satisfies the tightened contract today:

```python
# mellanox .../sonic_platform/liquid_cooling.py - return a bool on read failure
        else:
            logger.error(f"Failed to read leakage sensor {self.name} value: {content}")
            self.leaking = False
            return False

# aspeed .../ast2700/sonic_platform/leakage_sensor.py - record the reading
    def is_leak(self):
        self.leaking, _, _ = self._check_channel_value()
        return self.leaking
```

Without the aspeed one, `is_test_leak()` would read a `self.leaking` that aspeed never sets and always return `False` - a silent regression versus the code as it stands here.

**Option 2 - if you would rather not couple this PR to the vendor repos:** drop the state dependence and revert to `return self.leak_severity`, with the docstring describing it as a static property of the sensor rather than of the current leak. `get_leak_sensor_status()` already tells a consumer which sensors are leaking, so the severity of a sensor obtained from it is unambiguous. Zero extra I/O, no vendor coupling, and consistent with `get_leak_sensor_type()` / `get_leak_sensor_location()`, which are all plain attribute reads.

Option 1 is the more correct model, option 2 is the smaller change - either resolves what I flagged.

nikamirrr: Double checked, please consider either approach to keep self.leaking and self.leak_severity in sync

sonic_platform_base/liquid_cooling_base.py
nikamirrr: same as above. Double checked, please consider either approach to keep self.leaking and self.leak_severity in sync

sonic_platform_base/liquid_cooling_base.py
nikamirrr: More like a nit, make sure it returns the correct bool, set the return type of is_leak() etc

sonic_platform_base/leak_test_api_base.py
nikamirrr: Double checked, please see if need to WA the bare assert to make sure the error is raised even in the optimized python environment

sonic_platform_base/leak_test_api_base.py
nikamirrr: Consider asserting that self.leak_test is not None

sonic_platform_base/leak_test_api_base.py
nikamirrr: Double checked, please consider asserting the returned value

sonic_platform_base/leak_test_api_base.py
nikamirrr: Double checked. Consider using s unittest.SkipTest rather than asserting is_leak_test_supported is True

sonic_platform_base/leak_test_api_base.py
nikamirrr: Double checked. Please resolve the ambiguity, most likely it the docstring, 
True -> if the final state of the sensor is what is requested
False -> if not

Not indicating the sensor state change

tests/leakage_sensor_test_base_test.py
nikamirrr: Double checked, please consider moving _saved to the sensor, rather then keeping in PlatformLeakTest

sonic_platform_base/leak_test_api_base.py
nikamirrr: Double checked. Please review the datapaths of the SENSOR_NAMES, making sure it is not compared to itself

sonic_platform_base/liquid_cooling_base.py
chinmoy-nexthop: The severity default is now  LeakSeverity.MINOR , as suggested. 

- Plumbing the flag into the published leak record and honouring it in the health checker requires changes in  sonic-buildimage  ( hardware_checker.py ,  sonic-leak-control.yang ) and  sonic-utilities  ( show/platform.py ), which are outside this repo. This PR only lands the platform API. 

- the consumer changes should  be added when vendor enables injection. for now log should be good enough

sonic_platform_base/liquid_cooling_base.py
chinmoy-nexthop: Done — `self.test_leak = False` is now set in `__init__` alongside `leaking`,
`leak_sensor_ok`, `leak_type`, `leak_location` and `leak_severity`, so the initialization
convention is consistent across all six attributes:

    def __init__(self, name, *, type=None, location=None,
                 severity=LeakSeverity.CRITICAL):
        self.name = name
        self.leaking = False
        self.leak_sensor_ok = True
        self.leak_type = type or 'unknown'
        self.leak_location = location or 'unknown'
        self.leak_severity = severity
        self.test_leak = False

Any instance built through `__init__` now shadows the class attribute immediately, so
`LeakageSensorBase.test_leak = True` no longer flips those sensors.

I kept the class-level `test_leak = False` as well, purely as a read fallback for vendor
subclasses that override `__init__` without chaining `super().__init__()` — without it
`is_test_leak()` raises `AttributeError` on such a sensor, which would turn a missing
super() call into a crash in the leak polling path rather than a safe default. It is no
longer the primary initialization point, just a backstop. Happy to drop it if you'd rather
the missing-super() case fail loudly.

sonic_platform_base/liquid_cooling_base.py
chinmoy-nexthop: Took Option 1 — thanks for writing it out, it matches verbatim what is now in the file.

    def is_leak(self) -> bool:
        """
        ...
        Implementations that read hardware must record the result in
        self.leaking and must return a bool. The other accessors on this class
        read self.leaking rather than calling is_leak() again, so one is_leak()
        call per poll gives a consistent view of the sensor.
        """
        return self.leaking

    def get_leak_severity(self) -> LeakSeverity|None:
        return self.leak_severity if self.leaking else None

    def is_test_leak(self) -> bool:
        return bool(self.leaking and self.test_leak)

That removes the extra hardware read, the TOCTOU window between
`get_leak_sensor_status()` and `get_leak_severity()`, and the `'N/A'`-is-truthy trap,
and makes the `-> bool` annotation honest. `get_leak_sensor_status()` calling `is_leak()`
per sensor is the refresh point, so everything downstream reads one consistent snapshot.

On the two vendor follow-ups you identified — agreed, both are needed and neither can land
in this repo. I will raise them against `sonic-buildimage` alongside this PR:

  * mlnx-platform-api `sonic_platform/liquid_cooling.py` — set `self.leaking = False` and
    return `False` on the read-failure branch instead of `'N/A'`.
  * nvidia-bmc ast2700 `sonic_platform/leakage_sensor.py` — record the reading:
    `self.leaking, _, _ = self._check_channel_value()` before returning.

Until the aspeed one merges, `is_test_leak()` on that platform reads a `self.leaking` it
never sets, so it returns `False`. I have called this out explicitly in the PR description
so it is not a silent regression, and the API PR will not be a dependency for anything
until the vendor changes are in.

sonic_platform_base/liquid_cooling_base.py
judyjoseph: Wondering if we need special treatment for test_leak keeping it as class leak .. ideally the vendor implementation should be calling super().__init__() ?

judyjoseph (dismissed):
Looks ok to me .. 

One comment on the below defenition
```
test_leak: bool = False as a class attribute
```

chinmoy-nexthop:
> Looks ok to me ..
> 
> One comment on the below defenition
> 
> ```
> test_leak: bool = False as a class attribute
> ```

Thank you @judyjoseph , Removed the redundant  test_leak  class attribute from  LeakageSensorBase

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -82,21 +93,37 @@ def get_leak_sensor_location(self) -> str:
...
+
+        A test leak is observable end to end (it is published like any other
+        leak) but must never be used to trigger a mitigation action.
+
+        Returns:
+            bool: True if the reported leak was injected by a leak test,
+                  False otherwise
         """
-        return self.leak_severity
+        return bool(self.leaking and self.test_leak)
```
nikamirrr, line 118: **`is_test_leak()` raises `AttributeError` on a leaking subclass that doesn't chain `super().__init__()`.**

Following up on my line-34 comment. The reply there said the class-level `test_leak = False` was kept as a read fallback for exactly this case — on this head there is only the `__init__` assignment on line 43, so the fallback is gone.

A subclass that sets `name` / `leaking` / `leak_sensor_ok` / `leak_type` / `leak_location` / `leak_severity` itself and never chains — the `NoSuper` shape in `tests/liquid_cooling_base_test.py:95` — raises `AttributeError: 'NoSuper' object has no attribute 'test_leak'` as soon as it reports a leak.

`self.leaking and self.test_leak` short-circuits only while the sensor is *not* leaking, and `test_is_test_leak_without_init_chaining` sets `leaking = False`; its docstring is explicit that the flag "is only read once the sensor is leaking". So the test passes without covering the case that fails, and the failure first appears on a real leak on an out-of-tree platform.

This repo's `.github/copilot-instructions.md` is explicit about it:

> **Default implementations**: New methods MUST have safe defaults (return None, 'N/A', False, etc.)

Either restore the class attribute alongside the `__init__` assignment, or:

```suggestion
        return bool(self.leaking and getattr(self, 'test_leak', False))
```

@judyjoseph — agreed that vendors *should* chain `super().__init__()`, but `LeakageSensorBase` does not chain to `SensorBase` itself, so there is no precedent in the file, and out-of-tree plugins are what break.

sonic_platform_base/liquid_cooling_base.py
```diff
@@ -82,21 +93,37 @@ def get_leak_sensor_location(self) -> str:
...
+    def get_leak_severity(self) -> LeakSeverity|None:
         """
         Retrieves the severity of leak
 
         Returns:
-            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None if no leak
+            LeakSeverity: LeakSeverity.CRITICAL or LeakSeverity.MINOR, or None
+            if no leak
+        """
+        return self.leak_severity if self.leaking else None
```
nikamirrr, line 104: Following up on the Option 1 thread — the code here matches what we landed on, thanks.

The two vendor-side changes that Option 1 depends on do not appear to have been raised yet. I do not see PRs in `sonic-buildimage` for either:

* mlnx-platform-api `sonic_platform/liquid_cooling.py` — return `False` (and set `self.leaking = False`) rather than `'N/A'` on the read-failure branch.
* nvidia-bmc ast2700 `sonic_platform/leakage_sensor.py` — record the reading: `self.leaking, _, _ = self._check_channel_value()`.

Until the ast2700 change lands, `self.leaking` there is only ever whatever `set_test_leak()` leaves behind — `is_leak()` never records it. So injection via a hardware test bit (the model this ABC documents, and the only one ast2700's hardware-reading `is_leak()` can observe) gives `is_leak() == True` with `self.leaking == False`, and `is_test_leak()` returns `False`. In the polling daemon it is `False` either way, since that process never calls `set_test_leak()` — which is the process where the flag needs to be right.

That reply also said the gap was called out explicitly in the PR description. The description currently has the three-bullet summary plus the untouched template. Could the caveat, and links to the two `sonic-buildimage` PRs, go in before this merges? That is mainly what keeps it from being a silent regression for whoever debugs it later.

tests/leakage_sensor_test_base_test.py
```diff
@@ -0,0 +1,254 @@
...
+        return True
+
+    def is_test_leak_enabled(self, sensor_name):
+        sensor = self._sensors.get(sensor_name)
+        return bool(sensor and sensor.test_leak)
+
+    def clear_test_leaks(self):
+        return all(self.set_test_leak(name, False)
+                   for name, sensor in self._sensors.items()
+                   if sensor.test_leak)
```
nikamirrr: **`all()` over a generator short-circuits, leaving the remaining sensors armed.**

The ABC docstring says `clear_test_leaks()` "withdraws every injected test leak on the platform". If withdrawal on one sensor fails (a transient sysfs/BMC error), `all()` stops iterating there and every later sensor stays injected — with no further attempt, and nothing in the return value separating "one withdrawal failed" from "we stopped early".

This is the reference implementation vendors will copy, so it is worth making it attempt every sensor:

```suggestion
        return all([self.set_test_leak(name, False)
                    for name, sensor in self._sensors.items()
                    if sensor.test_leak])
```

nikamirrr (changes requested):
Thanks for the rework — the `is_leak()` / `get_leak_severity()` / `is_test_leak()` contract landed as we discussed on the Option 1 thread, and the `TYPE_CHECKING` guard for the `LeakageSensorTestBase` forward reference is in. CI is green here.

One code item left, plus a process one:

1. **`is_test_leak()` is no longer safe on subclasses that skip `super().__init__()`.** The reply on the line-34 thread said the class-level `test_leak = False` was kept as a read fallback; it is not on this head, so a leaking out-of-tree sensor now raises `AttributeError`. One-line fix, commented inline.
2. **The Option 1 follow-ups.** The two `sonic-buildimage` changes this depends on do not seem to be raised yet, and the caveat that was going into the PR description is not there. Until the ast2700 change lands, `is_test_leak()` is hardwired to `False` on that platform — worth stating in the description so it is discoverable.

The third comment is a suggestion on the reference `clear_test_leaks()`, which vendors will copy.

Happy to re-review as soon as the first one is addressed.

sonic_platform_base/liquid_cooling_base.py
judyjoseph: Currently vendor subclasses present in sonic-buildimage chain super().init(), we should be good @nikamirrr 
```
   mellanox mlnx-platform-api/sonic_platform/liquid_cooling.py line 43:                                                                                               
   super(LeakageSensor, self).init(name)                                   

   nvidia-bmc ast2700 .../sonic_platform/leakage_sensor.py line 51:                                                                                                   
   super().init(name, type=sensor_type, location=location)
```

judyjoseph (approved):
LGTM

nikamirrr (approved):

chinmoy-nexthop:
Thank you for looking into this, @judyjoseph  @nikamirr  @parvathi-nexthop . I really appreciate your valuable comments and feedback.

PR #742, opened 2026-08-18
Add CpoCmisApi

bgallagher-nexthop:
#### Description
`CmisApi` has aggregate functions that read multiple registers and return a dict containing logically grouped information back to `xcvrd`. For instance, `get_transceiver_info` is a good example of one of these functions.

This PR implements CPO-specific subclass of `CmisApi` that splits those aggregate functions into smaller functions that return banked or non-banked information. Taking `get_transceiver_info` for example, there will now be 3 functions:
- `get_banked_transceiver_info`, which will return only bank-dependent information
- `get_non_banked_transceiver_info`, which will return only bank-independent information
- `get_transceiver_info` will continue to exist, but internally it will just call the two above functions and return the union of their output.

This functionality is limited to a CPO subclass since the ability to read banked and non-banked information independently is only required by CPO at this time.

#### Motivation and Context
For CPO, there is a desire to limit reading non-banked information from the hardware for each port because many ports may share the same device. Reading non-banked information multiple times from the same device for each port would waste i2c read bandwidth, so the functions have been split to allow `xcvrd` to selectively read non-banked information only as much as necessary and avoid duplicate reads.

A new test class for `CpoCmisApi` has been introduced that will run all existing test-cases defined for `CmisApi` against the subclass. The subclass adds no new functionality, just exposes new functions.


prgeor (commented):
@bgallagher-nexthop I was wondering if we should deal with this nuance in cpo specific CMIS api override?

bgallagher-nexthop:
> @bgallagher-nexthop I was wondering if we should deal with this nuance in cpo specific CMIS api override?

I think that approach would work too:

1. We could create a `CpoCmisApi` subclass of `CmisApi`
2. We could re-implement the aggregate methods like `get_transceiver_info` in that subclass, with the banked and non-banked variants also implemented.

I think the downside is that we will have a maintenance cost of keeping `CpoCmisApi` up-to-date. Any update to the overriden aggregate methods in `CmisApi` will also require making those same changes to `CpoCmisApi`. Or we accept that the `CpoCmisApi` can become stale as time goes on.

What do you think?

bgallagher-nexthop:
Changed the approach to use a CPO specific subclass that splits some aggregate methods into banked / non-banked constituent methods instead of modifying `CmisApi` directly, as suggested by reviewers. Also got rid of `get_transceiver_info` from this class -- it does not require the banked / non-banked methods since it is only called once upon transceiver insertion.

sonic_platform_base/sonic_xcvr/cpo/cmis.py
```diff
@@ -0,0 +1,219 @@
...
+
+from sonic_platform_base.sonic_xcvr.api.public.cmis import CmisApi
+
+
+class CpoCmisApi(CmisApi):
+    def get_non_banked_transceiver_dom_real_value(self):
+        temp = self.get_module_temperature()
+        voltage = self.get_voltage()
+        if temp is None or voltage is None:
+            return None
```
Natgerbi, line 32: I prefer in such case to print what we can.. for the one with None we can set N/A or unknown

sonic_platform_base/sonic_xcvr/cpo/cmis.py
```diff
@@ -0,0 +1,219 @@
...
+        bulk_status = {
+            "temperature": temp,
+            "voltage": voltage
+        }
+
+        laser_temp_dict = self.get_laser_temperature()
+        try:
+            bulk_status['laser_temperature'] = laser_temp_dict['monitor value']
+        except (KeyError, TypeError):
+            pass
```
Natgerbi: can we have some log here for the failure?

sonic_platform_base/sonic_xcvr/cpo/cmis.py
```diff
@@ -0,0 +1,219 @@
...
+
+        return bulk_status
+
+    def get_banked_transceiver_dom_real_value(self):
+        tx_bias = self.get_tx_bias()
+        rx_power = self.get_rx_power()
+        tx_power = self.get_tx_power()
+        if tx_bias is None or rx_power is None or tx_power is None:
+            return None
```
Natgerbi, line 54: same as above, single failure shouldn't lead to returning nothing..

sonic_platform_base/sonic_xcvr/cpo/cmis.py
```diff
@@ -0,0 +1,219 @@
...
+        status_flags_dict = dict()
+        try:
+            dp_fw_fault, module_fw_fault, module_state_changed = self.get_module_firmware_fault_state_changed()
+            status_flags_dict.update({
+                'datapath_firmware_fault': dp_fw_fault,
+                'module_firmware_fault': module_fw_fault,
+                'module_state_changed': module_state_changed
+            })
+        except TypeError:
+            pass
```
Natgerbi: please log the error we had.. it will be a lot easier to debug with logs on failures..

sonic_platform_base/sonic_xcvr/cpo/cmis.py
bgallagher-nexthop: Added logging

> for the one with None we can set N/A or unknown

The reason for returning `None` if a single field is absent is to adhere to convention around retries -- `xcvrd` expects None if any read failed, so determining whether a retry is required is simple. See [this comment](https://github.com/sonic-net/sonic-platform-common/blob/c07f0740aeebcfd10417ba444381e7b3c71b212f/sonic_platform_base/sonic_xcvr/api/public/cmis.py#L398-L402).

Natgerbi: I couldn't find any retry or set SFP_EEPROM_NOT_READY to transceiver_dom_real_value at all.

from what i see only get_transceiver_info, get_transceiver_pm and get_transceiver_firmware_info will cause retry of read ..

bgallagher-nexthop: Yeah, this is true, the DOM task will not retry since it is a periodic poller (a "retry" is going to happen on the next periodic poll). I agree with you that it would be nice to return as much info as we can for the DOM-related functions specifically, but I am not sure this particular PR is the right place to change that expectation. If the CMIS API was to switch to that approach, I think `CmisApi` should also change so the contract around failed reads is the same for CPO and non-CPO hardware.

Natgerbi (approved):
LGTM

abelamit (approved):

aditya-nexthop (approved):

prgeor (approved):

PR #743, opened 2026-08-18
Fix bug where latched information was read for unrelated banks in ElsfpApi

bgallagher-nexthop:
#### Description

Screenshot 2026-08-17 at 10 28 36 PM

In bytes 166-169 and bytes 174-177 of page 1Ah of the ELSFP memory map, fault and warning flags for each lane of the ELSFP are laid out in memory in a non-banked fashion (see below screenshot):

Screenshot 2026-08-17 at 10 29 37 PM

This PR changes the API to make sure we only read the byte related to the bank associated with the current `ElsfpApi` object. Otherwise, the other bytes related to other banks are read and their contents are cleared since they are latched registers.

#### Motivation and Context
This fixes a correctness bug in the `ElsfpApi` that would cause latched information to be cleared for all banks when a read of these flags is issued by a particular port/bank.

Added unit-tests.


sonic_platform_base/sonic_xcvr/fields/elsfp_consts.py
```diff
@@ -39,8 +39,33 @@
 LANE_SUMMARY_STATUS = "ElsfpLaneSummaryStatus"
 LANE_SUMMARY_FAULT = "LaneSummaryFault"
 LANE_SUMMARY_WARNING = "LaneSummaryWarning"
-FAULT_FLAG_LANE_FIELD = "FaultFlagLanes"
-WARN_FLAG_LANE_FIELD = "WarnFlagLanes"
+
+FAULT_FLAG_LANE_FIELD_BANK_0 = "FaultFlagLanesBank0"
```
prgeor, line 43: @bgallagher-nexthop why only upto 4 banks?

bgallagher-nexthop: This is just the way it is defined in page 1Ah of the latest ELSFP spec. While CMIS 5.4 supports more than 4 banks, the ELSFP memory map only has four bytes each for these fault and warning flags (they are not banked, just laid out flat on the page).

sonic_platform_base/sonic_xcvr/api/public/elsfp.py
```diff
@@ -120,32 +120,42 @@ def get_per_lane_fault_flags(self) -> dict[str, int]:
...
+              "FaultFlagLane9": 0,
+              "FaultFlagLane10": 1,
+              "FaultFlagLane11": 0,
               # ... etc ...
-              "FaultFlagLane32": 1,
+              "FaultFlagLane16": 1,
           }
         """
-        return self.xcvr_eeprom.read(elsfp_consts.FAULT_FLAG_LANE_FIELD)
+        return self.xcvr_eeprom.read(elsfp_consts.FAULT_FLAG_LANE_FIELDS[self.xcvr_eeprom.mem_map.bank])
```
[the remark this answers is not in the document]
bgallagher-nexthop: I have let this throw on purpose, as an invalid bank will cause issues elsewhere and bank validation is better done at the time of SfpBase/CpoBase object creation somewhere like `ChassisBase`.

micas-net (approved):
LGTM!

abelamit (approved):

aditya-nexthop (approved):
good catch @bgallagher-nexthop !

prgeor (approved):

PR #759, opened 2026-09-16
[sonic_xcvr]: Preserve VDM control bits when freezing stats

mzwang-arista:
#### Description

Update CMIS VDM freeze and unfreeze operations to use read-modify-write
on the VDM control register. This changes only the freeze control bit
(bit 7), preserving other control bits defined by CMIS 5.4.

Add unit coverage for preserved control bits, EEPROM read failure without
a write, and EEPROM write failure.

#### Motivation and Context

The existing implementation writes a hardcoded value to the entire VDM
control byte (page 0x2F, byte 144). CMIS 5.4 defines additional control
bits in this byte; overwriting it can unintentionally clear or modify
them, causing VDM data to stop being reported.

Ran the focused CMIS unit-test suite using `sonic-py-common` from
SONiC buildimage:

```text
PYTHONPATH=/garage/mzwang/sonic-buildimage/src/sonic-py-common \
pytest -o addopts='' -q tests/sonic_xcvr/test_ccmis.py
```

Result:

```text
47 passed in 0.30s
```

Also tested on a module where the prior implementation cleared
additional VDM control bits, causing VDM data not to be reported.
With this change, freeze/unfreeze preserves those bits and VDM
reporting continues normally.


mzwang-arista:
Duplicate https://github.com/sonic-net/sonic-platform-common/pull/757

PR #761, opened 2026-09-17
Fixing logging used by sonic_platform_base classes to prevent log pollution

### LINKED ISSUE #760: [sonic_platform_base] Module level loggers overwriting log identifiers for processes that import `sonic_platform`
`sonic_platform_base/__init__.py` imports base classes unconditionally, so *any* process that touches `sonic_platform_base` silently has its syslog identifier replaced with base class loggers’s log tags if it uses `syslog.openlog` e.g. `sonic_py_common.logger.Logger`.

ZTP is the clearest case:

1. `ztp` service starts and calls `syslog.openlog(ident='sonic-ztp')`
2. `ztp-engine.py` imports `sonic_platform.platform` at module level
3. that pulls in `platform_base.py` → `sonic_platform_base/__init__.py` → `bmc_base.py`
4. `bmc_base.py` creates `Logger('bmc_base')`, calling `syslog.openlog(ident='bmc_base')`
5. every ZTP message from then on is logged as `bmc_base`

It is not specific to ZTP — anything that imports `sonic_platform.platform` and logs through `syslog` via `sonic_py_common.logger.Logger`.

gordon-nexthop:
#### Description

Convert the module-level loggers in `sonic_platform_base` from
`sonic_py_common.logger.Logger` to `sonic_py_common.syslogger.SysLogger`:
`bmc_base.py`, `redfish_client.py`, `sed_mgmt_base.py` and `bmc_fw_update.py`.

`Logger` is a thin wrapper around the C `syslog` API, and its constructor calls
`syslog.openlog()`, which sets **one** identifier for the **whole process**. These modules
build a `Logger` at import time, so simply importing them renames every subsequent `syslog`
message emitted anywhere in the importing process.

`SysLogger` instead attaches a `SysLogHandler` to a named `logging.Logger`, so the identifier
belongs to the logger instance rather than to the process. Importing a module can no longer
change how unrelated code in the same process is tagged, and each module still logs under its
own name. This is already the pattern used by `sonic_platform_base/sonic_xcvr/cdb/`.

#### Motivation and Context

fixes: https://github.com/sonic-net/sonic-platform-common/issues/760

**On hardware.** Nexthop 4010 (`x86_64-nexthop_4010-r0`, HwSKU NH-4010, SONiC master,
Python 3.13) — deliberately a **non-BMC** platform, since the bug reaches those too.

1. *ZTP import chain.* A probe that mirrors what `ztp-engine.py` does — construct
   `Logger('sonic-ztp')`, log, `import sonic_platform.platform`, log again — with
   `syslog.openlog` wrapped to record every identifier it is called with:

   | | `openlog()` idents, in order | effective process ident |
   |---|---|---|
   | before | `['sonic-ztp', 'redfish_client', 'bmc_base']` | `bmc_base` |
   | after | `['sonic-ztp']` | `sonic-ztp` |

   and the resulting `/var/log/syslog` lines:

   ```
   # before
   NOTICE sonic-ztp:  probe before-import: ZTP service started.
   NOTICE bmc_base:   probe after-import: ZTP service still running.
   # after
   NOTICE sonic-ztp:  probe before-import: ZTP service started.
   NOTICE sonic-ztp:  probe after-import: ZTP service still running.
   ```

2. *The real ztp service.* `systemctl start ztp` (ZTP admin mode disabled, so it just bails)
   now logs under its own identifier, and no `bmc_base`-tagged line was emitted by anything on
   the switch after the swap:

   ```
   INFO sonic-ztp[768842]: ZTP is administratively disabled.
   ```

3. *No regression to BMC logging.* Importing the modules and logging through them directly
   still produces their own identifiers — now with the PID, since `SysLogger` formats as
   `%(name)s[%(process)d]` — and `RedfishClient` still constructs:

   ```
   NOTICE redfish_client[768154]: ...
   NOTICE bmc_base[768154]: ...
   NOTICE sed_mgmt_base[768154]: ...
   ```

4. *Platform API smoke test.* `show platform summary`, `show platform psustatus` and
   `docker exec pmon supervisorctl status` were all healthy with the patched modules in place
   (pcied, psud, sensormond, stormond, syseepromd, thermalctld, xcvrd all RUNNING).


gordon-nexthop:
@benle7 would you mind taking a peak at this one!

